PR #103, opened 2021-07-05
Feature/remove calls per transfer when committing

shanefontaine:


packages/hop-node/src/watchers/CommitTransferWatcher.ts
```diff
@@ -107,6 +111,8 @@ class CommitTransfersWatcher extends BaseWatcherWithEventHandlers {
     if (this.isL1) {
       return
     }
+
+    const commitTransferWatcherPollInterval = 6 * 10 * 1000
```
miguelmota, line 115: Instead of creating a new variable, you can assign `this.pollIntervalSec = 60` in constructor

packages/hop-node/src/watchers/CommitTransferWatcher.ts
```diff
@@ -19,12 +19,16 @@ export interface Config {
...
+  sentCommitTx: boolean
+}
+
 const BONDER_ORDER_DELAY_MS = 60 * 1000
 
 class CommitTransfersWatcher extends BaseWatcherWithEventHandlers {
   siblingWatchers: { [chainId: string]: CommitTransfersWatcher }
-  minPendingTransfers: number = 1
   minThresholdAmount: BigNumber = BigNumber.from(0)
+  lastCommitForChainId: { [chainId: number]: LastCommitForChainId } = {}
```
miguelmota, line 31: The LastCommitForChainId only contains one property, so we can simplify by doing

```js
sentCommitTxForChainId: { [chainId: number]: boolean } = {}
...
 if (sentCommitTxForChainId[destinationChainId]) {
   ...
   return
 }
 ...
 this.sentCommitTxForChainId[destinationChainId] = true
```

packages/hop-node/src/watchers/classes/L2Bridge.ts
```diff
@@ -228,6 +228,23 @@ export default class L2Bridge extends Bridge {
...
+      index
+    )
+  }
+
+  async doPendingTransfersExist (chainId: number): Promise<boolean> {
+    try {
+      await this.getPendingTransferByIndex(chainId, 0)
+      return true
+    } catch (err) {
+      return false
```
miguelmota, line 244: In the case of an unrelated RPC error, this will always return false when there could be a pending transfer. 
The error message should be filtered

PR #105, opened 2021-07-07
hop-node: refactor bondWithdrawalWatcher RPC calls

shanefontaine:
This PR removes some RPC calls that appear to be unnecessary.

The biggest issue is that we would loop through each item marked `TransferSent` and make 5 RPC calls per chain on initial sync, in parallel, even if withdrawals were already bonded. This happened even if the transfer was bonded but the initial sync simply wasn't up-to-date.

- Do not allow txs to be sent during initial sync in order to guarantee DB state
  - We had a ton of calls being made during initial sync (5 RPC calls per any transfer ever sent in our system, in parallel + an event handler per tx)
- Removes an isBonder on chain call b/c we can assume only a bonder is running this software (1 RPC call per bond withdrawal)
- Remove `const isWithdrawalBonded = bondedAmount.gt(0) || isTransferIdSpent because` DB state is guaranteed (2 RPC calls per bond withdrawal + potential 1,000 block event sync)
- Do not validate that the transferRoot has been bonded on-chain because DB state is guaranteed (1 RPC call per bond withdrawal)
- Only check that the bonder has enough credit and not that the have >0 credit (1 RPC call per bond withdrawal)
- Increase wait confirmation checker from 2s to 20s. The total wait time is 144s, 210s, 248s on mainnet/xdai/poly, respectively. In the absolute worst case, a user will wait an extra 19.9s (saves 60-110 calls per bond withdrawal)

### DIFF
--- a/packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
+++ b/packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
@@ -150,6 +150,11 @@ class BondWithdrawalWatcher extends BaseWatcherWithEventHandlers {
   }
 
   async checkTransferSentFromDb () {
+    const initialSyncCompleted = await this.isAllSiblingWatchersInitialSyncCompleted()
+    if (!initialSyncCompleted) {
+      return false
+    } 
+
     const dbTransfers = await db.transfers.getUnbondedSentTransfers({
       sourceChainId: await this.bridge.getChainId()
     })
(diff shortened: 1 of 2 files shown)


packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
```diff
@@ -360,12 +324,6 @@ class BondWithdrawalWatcher extends BaseWatcherWithEventHandlers {
       )
       const l2Bridge = this.getSiblingWatcherByChainId(destinationChainId)
         .bridge as L2Bridge
-      const hasPositiveBalance = await l2Bridge.hasPositiveBalance()
```
shanefontaine, line 363: @miguelmota curious if you think this is ok to remove since we have the `credit.lt(amount)` check right below which seemingly does the same thing.

miguelmota: yes it's ok to remove

miguelmota:
> Do not allow txs to be sent during initial sync in order to guarantee DB state 

Should we just make this the default across the watchers?

PR #108, opened 2021-07-08
Do not attempt to process transfer if safeConfirmations hasn'n been met

shanefontaine:


packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
```diff
@@ -163,9 +163,16 @@ class BondWithdrawalWatcher extends BaseWatcherWithEventHandlers {
         `checking ${dbTransfers.length} unbonded transfers db items`
       )
     }
+
+    const headBlockNumber = await this.bridge.getBlockNumber()
     const promises: Promise<any>[] = []
     for (let dbTransfer of dbTransfers) {
-      const { transferId } = dbTransfer
+      const { transferId, transferSentBlockNumber } = dbTransfer
+      const targetBlockNumber = transferSentBlockNumber + getSafeWaitConfirmations(this.chainSlug)
```
miguelmota, line 171: alternatively, this

```js
getSafeWaitConfirmations(this.chainSlug)
```

can be read like this

```js
this.bridge.waitConfirmations
```

shanefontaine: good call. resolved with 5cd3f

PR #109, opened 2021-07-10
Feature/dont bond if settled

shanefontaine:


packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
```diff
@@ -220,15 +220,15 @@ class BondWithdrawalWatcher extends BaseWatcherWithEventHandlers {
...
-        .bridge as L1Bridge
-      const transferRootConfirmed = await l1Bridge.isTransferRootIdConfirmed(
+      const dbTransferRoot = await db.transferRoots.getByTransferRootId(
         dbTransfer.transferRootId
       )
-      if (transferRootConfirmed) {
-        logger.warn('transfer root already confirmed. Cannot bond withdrawal')
-        return
+
+      if (dbTransferRoot?.withdrawalBondSettleTxSentAt) {
```
miguelmota, line 227: The withdrawalBondSettleTxSentAt value will always be unset during a fresh sync because it's only set when the bonder actively initiates the tx. Would this cause any issues here?

Another thing to note would be that it's possible for this condition to always be true if withdrawalBondSettleTxSentAt is not reset after a tx failure (ie node restarted before receiving receipt)

miguelmota (changes requested):

packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
shanefontaine: These are good points.

I believe the answer is to introduce a `withdrawalBondSettled` boolean flag on `transferRoots` db entry. This will be set/unset everywhere `dbTransferRoot.withdrawalBondSettleTxSentAt` is set (before/after the sending of the transaction). In order to handle node resets/syncs, we can also add this to `handleMultipleWithdrawalsSettledEvent()` watcher.

This event handler does not need to be ordered after the `handleTransfersCommittedEventForTransferIds` handler that associates Ids <> Roots because the `handleMultipleWithdrawalsSettledEvent` event has the rootHash as part of the event data. So basically a syncUp will see `HandleMultipleSettledEvents` and then mark the transfer root at sent.

Going to think through this a little more and then implement that, but I believe that should handle all cases. LMK what you think.

shanefontaine: Fixes [here](https://github.com/hop-protocol/hop/commit/4df53a283686085a075f8961b72082960da81e59).

shanefontaine:
Logic has been updated on another branch.

PR #114, opened 2021-07-14
hop-node: handle replaced transactions in commitTransferWatcher

shanefontaine:


packages/hop-node/src/watchers/CommitTransferWatcher.ts
```diff
@@ -219,21 +221,21 @@ class CommitTransfersWatcher extends BaseWatcherWithEventHandlers {
...
           if (receipt.status !== 1) {
-            this.isCommitTxSent[destinationChainId] = false
+            this.commitTxSentAt[destinationChainId] = 0
             throw new Error('status=0')
           }
           this.emit('commitTransfers', {
             destinationChainId
           })
-          this.isCommitTxSent[destinationChainId] = false
+          this.commitTxSentAt[destinationChainId] = 0
```
miguelmota, line 235: If the tx is successful then there's no need to reset the timestamp. Just wondering if the tx will be re-attempted if it's reset here before the watcher receives the event

packages/hop-node/src/watchers/CommitTransferWatcher.ts
```diff
@@ -177,11 +178,12 @@ class CommitTransfersWatcher extends BaseWatcherWithEventHandlers {
       }
 
       // Define new object on first run after server restart
-      if (!this.isCommitTxSent[destinationChainId]) {
-        this.isCommitTxSent[destinationChainId] = false
-      }
-
-      if (this.isCommitTxSent[destinationChainId]) {
+      const timestampOk = this.commitTxSentAt[destinationChainId] + TX_RETRY_DELAY_MS < Date.now()
```
miguelmota, line 181: by default `this.commitTxSentAt[destinationChainId]` will be undefined so it'll evaluate to NaN < Date.now(). I don't think it affects anything here but an fyi

shanefontaine: Sounds good. We do this elsewhere and have tested it before, so I believe we are good to go here.

packages/hop-node/src/watchers/CommitTransferWatcher.ts
shanefontaine: If we do not reset it, we will always have to wait `TX_RETRY_DELAY_MS` (not the worst thing). I believe resetting it will allow us to optimally commit transfers (i.e. commit transfers twice in 2 minutes if the min commit transfers is hit during that small timeframe).

> Just wondering if the tx will be re-attempted if it's reset here before the watcher receives the event
I don't think it will because the onchain pending amt will be `0` and the poller will `return` [here[(https://github.com/hop-protocol/hop/blob/develop/packages/hop-node/src/watchers/CommitTransferWatcher.ts#L193-L205).

miguelmota: Sounds good 👍

PR #115, opened 2021-07-15
hop-node: Update transfer sent item block number on reorge or mark as…

miguelmota:
… unbondable


packages/hop-node/src/watchers/classes/ContractBase.ts
```diff
@@ -1,10 +1,17 @@
 import rateLimitRetry from 'src/decorators/rateLimitRetry'
-import { BigNumber, Contract, Transaction, providers } from 'ethers'
+import { BigNumber, Contract, Transaction as EthersTransaction, providers } from 'ethers'
 import { Chain } from 'src/constants'
 import { EventEmitter } from 'events'
 import { chainIdToSlug, chainSlugToId, wait } from 'src/utils'
 import { config } from 'src/config'
 
+export class NotFoundError extends Error {}
```
shanefontaine, line 8: What is the purpose of this new Error class?

miguelmota: it's an artifact from something i was trying earlier. Should be removed now

PR #124, opened 2021-07-23
Feature/updated challenge watcher

shanefontaine:


packages/hop-node/src/watchers/ChallengeWatcher.ts
```diff
@@ -68,184 +76,137 @@ class ChallengeWatcher extends BaseWatcherWithEventHandlers {
...
+        continue
       }
-    }
 
-    if (found) {
-      logger.warn('transfer root committed')
-      return
+      const isRootHashCommitted = dbTransferRoot?.committed
+      if (isRootHashCommitted) {
+        sourceChainId = chainId
```
miguelmota, line 179: Why is sourceChainId here assigned the first chainId that is not L1 in the list? Generally confused about this for loop

packages/hop-node/src/watchers/watchers.ts
```diff
@@ -152,7 +153,7 @@ function startWatchers (
   const bondTransferRootWatchers: any = {}
   const settleBondedWithdrawalWatchers: any = {}
   const commitTransferWatchers: any = {}
-  for (const network of [Chain.Ethereum as string].concat(_networks)) {
+  for (const network of _networks) {
```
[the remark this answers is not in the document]
shanefontaine: Resolved with 3c57cd11af46c893b39742d7c7eee5149585c482

packages/hop-node/src/watchers/classes/BaseWatcherWithEventHandlers.ts
```diff
@@ -122,33 +122,36 @@ class BaseWatcherWithEventHandlers extends BaseWatcher {
...
       logger.debug(`event transactionHash: ${transactionHash}`)
       logger.debug(`bonder: ${bonder}`)
 
-      await this.db.transferRoots.update(transferRootHash, {
-        transferRootHash,
+      await this.db.transferRoots.update(root, {
+        transferRootHash: root,
         transferRootId,
-        committed: true,
+        confirmed: true,
```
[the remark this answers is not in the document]
shanefontaine: You are correct. `confirmed` and `confirmedAt` should be removed.

1. I believe `committed` should still be removed here, since `committed` should be handled in `checkTransfersCommitted`. In the case where this is truly fraudulent, the root will have never _actually_ been committed. Do you agree?
2. Removing `confirmed` and `confirmedAt` will require a change in the `getChallengeableTransferRoots()` function. This makes sense. I believe `item.confirmed` and `item.ConfirmedAt` should simply be replaced with `item.bonded`, as that makes the most logical sense. Do you agree?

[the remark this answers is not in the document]
miguelmota: 1. I think `committed: true` should only be set on the transfersComitted event, so it should be removed [here](https://github.com/hop-protocol/hop/blob/16529b7854b6dd736a48e67f014413ef291f0152/packages/hop-node/src/watchers/classes/BaseWatcherWithEventHandlers.ts#L150) and then the db getter should have `item.committed`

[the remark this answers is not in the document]
miguelmota: Are you still planning on removing confirmed and confirmedAt here? since the bonded root doesn't always imply the root is confirmed

packages/hop-node/src/watchers/classes/BaseWatcherWithEventHandlers.ts
```diff
@@ -122,33 +122,36 @@ class BaseWatcherWithEventHandlers extends BaseWatcher {
...
+        confirmedAt: eventTimestamp,
         bonded: true,
+        totalAmount: amount,
```
miguelmota, line 154: This `totalAmount` is set in the commitTransfers handler but if there's an transfer root published with the incorrect amount then it will be overridden here because the root hash is the same. I think it's best to keep them separate

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -133,34 +135,20 @@ class TransferRootsDb extends BaseDb {
...
-  // This will add new requirements to this return statement
   async getChallengeableTransferRoots (
     filter: Partial<TransferRoot> = {}
   ): Promise<TransferRoot[]> {
     const transfers = await this.getTransferRoots()
     return transfers.filter(item => {
+      // Do not check if a rootHash has been committed. A rootHash can be committed and bonded,
+      // but if the bond uses a different totalAmount then it is fraudulent. Instead, use the
+      // transferRootId. If transferRootIds do not match then we know the bond is fraudulent.
+      const isTransferRootIdValid = item.bondTransferRootId === item.transferRootId
```
miguelmota, line 148: Maybe I missed it but where does `bondTransferRootId` get set before any challenges?

packages/hop-node/src/watchers/ChallengeWatcher.ts
shanefontaine: Closing, as this is now stale.

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -133,34 +134,17 @@ class TransferRootsDb extends BaseDb {
...
-    const transfers = await this.getTransferRoots()
-    return transfers.filter(item => {
-      return (
-        !item.confirmed &&
-        !item.confirmedAt &&
-        item.bonded &&
-        !item.sentConfirmTx &&
-        !item.sentConfirmTxAt
+        item.transferRootHash &&
+        item.confirmed &&
```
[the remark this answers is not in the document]
shanefontaine: Closing, as this has been handled with a newer commit.

packages/hop-node/src/watchers/ChallengeWatcher.ts
```diff
@@ -68,184 +76,137 @@ class ChallengeWatcher extends BaseWatcherWithEventHandlers {
...
-      ) {
-        found = true
-        break
+
+    const currentChainId: number = await this.bridge.getChainId()
+    const activeChainIds: number[] = await this.bridge.getChainIds()
+    let isRootHashCommitted: boolean
+    let sourceChainId: number
+    for (const chainId of activeChainIds) {
+      if (currentChainId === chainId) {
```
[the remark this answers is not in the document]
shanefontaine: No longer relevant based on further conversation.

packages/hop-node/src/db/TransferRootsDb.ts
shanefontaine: You are correct.

`bondTransferRootId` (and `bondedAt`) were incorrectly placed in the `handleTransferBondChallengedEvent()` instead of `handleTransferRootBondedEvent()`. 

They will be moved.

shanefontaine: Updated 639073aa3c32acdaec0ecb6407c418ef073c8d58 and d8db75760d96b7e9754dac0c8f049a11da073376

miguelmota: perfect

packages/hop-node/src/watchers/classes/L1Bridge.ts
```diff
@@ -295,4 +314,10 @@ export default class L1Bridge extends Bridge {
     )
     return address !== ethers.constants.AddressZero
   }
+
+  @rateLimitRetry
+  async getChallengePeriod (): Promise<number> {
+    const challengePeriod = await this.getReadBridgeContract().challengePeriod()
+    return challengePeriod
```
miguelmota, line 321: Does this need to be casted to number type, ie `Number(challengePeriod.toString())`?

packages/hop-node/src/db/TransferRootsDb.ts
miguelmota: Just to confirm, isTransferRootIdValid will be false if only the transfer root is bonded and there's no commit event correct (because the transferRootId is not set elsewhere)?

shanefontaine: That is correct.

packages/hop-node/src/watchers/classes/L1Bridge.ts
shanefontaine: Good catch. Updated with b5f414b3c40ab5bfeb96d099d57da95aef47e44a

PR #125, opened 2021-07-28
Feature/remove real time watcher

shanefontaine:


packages/hop-node/src/watchers/classes/BaseWatcher.ts
```diff
@@ -36,7 +36,7 @@ class BaseWatcher extends EventEmitter implements IBaseWatcher {
   order: () => number = () => 0
   started: boolean = false
   pollIntervalSec: number = 10 * 1000
-  resyncIntervalSec: number = 10 * 60 * 1000
+  resyncIntervalSec: number = 60 * 1000
```
miguelmota, line 39: I think both `pollIntervalSec` and `resyncIntervalSec` should be renamed to use `Ms` instead of  `Sec` here since it's being multiplied by `1000`

packages/hop-node/src/watchers/classes/BaseWatcher.ts
```diff
@@ -47,6 +47,7 @@ class BaseWatcher extends EventEmitter implements IBaseWatcher {
   dryMode: boolean
   tag: string
   prefix: string
+  syncIndex: number
```
miguelmota, line 50: nit pick: the default assignment should be here instead of in the constructor

packages/hop-node/src/watchers/classes/BaseWatcher.ts
```diff
@@ -95,23 +97,20 @@ class BaseWatcher extends EventEmitter implements IBaseWatcher {
   }
 
   async preSyncHandler () {
-    this.logger.debug('syncing up events')
+    this.syncIndex++
```
miguelmota, line 100: i think the incrementer should be moved to the post handler before the wait (so it starts at 0)

miguelmota (commented):
PR looks great 👍. Left a few minor best practice suggestions.

packages/hop-node/src/watchers/classes/BaseWatcher.ts
shanefontaine: Fix 1b4fc481ea7398f99fbd7364d4ae3a6a38eaf4dc

packages/hop-node/src/watchers/classes/BaseWatcher.ts
shanefontaine: Fix 1b4fc481ea7398f99fbd7364d4ae3a6a38eaf4dc

packages/hop-node/src/watchers/classes/BaseWatcher.ts
shanefontaine: Fix 1b4fc481ea7398f99fbd7364d4ae3a6a38eaf4dc

PR #127, opened 2021-07-29
hop-node: Add syncer class

miguelmota:


shanefontaine:
Code LGTM. Some higher level questions:

1. This will force a full resync on the next run because the new cachekey uses the `SyncWatcher` tag for everything now, correct?
2. If (1) is true, will any of the DB dump logic need to be updated to look at this new DB instead of the individual watchers?
3. Since the syncs are no longer per-watcher, should we not start any poller until full sync is complete? We currently do the following in some watchers. I'd imagine this gets removed and something like it put somewhere in `BaseWatcher.ts`?

```
 const initialSyncCompleted = this.isAllSiblingWatchersInitialSyncCompleted()
    if (!initialSyncCompleted) {
      return false
    }
```

miguelmota:
1. Yes
2. No because the db's are not tied to specific watchers, so dumping transfers and roots will be as usual. The sync state db will have deprecated items however.
3. you're correct, i forgot about that piece. Will updated and request your review

miguelmota:
The update adds a `syncWatcher` reference to all the watchers and `isAllSiblingWatchersInitialSyncCompleted` on the watcher true only if all the sibling syncWatchers have completed

shanefontaine (approved):
LGTM 👍

PR #128, opened 2021-07-30
hop-node: remove unused code

shanefontaine:
@miguelmota This PR removes newly unused code because of recent code/sync optimizations. Code removal includes "too many blocks" error.

Three top level questions that relate to these and prior updates:
1. `ready` in the ContractBase class is set to `true` [upon initialization](https://github.com/hop-protocol/hop/blob/26c98128973985ab3caf3a572b51582e85d470fc/packages/hop-node/src/watchers/classes/ContractBase.ts#L13). If I understand that correctly, this class will be set to `ready` even before the completion of the constructor code. Does it make sense to initialize it to `false` and set it to `true` on the last line of the constructor?
2. Because constructors are synchronous, is it guaranteed the last line of the most base class (in this case, ContractBase), is truly the last thing to happen upon initialization?
3. If `this.bridgeDeployedBlockNumber` is `undefined` [here](https://github.com/hop-protocol/hop/blob/26c98128973985ab3caf3a572b51582e85d470fc/packages/hop-node/src/watchers/classes/Bridge.ts#L65), the process continues as if there is nothing wrong. This might happen if `bridgeDeployedBlockNumber` is not correctly set in the addresses package or the local address.json that the process is pointing to. Even though it runs with no errors, there are places in the code where we assume this is set. Does it make sense to perform an `if (!this.bridgeDeployedBlockNumber)` check after this is set? My only hesitation would be that there are many state vars that _could_ be checked as well.

### DIFF
--- a/packages/hop-node/src/watchers/SyncWatcher.ts
+++ b/packages/hop-node/src/watchers/SyncWatcher.ts
@@ -403,12 +403,7 @@ class SyncWatcher extends BaseWatcher {
           continue
         }
 
-        const eventRootHash = event.args.rootHash
         const eventDestinationChainId = Number(event.args.destinationChainId.toString())
-        const eventDbTransferRoot = await this.db.transferRoots.getByTransferRootHash(
-          eventRootHash
-        )
-
         const isSameChainId = eventDestinationChainId === destinationChainId
         if (endEvent && isSameChainId) {
           startEvent = event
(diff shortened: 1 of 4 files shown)


miguelmota:
1. class constructor are synchronous so it won't make a difference here. The ready flag was added because of the async call that was there before, so we don't really need the ready flag anymore since the bridge is now immediately ready. You may remove it and remove `waitTilReady` if you want.
2. The most base class (ContractBase) constructor will be executed first 
DeepinScreenshot_select-area_20210730142437

3. I think we should throw if it is not set and then set it the config so it's alway defined
```js
const deployedBlockNumber = config.tokens[this.tokenSymbol]?.[this.chainSlug]?.bridgeDeployedBlockNumber
if (!bridgeDeployedBlockNumber) {
}
this.bridgeDeployedBlockNumber = deployedBlockNumber
```

The PR itself looks great

miguelmota (approved):

PR #129, opened 2021-08-02
Feature/bond already settled transactions

shanefontaine:
**_Do not merge until tests are complete._**

This PR allows the system to bond transfers whose `transferRoots` were already committed/confirmed/settled.

* Marks a transfer as bondable/unbondable upon seeing `TransferSent` event.
* Does not bond if not enough credit is available

With these changes, transfer roots will be settled as many times as needed in order fully "complete" a root hash.


packages/hop-node/src/watchers/classes/Bridge.ts
```diff
@@ -31,6 +31,8 @@ export default class Bridge extends ContractBase {
   bridgeContract: Contract
   readProvider?: providers.Provider
   bridgeDeployedBlockNumber: number
+  minAmount: BigNumber
```
miguelmota, line 34: I suggest making these names more descriptive in the context of the bridge; ie `minBondwithdrawalAmount`

packages/hop-node/src/watchers/classes/Bridge.ts
```diff
@@ -67,6 +69,8 @@ export default class Bridge extends ContractBase {
       throw new Error('bridge deployed block number is required')
     }
     this.bridgeDeployedBlockNumber = config.tokens[this.tokenSymbol]?.[this.chainSlug]?.bridgeDeployedBlockNumber
+    this.minAmount = config?.bondWithdrawals?.[this.chainSlug]?.[this.tokenSymbol]?.min ?? 0
```
miguelmota, line 72: shouldn't these be converted to BigNumber types?

packages/hop-node/src/watchers/classes/Bridge.ts
shanefontaine: Resolved with ae73d900762faa1f5898955d07d840716155bc9b

packages/hop-node/src/watchers/classes/Bridge.ts
shanefontaine: Resolved with ae73d900762faa1f5898955d07d840716155bc9b

PR #131, opened 2021-08-03
Feature/gasboost

miguelmota:
do not merge yet


packages/hop-node/src/gasboost/GasBoostTransaction.ts
```diff
@@ -0,0 +1,297 @@
...
+    if (tx.gasPrice) {
+      this.gasPrice = BigNumber.from(tx.gasPrice.toString())
+    }
+    if (tx.gasLimit) {
+      this.gasLimit = BigNumber.from(tx.gasLimit.toString())
+    }
+  }
+
+  get hash ():string {
+    if (!this.txHash) {
```
shanefontaine, line 87: I believe this should be `if (this.txHash)`

packages/hop-node/src/gasboost/GasBoostTransaction.ts
```diff
@@ -0,0 +1,297 @@
...
+  }
+
+  private handleConfirmation (txHash: string) {
+    this.txHash = txHash
+    this.clearInflightTxs()
+    const receipt = this.getReceipt(txHash)
+    this.emit(State.Confirmed, receipt)
+  }
+
+  private async getReceipt (txHash: string) {
```
shanefontaine, line 206: Opinion: If I understand this correctly, I would expect `getReceipt` to return a receipt immediately. If this is waiting for a tx to complete, I would expect this to be called `waitForReceipt`.

Similar to ethers.js `provider.getTransactionReceipt`: "Returns the transaction receipt for hash or null if the transaction has not been mined."

packages/hop-node/src/gasboost/GasBoostTransaction.ts
```diff
@@ -0,0 +1,297 @@
...
+  gasLimit: string
+}
+
+class GasBoostTransaction extends EventEmitter implements providers.TransactionResponse {
+  started: boolean = false
+  pollMs: number = 10 * 1000
+  timeTilBoostMs: number = 1 * 60 * 1000
+  gasPriceMultiplier: number = 1.5
+  maxGasPriceGwei: BigNumber = BigNumber.from(500)
+  boostIndex: number = 0
```
shanefontaine, line 41: This appears to be unused

packages/hop-node/src/gasboost/GasBoostTransaction.ts
```diff
@@ -0,0 +1,297 @@
...
+      to: this.to,
+      data: this.data,
+      value: this.value,
+      nonce: this.nonce,
+      gasPrice,
+      gasLimit: this.gasLimit
+    })
+    this.gasPrice = tx.gasPrice
+    this.boostIndex++
+    this.emit(State.Boosted, tx)
```
shanefontaine, line 277: It does not appear as though anything handles this emission

shanefontaine (commented):
Nice implementation! Some higher level questions:

* How would this implementation handle a "dropped" transaction?
* If a transaction was "dropped" or not seen by a node, I think there is a case where a new, unrelated transaction can submitted with the same nonce. In this case, the DB would have a two transactions with the same nonce. How would this implementation handle that?
  * I think this case might be extremely, extremely rare. If so, maybe adding complexity to handle this "one off" case might not be a great idea

packages/hop-node/src/gasboost/GasBoostTransaction.ts
miguelmota: nice catch, thanks

packages/hop-node/src/gasboost/GasBoostTransaction.ts
miguelmota: this will be for outside subscribers, like for testing

packages/hop-node/src/gasboost/GasBoostTransaction.ts
miguelmota: true

PR #134, opened 2021-08-08
Feature/state changer

shanefontaine:
4 states -- can be in 1 at any given time
* 0 - dry mode off, pause mode off (normal)
* 1 - dry mode on, pause mode off
* 2 - dry mode off, pause mode on
* 3 - restart bonder

Pause mode means the bonder is on & syncing, but not polling.

### DIFF
--- a/packages/hop-node/src/watchers/classes/BaseWatcher.ts
+++ b/packages/hop-node/src/watchers/classes/BaseWatcher.ts
@@ -2,6 +2,7 @@ import L1Bridge from './L1Bridge'
 import L2Bridge from './L2Bridge'
 import Logger from 'src/logger'
 import SyncWatcher from '../SyncWatcher'
+import { Chain } from 'src/constants'
 import { Contract } from 'ethers'
 import { Db, getDbSet } from 'src/db'
 import { EventEmitter } from 'events'
@@ -174,6 +175,47 @@ class BaseWatcher extends EventEmitter implements IBaseWatcher {
 
   cacheKey (key: string) {
     return `${this.tag}:${key}`
+  }
+
+  handleStateSwitch = async () => {
+    if (!this.stateUpdateAddress) {
+      return
+    }
+
+    let state: number
+    try {
+      const l1ChainId = this.chainSlugToId(Chain.Ethereum)
+      const l1Bridge = this.getSiblingWatcherByChainId(l1ChainId).bridge
+      state = await l1Bridge.getStateUpdateStatus(this.stateUpdateAddress, this.bridge.chainId)
+    } catch (err) {
+      this.logger.log(`getStateUpdateStatus failed with ${err}`)
+      return
+    }
+
+    switch (state) {
+      case 0: {
+        this.setDryMode(false)
+        this.setPauseMode(false)
+        break
+      }
+      case 1: {
+        this.setDryMode(true)
+        this.setPauseMode(false)
+        break
+      }
+      case 2: {
+        this.setDryMode(false)
+        this.setPauseMode(true)
+        break
+      }
+      case 4: {
+        this.quit()
+      }
+    }
+  }
+
+  get isDryOrPauseMode () {
+    return this.dryMode || this.pauseMode
   }
 
   setDryMode (enabled: boolean) {
--- a/packages/hop-node/src/watchers/classes/BaseWatcher.ts
+++ b/packages/hop-node/src/watchers/classes/BaseWatcher.ts
@@ -208,7 +208,7 @@ class BaseWatcher extends EventEmitter implements IBaseWatcher {
         this.setPauseMode(true)
         break
       }
-      case 4: {
+      case 3: {
         this.quit()
       }
     }
--- a/packages/hop-node/src/watchers/classes/BaseWatcher.ts
+++ b/packages/hop-node/src/watchers/classes/BaseWatcher.ts
@@ -105,6 +105,9 @@ class BaseWatcher extends EventEmitter implements IBaseWatcher {
           this.notifier.error(`poll check error: ${err.message}`)
           console.trace()
         }
+      } else {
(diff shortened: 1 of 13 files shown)


packages/hop-node/src/watchers/classes/BaseWatcher.ts
```diff
@@ -181,6 +187,62 @@ class BaseWatcher extends EventEmitter implements IBaseWatcher {
...
+    try {
+      const l1ChainId = this.chainSlugToId(Chain.Ethereum)
+      const l1Bridge = this.getSiblingWatcherByChainId(l1ChainId).bridge
+      state = await l1Bridge.getStateUpdateStatus(this.stateUpdateAddress, this.bridge.chainId)
+    } catch (err) {
+      this.logger.log(`getStateUpdateStatus failed with ${err}`)
+      return
+    }
+
+    switch (state) {
```
miguelmota, line 205: This case statement section can replaced and simplified with an enum check


```js
enum State {
  Normal = 0,
  DryMode = 1,
  PauseMode = 2,
  Exit = 3
}

const state: State = Numer(await l1Bridge.getStateUpdateStatus(this.stateUpdateAddress, this.bridge.chainId))
this.setDryMode(state === State.DryMode)
this.setPauseMode(state === State.PauseMode)

if (state === State.Exit) {
  this.logger.error('exit mode enabled')
  this.quit()
}
```

packages/hop-node/src/watchers/classes/Bridge.ts
```diff
@@ -66,10 +68,15 @@ export default class Bridge extends ContractBase {
     }
     this.db = getDbSet(this.tokenSymbol)
     const bridgeDeployedBlockNumber = config.tokens[this.tokenSymbol]?.[this.chainSlug]?.bridgeDeployedBlockNumber
+    const l1CanonicalTokenAddress = config.tokens[this.tokenSymbol]?.[Chain.Ethereum]?.l1CanonicalToken
```
miguelmota, line 71: Out of curiouslity, why is `stateUpdateAddress` passed in via constructor config argument but `l1CanonicalTokenAddress` is being read directly from global config here? Should stateUpdateAddress be read the same way as well, so it's less code?

packages/hop-node/src/watchers/classes/Bridge.ts
```diff
@@ -66,10 +68,15 @@ export default class Bridge extends ContractBase {
...
     if (!bridgeDeployedBlockNumber) {
       throw new Error('bridge deployed block number is required')
     }
-    this.bridgeDeployedBlockNumber = config.tokens[this.tokenSymbol]?.[this.chainSlug]?.bridgeDeployedBlockNumber
+    if (!l1CanonicalTokenAddress) {
+      throw new Error('l1 token address is required')
```
miguelmota, line 76: suggestion: capitalize `L1` for readability

packages/hop-node/src/watchers/classes/Bridge.ts
```diff
@@ -527,6 +534,21 @@ export default class Bridge extends ContractBase {
...
+    const ethersInterface = new ethersUtils.Interface(abi)
+    const data = ethersInterface.encodeFunctionData(
+      'currentState', [this.l1CanonicalTokenAddress, chainId]
+    )
+    const tx = {
+      to: stateUpdateAddress,
+      data
+    }
+    const res: string = await this.contract.provider.call(tx)
+    return Number(res)
```
miguelmota, line 549: maybe a dumb question but would `res` here ever be a big number value that `Number` won't be a be to handle? (ie does it need to be wrapped in a big number type and then converted to Number)

packages/hop-node/src/watchers/classes/Bridge.ts
shanefontaine: Does `stateUpdateAddress` being read from global config imply that it lives in the `addresses` package?

packages/hop-node/src/watchers/classes/Bridge.ts
shanefontaine: Technically yes if we pass in a number that is outside of our options for state code. In practice, the result should only every be 1, 2, 3, or 4 (or 5, 6, etc. if we add more states). The `res` for 1, for example, looks like `0x0000000000000000000000000000000000000001`.

If `0xfff...` is the currents state, then I do think this would break. That should never happen but I'd be happy to add a check here (or on the contract) that would confirm this. LMK what you think.

packages/hop-node/src/watchers/classes/Bridge.ts
shanefontaine: Resolved with 46c229881e4ee54f3f6259e30a89cbed5fbead07

packages/hop-node/src/watchers/classes/BaseWatcher.ts
shanefontaine: Nice call. Resolved with 979427f8a53874ecbadc5887ab87cf17fc9699b6

packages/hop-node/src/watchers/classes/Bridge.ts
miguelmota: The global config contains things from addresses package but also things outside of it (eg like the min/max limits)

packages/hop-node/src/watchers/classes/Bridge.ts
miguelmota: The way it currently is works as expected, so we can leave as is

```
Number('0x0000000000000000000000000000000000000001') // 1
Number('0x0000000000000000000000000000000000000002') // 2
```

packages/hop-node/src/watchers/classes/Bridge.ts
shanefontaine: Resolved with f7551d37755cb17e805c105691b41a8520890963

PR #136, opened 2021-08-18
Feature/update bond transfer root watcher

shanefontaine:


packages/hop-node/src/watchers/BondTransferRootWatcher.ts
```diff
@@ -151,72 +121,34 @@ class BondTransferRootWatcher extends BaseWatcher {
...
       return
     }
 
-    const hasPositiveBalance = await l1Bridge.hasPositiveBalance()
-    if (!hasPositiveBalance) {
-      logger.warn('bonder requires positive balance to bond transfer root')
+    const availableCredit = await l1Bridge.getAvailableCredit()
+    const bondAmount = await l1Bridge.getBondForTransferAmount(totalAmount)
+    if (availableCredit.lt(bondAmount)) {
+      logger.warn(
```
miguelmota, line 147: would adding a slack notification be good to have here as well?

packages/hop-node/src/watchers/BondTransferRootWatcher.ts
```diff
@@ -44,90 +37,56 @@ class BondTransferRootWatcher extends BaseWatcher {
   }
 
   async pollHandler () {
+    if (!this.isL1) {
```
miguelmota, line 40: Since this watcher only requires L1 and doesn't use sibling watchers, does it make to sense to only initialize the L1 watcher and not the others in watchers.ts?

miguelmota (changes requested):

packages/hop-node/src/watchers/BondTransferRootWatcher.ts
```diff
@@ -136,10 +95,21 @@ class BondTransferRootWatcher extends BaseWatcher {
...
+      const timestamp = await this.bridge.getEventTimestamp(event)
+
       await this.db.transferRoots.update(transferRootHash, {
-        transferRootId,
         transferRootHash,
-        bonded: true
+        bonded: true,
+        bonder: sender,
+        bondTotalAmount: totalAmount,
+        bondTxHash: transactionHash,
```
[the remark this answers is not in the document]
shanefontaine: Yes. In the case where the syncer misses the event for some reason, we need a way to re-add the data to the DB so that the missed event is now "seen". This is similar to the logic we have in the BondWithdrawalWatcher [here](https://github.com/hop-protocol/hop/blob/develop/packages/hop-node/src/watchers/BondWithdrawalWatcher.ts#L101-L117).

I agree it should be DRYer, but it would take a slight architecture change because the watchers don't have access to the syncer functions. Ideally these lines would just be replaced with a `handleTransferRootBondedEvent()` call.

packages/hop-node/src/watchers/BondTransferRootWatcher.ts
shanefontaine: I think that makes a lot of sense. It looks like we have a similar setup for these watchers as well—would you imagine these watcher initializations are all handled in `watcher.ts` as well?

Screen Shot 2021-08-18 at 8 40 30 AM

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -329,10 +330,15 @@ class SyncWatcher extends BaseWatcher {
       )
       const blockNumber: number = event.blockNumber
 
+      const sourceChainSlug = this.chainIdToSlug(sourceChainId)
+      const skipChains: string[] = [Chain.xDai, Chain.Polygon]
+      const shouldBondTransferRoot = !skipChains.includes(sourceChainSlug)
```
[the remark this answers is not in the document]
shanefontaine: I like this. Fixed in d88e3e18b3c41f94bc0ca094df6864013e7eb9bf.

packages/hop-node/src/watchers/BondTransferRootWatcher.ts
```diff
@@ -226,31 +158,10 @@ class BondTransferRootWatcher extends BaseWatcher {
...
-        ) - this.bridge.formatUnits(debit)}, need ${this.bridge.formatUnits(
-          totalAmount
-        )}`
-      )
-      return
-    }
-
     logger.debug(
-      `bonding transfer root ${transferRootHash} on chain ${destinationChainId}`
+      `bonding transfer root ${transferRootHash} with dest chain ${destinationChainId}`
```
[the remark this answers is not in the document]
shanefontaine: Fixed in d88e3e18b3c41f94bc0ca094df6864013e7eb9bf.

packages/hop-node/src/watchers/BondTransferRootWatcher.ts
shanefontaine: Fixed in d88e3e18b3c41f94bc0ca094df6864013e7eb9bf.

PR #140, opened 2021-08-28
hop-node: SettleBondedWithdrawalsWatcher: Check by transfer root with rootSet events and unsettled bonded transfers

miguelmota:


packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -172,6 +173,44 @@ class TransferRootsDb extends BaseDb {
       )
     })
   }
+
+  async getUnsettledTransferRoots (
```
shanefontaine, line 177: Now that we have this in the TransferRootsDb, I believe we should remove `getUnsettledBondedWithdrawalTransfers()` from the TransfersDb.

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -172,6 +173,44 @@ class TransferRootsDb extends BaseDb {
...
+  ): Promise<TransferRoot[]> {
+    const transfers: TransferRoot[] = await this.getTransferRoots()
+    return transfers.filter(item => {
+      if (filter?.destinationChainId) {
+        if (filter.destinationChainId !== item.destinationChainId) {
+          return false
+        }
+      }
+
+      let rootSetTimestampOk = true
```
shanefontaine, line 188: I believe this timestamp was introduced to handle the case where a transfer from xDai -> L1 also commits the root hash. In this case, we were waiting x blocks for reorg, but the commit got sent to L1 before we were bonded the transfer. With our old software, we did not re-settle if a transfer got bonded after the first settlement.

We have updated the software in a few ways since this was introduced:
* We now resettled transferRoots if they should be resettled
* We now only wait 12 blocks on xDai instead of 24. This matches that of the Omni bridge and I believe fixes this issue

Given that, what are your thoughts on this piece of code? I don't _love_ it, but I think in a crazy edge case it could save us an L1 tx. On the downside, it makes settling and debugging slightly longer and more annoying. If we keep it, maybe we either cut the time in 1/2 or scope it to `if (item?.rootSetTimestamp && this.chainSlug === xDai)`. What are your thoughts on all that?

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -34,6 +34,7 @@ export type TransferRoot = {
   bondTransferRootId?: string
   challenged?: boolean
   challengeExpired?: boolean
+  allSettled?: boolean
```
shanefontaine, line 37: I believe we need to set in the event handlers. In the syncWatcher, `handleMultipleWithdrawalsSettledEvent()` loops through all transfer IDs and marks them all `withdrawalBondSettled` if they are settled. I think `allSettled` should be set in this function to `true` if all of those IDs are settled and `false` if not. This will guarantee the state for the initial sync. It also marks them as settled so the settle watcher does not keep trying to settle then updating the DB downstream.

shanefontaine (changes requested):

packages/hop-node/src/db/TransferRootsDb.ts
miguelmota: I like the latter suggestion to scope to xdai with a shorter time to prevent making extra L1 txs

packages/hop-node/src/db/TransferRootsDb.ts
miguelmota: Good idea, will add it to the sync handler as well

packages/hop-node/src/db/TransferRootsDb.ts
miguelmota: good idea, will remove it

packages/hop-node/src/watchers/SettleBondedWithdrawalWatcher.ts
```diff
@@ -60,133 +60,138 @@ class SettleBondedWithdrawalWatcher extends BaseWatcher {
...
+        // if all transfers have been settled that belong to a bonder
+        // then don't attempt to settle root with that bonder
+        // because there is nothing to settle anymore
+        const allSettledByBonder = dbTransfers.filter(
+          (dbTransfer: Transfer) => dbTransfer.withdrawalBonder === bonder
+        )
+          .every((dbTransfer: Transfer) =>
+            dbTransfer?.withdrawalBondSettled
+          )
+        if (!allSettledByBonder) {
```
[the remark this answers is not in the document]
miguelmota: you are correct

PR #144, opened 2021-09-02
hop-node: Check min bonder fee for bond withdrawal and attempt swaps

miguelmota:


packages/hop-node/src/watchers/classes/L2Bridge.ts
```diff
@@ -39,6 +43,19 @@ export default class L2Bridge extends Bridge {
             this.bridgeContract.signer
           )
           this.ammWrapper = new L2AmmWrapper(ammWrapperContract)
+
+          if (ammWrapperContract.exchangeAddress) {
+            ammWrapperContract
+              .exchangeAddress()
+              .then((address: string) => {
```
shanefontaine, line 50: do we need to add a wait so that this can complete successfully before `this.start = true`? Similar to the other watcher you updated yesterday?

miguelmota: It's updated now to use the addresses from package instead of having to do an onchain call so there won't be anything async in constructor anymore

PR #145, opened 2021-09-08
hop-node: Check max bond withdrawal amount for destination chain

miguelmota:


shanefontaine:
Recapping our discussion about this PR:

* the concept of `minBondWithdrawalAmount` and `maxBondWithdrawalAmount` should now be removed from the package.
* The available amount check [here](https://github.com/hop-protocol/hop/blob/develop/packages/hop-node/src/watchers/BondWithdrawalWatcher.ts#L123) should
  * Take into consideration additional amounts: `availableLiquidity > pendingAmount + (2 * transferAmount)`
  * Exponential backoff (7 days) for this available credit check
    * There are two goals of this is to replace the `maxBondWithdrawalAmount` by not bonding if the user sends a transaction whose amount far exceeds any credit the bonder will ever have.

miguelmota:
Closing. Replaced by https://github.com/hop-protocol/hop/pull/146

PR #146, opened 2021-09-10
Update max bond withdrawal check

miguelmota:


packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
```diff
@@ -121,13 +112,28 @@ class BondWithdrawalWatcher extends BaseWatcher {
       return
     }
 
-    const availableCredit = await destBridge.getAvailableCredit()
+    let availableCredit = await destBridge.getAvailableCredit()
+    const includePendingAmount = ([Chain.Optimism, Chain.Arbitrum] as string[]).includes(destinationChain)
+    if (includePendingAmount) {
+      const pendingAmount = await sourceL2Bridge.getPendingAmountForChainId(destinationChainId)
+      availableCredit = availableCredit.sub(pendingAmount).sub(amount)
```
shanefontaine, line 119: 1. Why is the `2*transferAmount` not included here?
2. I thought we were going to get this from an sdk call so that this piece of code was only defined once but used on both the FE/SDK and here so we don't get out of sync. I might have misunderstood.

packages/sdk/src/HopBridge.ts
```diff
@@ -454,11 +454,30 @@ class HopBridge extends Base {
...
+        includePendingAmount = true
+      }
+    }
+
+    if (includePendingAmount) {
+      const bridgeContract = await this.getBridgeContract(sourceChain)
+      const pendingAmount = await bridgeContract.pendingAmountForChainId(
+        destinationChain.chainId
+      )
+      requiredLiquidity = requiredLiquidity.add(pendingAmount).add(hTokenAmount) // account for transfer root bonds
```
shanefontaine, line 473: Same question about the `2x` as above.

packages/sdk/src/HopBridge.ts
```diff
@@ -575,10 +594,17 @@ class HopBridge extends Base {
     sourceChain = this.toChainModel(sourceChain)
     destinationChain = this.toChainModel(destinationChain)
 
-    const requireFee =
-      destinationChain?.isL1 ||
-      (!sourceChain?.equals(Chain.Ethereum) &&
-        destinationChain?.equals(Chain.Optimism))
+    let requireFee = false
+    if (destinationChain?.isL1) {
```
shanefontaine, line 598: The tx fees on polygon and xDai are negligible (< $0.01 and lower). Because of this, including those fees in the calcs would effectively make no difference, especially given that there is already (a) amm slippage and (b) a bonder fee.

Does it make sense to remove this conditional completely in order to simplify and charge a tx fee on every network?

packages/sdk/src/HopBridge.ts
```diff
@@ -745,33 +771,41 @@ class HopBridge extends Base {
...
+    let minusBuffer = false
+    if (destinationChain.equals(Chain.Ethereum)) {
+      if (
+        sourceChain.equals(Chain.Optimism) ||
+        sourceChain.equals(Chain.Arbitrum)
+      ) {
+        minusBuffer = true
       }
     }
+    if (minusBuffer) {
```
shanefontaine, line 792: I was under the impression we were removing this in favor of 2x. If so, the whole concept of `UnbondedRootsBufferAmount` should be removed.

packages/sdk/src/constants/constants.ts
```diff
@@ -27,4 +27,4 @@ export const LpFeeBps = '4'
 export const GasPriceMultiplier = '1.6'
 export const L2ToL1BonderFeeBps = '18'
 export const L2ToL2BonderFeeBps = '30'
-export const UnbondedRootsBuffer = '50000'
+export const UnbondedRootsBufferAmount = '50000'
```
shanefontaine, line 30: Potentially remove based on prior comments.

shanefontaine (commented):
Back to you @miguelmota

packages/sdk/src/HopBridge.ts
miguelmota: It's already being double counted because  `requiredLiquidity` is initialized as `hTokenAmount`

packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
```diff
@@ -121,13 +112,28 @@ class BondWithdrawalWatcher extends BaseWatcher {
...
           availableCredit
         )}, need ${this.bridge.formatUnits(amount)}`
       )
+      let { withdrawalBondBackoffIndex } = await this.db.transfers.getByTransferId(transferId)
+      if (!withdrawalBondBackoffIndex) {
+        withdrawalBondBackoffIndex = 0
+      }
+      withdrawalBondBackoffIndex++
+      await this.db.transfers.update(transferId, {
+        bondWithdrawalAttemptedAt: Date.now(),
```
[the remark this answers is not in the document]
miguelmota: That sounds good. Can be done in a separate commit

packages/hop-node/src/db/TransfersDb.ts
```diff
@@ -127,18 +127,19 @@ class TransfersDb extends BaseDb {
       }
 
       let timestampOk = true
-      if (item.sentBondWithdrawalTxAt) {
-        if (item.withdrawalBondTxError === TxError.BonderFeeTooLow) {
+      if (item.bondWithdrawalAttemptedAt) {
+        const checkBackoff = ([TxError.BonderFeeTooLow, TxError.NotEnoughLiquidity] as string[]).includes(item.withdrawalBondTxError)
```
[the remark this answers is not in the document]
miguelmota: Fixed

packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
```diff
@@ -121,13 +112,28 @@ class BondWithdrawalWatcher extends BaseWatcher {
       return
     }
 
-    const availableCredit = await destBridge.getAvailableCredit()
+    let availableCredit = await destBridge.getAvailableCredit()
+    const includePendingAmount = ([Chain.Optimism, Chain.Arbitrum] as string[]).includes(destinationChain)
```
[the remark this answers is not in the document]
miguelmota: Fixed

packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
miguelmota: 1. Fixed
2. tried importing the sdk but didn't seem right because it would be basically reinitializing the contracts and providers which can complicate things (especially since the sdk doesn't have retry decorators) for this single check. If the sdk was more robust then it would make more sense I think.

packages/sdk/src/HopBridge.ts
miguelmota: Sounds good; it's removed so all chains have same fee requirement here

packages/sdk/src/HopBridge.ts
miguelmota: Fixed

packages/sdk/src/constants/constants.ts
miguelmota: Fixed

packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
```diff
@@ -121,13 +112,28 @@ class BondWithdrawalWatcher extends BaseWatcher {
       return
     }
 
-    const availableCredit = await destBridge.getAvailableCredit()
+    let availableCredit = await destBridge.getAvailableCredit()
+    const includePendingAmount = [Chain.Optimism, Chain.Arbitrum].includes(destinationChain)
```
shanefontaine, line 116: This pendingAmount check should only apply when the destination chain is L1 and the source chain is a chain whose transfer root is being bonded. This is because, for example, a bonding of a withrdawal from o -> e would require `100` credit and then the bonding of that transfer root would require another `100` of credit, meaning the bonder would need `200` credit for a simple bond withdrawal of `100`. This double counting will be alleviated in v2 with `bondTransferRootAndSettle`.

Because of that, this we need to check here if source = opt/arb and destination = eth

Sorry, I missed this the first time around.

Edit: it looks like the check is correct in the SDK, so only this would need to be updated.

shanefontaine (commented):
@miguelmota back to you

packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
shanefontaine: 2. That SGTM & I agree w/ your analysis

packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
miguelmota: you're completely right. [Fixed](https://github.com/hop-protocol/hop/pull/146/files#diff-0c3ab79e1932de7478f738aa7851502fe92140eb706c6d86cb29b6dd831c94aeR117)

PR #147, opened 2021-09-11
hop-node: GasBoostSigner: Set maxFeePerGas and maxPriorityFeePerGas for Ethereum transactions

miguelmota:


shanefontaine:
Changes look good but I think we should discuss this PR on a call before merge so that we are on the same page -- specifically for debugging purposes in non-happy path cases.

Some top level notes. Some of these may be my lack of understanding of 1559.
* With this change, are we trying to maintin support for both type 0 and type 2 txs? Does it make sense to simplify by removing support for type 0 txs?
* Does the concept of `gasPrice` exist in type 2 txs?
* [This line](https://github.com/hop-protocol/hop/pull/147/files#diff-be5791e55b6353175456441175c8d86f188ce21871bbf9571f7e0d8de89a3351R537) does not make sense to me (might be 1559 misunderstanding). I wouldn't expect to set a variable named `gasPrice` equal to `gasFeeData.maxFeePerGas`. I don't have a good enough understanding to make a suggestion, but if `gasPrice` and `maxFeePerGas` are truly interchangeable, then I'd imagine that is reflected everywhere in the system.
* We have a lot of terms for gas price stuff (not gas limits). This might be required because of 1559, but do we think any of these can be simplified, combined, or removed? I'll list them below:
  * `minPriorityFeePerGas`
  * `maxPriorityFeePerGas`
  * `priorityFeePerGasCap`
  * `maxFeePerGas`
  * `gasCost`
  * `maxGasPrice`
  * `gasPrice`
    * My very loose opinions are that (1) `gasCost` should be renamed to `txCost` _just_ to remove the number of variables having to do with `gas` (less mental hoops to jump through), (2) `gasPrice` should be removed _if_ we are removing support for type 0 txs, (3) _if we keep type 0_, change `maxGasPrice` to `gasPriceCap` for consistency.

miguelmota:
> With this change, are we trying to maintain support for both type 0 and type 2 txs? 

Yes, because chains such as xDai and Polygon don't support EIP1559. Ethers `getFeeData()` returns null for `maxFeePerGas` and `maxPriorityFeePerGas` and throws an `INVALID_ARGUMENT` when they are used. These chains still require only using `gasPrice`

> Does it make sense to simplify by removing support for type 0 txs?

 No because xDai and Polygon only support type 0 txs

> Does the concept of gasPrice exist in type 2 txs?

No but it is maintained for chains such as xDai and Polygon that only support type 0 txs

> This line does not make sense to me (might be 1559 misunderstanding).

The gas cost (total ETH spent in tx fee) calculation is `gasLimit * gasPrice` for type 0 tx  or `gasLimit * maxFeePerGas` for type 2 tx which seems right based on existing txs on etherscan, but let me if you don't think this the right way

DeepinScreenshot_select-area_20210911130625


>  This might be required because of 1559, but do we think any of these can be simplified, combined, or removed?

Maybe but these are the reasons they exist:

- `minPriorityFeePerGas` - we use this priorityFeePerGas or the ethers suggestions. Which ever one is greater, as discussed yesterday in dev slack channel
- `maxPriorityFeePerGas` -  type 2 required property
- `priorityFeePerGasCap` - this the max we'll keep bumping `maxPriorityFeePerGas` to in type 2 txs. Since `maxPriorityFeePerGas` is already a type 2 argument, it uses the term cap instead
- `maxFeePerGas` - type 2 required property
- `gasPrice` - type 0 required property
- `maxGasPrice` - the max we'll keep bumping gasPrice in type 0 txs
- `gasCost` - total estimated gas cost, used for checking if account has enough ETH to execute transaction

packages/hop-node/src/gasboost/GasBoostTransaction.ts
```diff
@@ -355,74 +470,60 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
   }
 
   private async boost (item: InflightItem) {
-    const gasPrice = await this.getBumpedGasPrice()
-    const maxGasPrice = parseUnits(this.maxGasPriceGwei.toString(), 9)
-    if (gasPrice.gt(maxGasPrice)) {
+    const gasFeeData = await this.getBumpedGasFeeData()
+    const maxGasPrice = this.parseGwei(this.maxGasPriceGwei)
+    const priorityFeePerGasCap = this.parseGwei(this.priorityFeePerGasCap)
```
shanefontaine, line 475: It looks like `this.priorityFeePerGasCap` is not set in the constructor or variable instantiation. If it is `undefined`, I believe `this.parseGwei()` would fail. On https://playground.ethers.org/

```
parseUnits(undefined, 9)
"value must be a string (argument=\"value\", value=undefined, code=INVALID_ARGUMENT, version=units/5.4.0)"
```

Is it expected that this is always set with the `setter`? If not, maybe we set this in the constructor

packages/hop-node/src/gasboost/GasBoostTransaction.ts
```diff
@@ -355,74 +470,60 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
...
-  private async _sendTransaction (gasPrice: BigNumber):Promise<providers.TransactionResponse> {
+  private async _sendTransaction (gasFeeData: Partial<GasFeeData>):Promise<providers.TransactionResponse> {
     const maxRetries = 10
     let i = 0
     while (true) {
       i++
       try {
         if (i > 1) {
-          gasPrice = await this.getBumpedGasPrice(this.gasPriceMultiplier * i)
+          gasFeeData = await this.getBumpedGasFeeData(this.gasPriceMultiplier * i)
```
shanefontaine, line 504: IIUC, optimism will not bump here. If that is the case, then the `gasFeeData` will use the same `gasPrice` or `maxPriorityFee`. This, I believe, would cause the node to throw an error (as it expects 1.1x on each retry).

If that is all true, then I think we should somehow not bump at all on Optimism. I _think_ this is alright, as optimism txs should go through very quickly, but I believe the code should reflect that we are not re-boosting optimism so that we don't leave it up to a weird edge case.

I may be misunderstanding though so please LMK!

shanefontaine (commented):
@miguelmota can I get your opinion on these comments. Thx!

packages/hop-node/src/gasboost/GasBoostTransaction.ts
miguelmota: it's definitely being set in the constructor

DeepinScreenshot_select-area_20210914141526

constant 

DeepinScreenshot_select-area_20210914141542

shanefontaine: Oversight on my part. You are correct -- thx for pointing that out. Will resolve this now.

packages/hop-node/src/gasboost/GasBoostTransaction.ts
miguelmota: for optimism, the gas price is not bumped (always uses market gas price)

DeepinScreenshot_select-area_20210914141749

DeepinScreenshot_select-area_20210914141732

the loop you're looking at is specific to when sending the transaction and it throws an "already known error" because a really low gas price was used and the tx with same nonce is still mempool (so it tries again with a higher pas price). It's an edge case for xdai and polygon but wouldn't expect it to occur on optimism or arbitrum because the transaction is rejected or accepted immediately

DeepinScreenshot_select-area_20210914142149

shanefontaine: Ok nice I didn't see the `isAlreadyKnown` -- makes sense and SGTM.

PR #148, opened 2021-09-14
hop-node: Update pending amount check for bondable chains

miguelmota:


packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
```diff
@@ -114,10 +115,18 @@ class BondWithdrawalWatcher extends BaseWatcher {
...
 
     let availableCredit = await destBridge.getAvailableCredit()
-    const includePendingAmount = destinationChain === Chain.Ethereum && [Chain.Optimism, Chain.Arbitrum].includes(sourceChain)
+    const includePendingAmount = destinationChain === Chain.Ethereum && bondableChains.includes(sourceChain)
     if (includePendingAmount) {
-      const pendingAmount = await sourceL2Bridge.getPendingAmountForChainId(destinationChainId)
-      availableCredit = availableCredit.sub(pendingAmount).sub((amount).mul(2))
+      let pendingAmounts = BigNumber.from(0)
+      for (const chain of bondableChains) {
+        const bridge = this.getSiblingWatcherByChainSlug(chain)
```
shanefontaine, line 122: I believe this needs a `.bridge`. Probably a `.bridge as L2Bridge` since the pending amount check will only be on the L2 Bridge.

shanefontaine (changes requested):
This update will also need to be reflected in the SDK in the same places it was in [this](https://github.com/hop-protocol/hop/pull/146) PR, I believe.

@miguelmota back to you

packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
miguelmota: good catch, fixed

PR #149, opened 2021-09-16
hop-node: GasBoostSigner: Use market price or max, which ever one is lesser

miguelmota:


packages/hop-node/src/gasboost/GasBoostTransaction.ts
```diff
@@ -473,7 +476,7 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
     const gasFeeData = await this.getBumpedGasFeeData()
     const maxGasPrice = this.parseGwei(this.maxGasPriceGwei)
     const priorityFeePerGasCap = this.parseGwei(this.priorityFeePerGasCap)
-    const isMaxReached = gasFeeData.gasPrice?.gt(maxGasPrice) || gasFeeData.maxPriorityFeePerGas?.gt(priorityFeePerGasCap)
+    const isMaxReached = this.gasPrice?.gte(maxGasPrice) || this.maxPriorityFeePerGas?.gte(priorityFeePerGasCap)
```
shanefontaine, line 479: This looks good. To be clear, was this changed because the old code had a bug, correct?

packages/hop-node/src/gasboost/GasBoostTransaction.ts
```diff
@@ -301,31 +302,33 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
...
+      return BNMin(marketGasPrice, maxGasPrice)
     }
     const prevGasPrice = this.gasPrice || marketGasPrice
     const bumpedGasPrice = getBumpedGasPrice(prevGasPrice, multiplier)
     if (!this.compareMarketGasPrice) {
-      return bumpedGasPrice
+      return BNMin(bumpedGasPrice, maxGasPrice)
     }
-    return BNMax(marketGasPrice, bumpedGasPrice)
+    return BNMin(BNMax(marketGasPrice, bumpedGasPrice), maxGasPrice)
```
shanefontaine, line 315: I think there is an edge case here where we return the `maxGasPrice` but that value is w/in 10% of the old value. For example, we send a tx at 490 Gwei, boost it, and then the boost tries 500 Gwei, but the client returns it because 500 is not > 10% of 490.

I _think_ we can just do a 10% check before returning it and leave it at 490, for example, if if is <10% of the max. Worth thinking through because I haven't thought of all the implications of that.

packages/hop-node/src/gasboost/GasBoostTransaction.ts
```diff
@@ -301,31 +302,33 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
...
     const prevMaxPriorityFeePerGas = this.maxPriorityFeePerGas || marketMaxPriorityFeePerGas
     const minPriorityFeePerGas = this.parseGwei(this.minPriorityFeePerGas)
     let bumpedMaxPriorityFeePerGas = getBumpedBN(prevMaxPriorityFeePerGas, multiplier)
     bumpedMaxPriorityFeePerGas = BNMax(minPriorityFeePerGas, bumpedMaxPriorityFeePerGas)
     if (!this.compareMarketGasPrice) {
-      return bumpedMaxPriorityFeePerGas
+      return BNMin(bumpedMaxPriorityFeePerGas, priorityFeePerGasCap)
     }
-    return BNMax(marketMaxPriorityFeePerGas, bumpedMaxPriorityFeePerGas)
+    return BNMin(BNMax(marketMaxPriorityFeePerGas, bumpedMaxPriorityFeePerGas), priorityFeePerGasCap)
```
shanefontaine, line 331: Same as above.

Do you know if the nodes impose a `maxPrioirtyGasFee` increase minimum (like the 10% `gasPrice` minimum)?

shanefontaine (changes requested):
@miguelmota back to you

packages/hop-node/src/gasboost/GasBoostTransaction.ts
miguelmota: I believe it does

https://twitter.com/TimBeiko/status/1413541934719770628?s=20

packages/hop-node/src/gasboost/GasBoostTransaction.ts
miguelmota: old code:
return bumped gas fee data and check max. If it's below max, then then submit transaction and set `this.` properties. The value of getBumpedGasFeeData was not capped so this check was expected and not a bug

new code:
return bumped gas fee data with cap included. It's updated to check `this.` instead now because getBumpedGasFeeData will always returned the capped fee so it won't be able to get past the max fee check here if it's checking the returned value and therefore not be able to send tx with bumped gas fee. So it checks `this.` on the next time it has to boost to see if it already sent a tx with the max gas price

packages/hop-node/src/gasboost/GasBoostTransaction.ts
miguelmota: I think we can do if `maxGasPrice - lastGasPriceUsd < 10%` then use use `lastGasPriceUsd + 10%` which would put it over a little maxGasPrice (as expected) and no longer be boosted after that (as expected)

miguelmota: (no rush) let me know what you think

https://github.com/hop-protocol/hop/pull/149/commits/12e6b2e540f5ea4411febc3caa26040178d98602

packages/hop-node/src/gasboost/GasBoostTransaction.ts
shanefontaine: See comment above about this change.

packages/hop-node/src/gasboost/GasBoostTransaction.ts
shanefontaine: I believe I had a bad assumption about this code when I wrote this comment and, if I now understand it correctly, your old code was good and this latest commit should be reverted.

My assumption was that a boosted transaction would be attempted no matter what. So, for example, I was under the impression that if we sent a 490 GWei tx, then boosted it, we would cap it at 500 Gwei, which **we would then submit on-chain**. I see this is wrong—with your old code we would just throw away the tx instead of attempt the 500 Gwei.

I think there are 2 action items:

1. If you agree that my assumption was bad, then I think your code before 12e6b2e540f5ea4411febc3caa26040178d98602 was good and that commit should be reverted. The concept of `maxGasPrice` should not exist in `getBumpedGasPrice()`.

2. Based on this, I think the following line should go back to what it used to be. The `max` check should be based on `gasPrice` that is _about_ to be sent, not the one previously sent.

`const isMaxReached = gasFeeData.gasPrice?.gt(maxGasPrice) || gasFeeData.maxPriorityFeePerGas?.gt(priorityFeePerGasCap)`

Sorry for the miscommunication

---

Just so we are on the same page:

> which would put it over a little maxGasPrice (as expected)

I would never expect any transaction to go on chain if it is > than the max (500Gwei). I believe by this line you mean that we would allow a single tx over 500 Gwei if it was w/in the boost. I do not think that should be the case. Let me know if you agree.

miguelmota: I think I follow and agree but maxGasPrice was added to getBumpedGasPrice() so that the initial tx send didn't use a gas price over maxGasPrice. Are you suggesting that clamping to maxGasPrice should happen outside of getBumpedGasPrice() for initial tx send then?

miguelmota: @shanefontaine

shanefontaine: I believe I understand. I believe that the `getBumped...()` functions should still not care about the `maxGasPrice`. 

IIUC, the changes made to `getBumpedGasPrice()` and `getBumpedMaxPriorityFeePerGas()` were to handle the initial boost only. I don't think that should be the case and I think that these functions should care about the max. If we are handling the first send only here, I'd imagine that logic would live in a place where only the first send happens, which I believe is `send()`. The code would be:

```
function send() {
  ...
  if (gasFeeData.gasPrice > max) {
    gasFeeData = getCappedGasFeeData(gasFeeData) // this function just returns the max `gasPrice` or type2 equivalent
  }
  const tx = await this._sendTransaction(gasFeeData)
  ...
}
```

I think that the `getBumped...()` functions should still not care about the max. If I pass in `x` to one of these functions, I'd expect to get a bumped `x` and not a bumped or capped `x`. I would think the capping happens independently of the bumping.

miguelmota:
@shanefontaine give a quick thumbs up/down ([latest commit](https://github.com/hop-protocol/hop/pull/149/commits/350513d182718dbcedfdfc56fc334cecadb1f3a1))

shanefontaine:
LGTM

PR #151, opened 2021-09-17
Feature/balances stats

shanefontaine:
@miguelmota Can I get a review please. It appears to work and the end result is something like this.

Screen Shot 2021-09-17 at 5 27 51 AM

 I think there are a few optimizations/best practices that I can do better with. Specifically:

* How to not use an arbitrary token [here](https://github.com/hop-protocol/hop/compare/feature/balances-stats?expand=1#diff-80ab0f08ae45d91eb42b384c92869fa289b01be5e0924a2bfd2083bb6734c3afR290)
* I wouldn't expect this to reload unless I refresh the page. Should I remove [this](https://github.com/hop-protocol/hop/compare/feature/balances-stats?expand=1#diff-80ab0f08ae45d91eb42b384c92869fa289b01be5e0924a2bfd2083bb6734c3afR390)?
* All virtual amounts are on L1. Is there a better way to do that without hardcoding `ethereum [here](https://github.com/hop-protocol/hop/compare/feature/balances-stats?expand=1#diff-80ab0f08ae45d91eb42b384c92869fa289b01be5e0924a2bfd2083bb6734c3afR346)?
* The SDK in general. There are a few things.
  * It is unclear if passing `chain` into each of the calls is expected.
  * The types. Specifically when to use `number` vs `BigNumber` vs `BigNumberish`
  * All of the calls I added are to the L1_Bridge. Is this the appropriate place/file for these calls?

### DIFF
--- a/packages/sdk/src/HopBridge.ts
+++ b/packages/sdk/src/HopBridge.ts
@@ -11,7 +11,7 @@ import {
   l2AmmWrapperAbi,
   wethAbi
 } from '@hop-protocol/core/abi'
-import { TChain, TToken, TAmount, TProvider } from './types'
+import { TChain, TToken, TAmount, TProvider, TTime, TTimeSlot } from './types'
 import Base, { ChainProviders } from './Base'
 import AMM from './AMM'
 import _version from './version'
@@ -1142,16 +1142,12 @@ class HopBridge extends Base {
   /**
    * @readonly
    * @desc The time slot for the current time.
-   * @param {Object} chain - Chain model.
-   * @param {Number} time - Unix timestamp (in seconds) to get the time slot.
+   * @param {Object} time - Unix timestamp (in seconds) to get the time slot.
    * @returns {Object} Time slot for the given time as BigNumber.
    */
-  public async getTimeSlot(
-    chain: TChain,
-    time: number
-  ): Promise<BigNumber> {
-    chain = this.toChainModel(chain)
-    const bridge = await this.getBridgeContract(chain)
+  public async getTimeSlot(time: TTime): Promise<BigNumber> {
+    const bridge = await this.getL1Bridge()
+    time = BigNumber.from(time.toString())
 
     return bridge.getTimeSlot(time)
   }
@@ -1159,12 +1155,10 @@ class HopBridge extends Base {
   /**
    * @readonly
    * @desc The challenge period.
-   * @param {Object} chain - Chain model.
    * @returns {Object} The challenge period for the bridge as BigNumber.
    */
-  public async challengePeriod(chain: TChain): Promise<BigNumber> {
-    chain = this.toChainModel(chain)
-    const bridge = await this.getBridgeContract(chain)
+  public async challengePeriod(): Promise<BigNumber> {
+    const bridge = await this.getL1Bridge()
 
     return bridge.challengePeriod()
   }
@@ -1172,12 +1166,10 @@ class HopBridge extends Base {
   /**
    * @readonly
    * @desc The size of the time slots.
-   * @param {Object} chain - Chain model.
    * @returns {Object} The size of the time slots for the bridge as BigNumber.
    */
-  public async timeSlotSize(chain: TChain): Promise<BigNumber> {
-    chain = this.toChainModel(chain)
-    const bridge = await this.getBridgeContract(chain)
+  public async timeSlotSize(): Promise<BigNumber> {
+    const bridge = await this.getL1Bridge()
 
     return bridge.TIME_SLOT_SIZE()
   }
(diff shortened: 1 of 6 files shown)


miguelmota:
> How to not use an arbitrary token here

The bridge is only being used here to get the eth balance, so it's best to use the provider only I think

> I wouldn't expect this to reload unless I refresh the page. Should I remove this?

Yes I think so because pendingAmounts isn't being used in the useEffect. The array should only contain values that are used in the useEffect to trigger an update

> All virtual amounts are on L1. Is there a better way to do that without hardcoding `ethereum here?

I suggest removing the chain argument and using ethereum chain value directly in contract getter in method

```js
const bridge = await this.getBridgeContract(Chain.Ethereum)
```

> It is unclear if passing chain into each of the calls is expected.

It's only required if the method needs to know what chain to talk to. No need for chain argument if the method only interacts with the same chain all the time

> The types. Specifically when to use number vs BigNumber vs BigNumberish

General rule is make the argument type big numberish (`amount: TAmount`) and then convert to bignumber is the method (`amount = BigNumber.from(amount.toString())`)

> All of the calls I added are to the L1_Bridge. Is this the appropriate place/file for these calls?

I think it's fine placing them in the `HopBridge` classs. Down the line there can be an `L1Bridge` class that the hopbridge class can extend but for now it's not necessary

---

Reviewing the rest of the code

packages/sdk/src/HopBridge.ts
```diff
@@ -1139,6 +1139,68 @@ class HopBridge extends Base {
...
+   * @param {Object} chain - Chain model.
+   * @param {Number} time - Unix timestamp (in seconds) to get the time slot.
+   * @returns {Object} Time slot for the given time as BigNumber.
+   */
+  public async getTimeSlot(
+    chain: TChain,
+    time: number
+  ): Promise<BigNumber> {
+    chain = this.toChainModel(chain)
+    const bridge = await this.getBridgeContract(chain)
```
miguelmota, line 1154: Remove `chain` argument and use `Chain.Ethereum` or `getL1Bridge` if the chain is always ethereum

packages/sdk/src/HopBridge.ts
```diff
@@ -1139,6 +1139,68 @@ class HopBridge extends Base {
...
+
+  /**
+   * @readonly
+   * @desc The challenge period.
+   * @param {Object} chain - Chain model.
+   * @returns {Object} The challenge period for the bridge as BigNumber.
+   */
+  public async challengePeriod(chain: TChain): Promise<BigNumber> {
+    chain = this.toChainModel(chain)
+    const bridge = await this.getBridgeContract(chain)
```
miguelmota, line 1167: Remove `chain` argument and use `Chain.Ethereum` or `getL1Bridge` if the chain is always ethereum

packages/sdk/src/HopBridge.ts
```diff
@@ -1139,6 +1139,68 @@ class HopBridge extends Base {
...
+
+  /**
+   * @readonly
+   * @desc The size of the time slots.
+   * @param {Object} chain - Chain model.
+   * @returns {Object} The size of the time slots for the bridge as BigNumber.
+   */
+  public async timeSlotSize(chain: TChain): Promise<BigNumber> {
+    chain = this.toChainModel(chain)
+    const bridge = await this.getBridgeContract(chain)
```
miguelmota, line 1180: Remove `chain` argument and use `Chain.Ethereum` or `getL1Bridge` if the chain is always ethereum

packages/sdk/src/HopBridge.ts
```diff
@@ -1139,6 +1139,68 @@ class HopBridge extends Base {
...
+   * @desc The amount bonded for a time slot for a bonder.
+   * @param {Object} chain - Chain model.
+   * @param {Number} timeSlot - Time slot to get.
+   * @param {String} bonder - Address of the bonder to check.
+   * @returns {Object} Amount bonded for the bonder for the given time slot as BigNumber.
+   */
+  public async timeSlotToAmountBonded(
+    chain: TChain,
+    timeSlot: number,
+    bonder: string
```
miguelmota, line 1196: Set default bonder address 


```
bonder: string = this.getBonderAddress(this.tokenSymbol)
```

packages/sdk/src/HopBridge.ts
```diff
@@ -1139,6 +1139,68 @@ class HopBridge extends Base {
...
+  ): Promise<BigNumber> {
+    chain = this.toChainModel(chain)
+    const bridge = await this.getBridgeContract(chain)
```
miguelmota, line 1199: Remove `chain` argument and use `Chain.Ethereum` or `getL1Bridge` if the chain is always ethereum

packages/sdk/src/HopBridge.ts
shanefontaine: Good call - updated

packages/sdk/src/HopBridge.ts
shanefontaine: Good call - updated

packages/sdk/src/HopBridge.ts
shanefontaine: Good call - updated

shanefontaine:
Thanks for the review. All of it made sense. I updated all of it except for the balances from the provider. I'll add that at a later time.

PR #152, opened 2021-09-17
frontend/sdk: Update available liquidity check

miguelmota:


packages/sdk/src/HopBridge.ts
```diff
@@ -801,7 +774,47 @@ class HopBridge extends Base {
...
-    const availableLiquidity = credit.sub(debit)
+    let availableLiquidity = credit.sub(debit)
+
+    const isBondableChain = bondableChains.includes(sourceChain.slug)
+    const includePendingAmount =
+      isBondableChain && destinationChain.equals(Chain.Ethereum)
+
+    if (includePendingAmount) {
+      const bridgeContract = await this.getBridgeContract(sourceChain)
+      const pendingAmount = await bridgeContract.pendingAmountForChainId(
```
shanefontaine, line 785: IIUC this can be removed.

miguelmota: remove the pending amounts subtraction?

miguelmota: my understanding was having `((availableLiquidity - pendingAmounts) / 2) - 50k`

packages/sdk/src/HopBridge.ts
```diff
@@ -457,39 +458,11 @@ class HopBridge extends Base {
...
-        .add(pendingAmounts)
-        .add(hTokenAmount) // account for transfer root bonds
-    }
-
     return {
       amountOut,
       rate,
       priceImpact,
-      requiredLiquidity,
+      requiredLiquidity: hTokenAmount,
```
shanefontaine, line 465: Will this affect [this line](https://github.com/hop-protocol/hop/blob/8eee9b4c3c594ea65a45489984f128e7d830ba9b/packages/frontend/src/pages/Send/Send.tsx#L431) downstream?

packages/sdk/src/HopBridge.ts
shanefontaine: Maybe I'm not seeing it correctly, but it looks like this line 785 is not doing anything?

The equation you said is on line 812.

miguelmota: sorry i see the duplication now, will remove

packages/sdk/src/HopBridge.ts
```diff
@@ -801,7 +774,47 @@ class HopBridge extends Base {
...
+        token.canonicalSymbol
+      )
+      const tokenPriceBn = parseUnits(tokenPrice.toString(), token.decimals)
+      const bufferAmountBn = parseUnits(PendingAmountBuffer, token.decimals)
+      const precision = parseUnits('1', token.decimals)
+      const bufferAmountTokensBn = bufferAmountBn
+        .div(tokenPriceBn)
+        .mul(precision)
+
+      availableLiquidity = availableLiquidity
```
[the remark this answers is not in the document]
miguelmota: you're correct, should be sub

packages/sdk/src/HopBridge.ts
miguelmota: requiredLiquidity is simply the amount the bonder will need to bond (in this case it's the hToken amount output from swap)

packages/sdk/src/HopBridge.ts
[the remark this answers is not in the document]
shanefontaine: @miguelmota I think it also needs the parenthesis here, assuming BigNumber follows PEMDAS. So adding `()` around `availableLiquidity.sub(pendingAmounts)`

packages/sdk/src/HopBridge.ts
shanefontaine: Understood. I was reading this incorrectly. Makes sense & will resolve this now.

PR #153, opened 2021-09-18
hop-node: Set up optional channels to use for different log types

miguelmota:
Optional channel environment variables that can be used to direct different notification types to:

- `SLACK_WARN_CHANNEL`
- `SLACK_ERROR_CHANNEL`
- `SLACK_INFO_CHANNEL`
- `SLACK_LOG_CHANNEL`
- `SLACK_SUCCESS_CHANNEL`
- `GAS_BOOST_WARN_SLACK_CHANNEL`
- `GAS_BOOST_ERROR_SLACK_CHANNEL`

By default it'll use `SLACK_CHANNEL`

Example; if `SLACK_ERROR_CHANNEL` is set, then all `notifier.error()` messages go to the `SLACK_ERROR_CHANNEL` channel and `.info`, `.log`, etc go to `SLACK_CHANNEL`. If additionally `GAS_BOOST_ERROR_SLACK_CHANNEL` is set, then all gas boost class `notifier.error()` will go to `GAS_BOOST_ERROR_SLACK_CHANNEL` channel.


shanefontaine (commented):
Code LGTM! 

Only suggestion related to this is that I think we should log the `host` in the `queue.ts` and `rateLimitRetry.ts`. I do not believe we do those and therefore those are the only logs that are not explicitly associated with a host.

miguelmota:
sounds good, those can be added

PR #154, opened 2021-09-19
Feature/retry not enough credit quicker

shanefontaine:
This PR allows the bonder to re-bond transfers that are unbonded because there was not enough bonder liquidity at the time of the first attempted bond. At this time, we are seeing that a user in this case exponentially backs-off. In this case, the bonder's liquidity is replenished in 2 hours, for example, but the exponential backoff forces this transfer to take another few hours after the replenishment.

In order to optimize the time it takes to bond a transfer, we now store, in memory, the last available credit that the bonder had available for a given source/dest chainId. We now do not process transfers that are sending more value than the bonder has available.

### DIFF
--- a/packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
+++ b/packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
@@ -27,6 +27,7 @@ class BondError extends Error {}
 
 class BondWithdrawalWatcher extends BaseWatcher {
   siblingWatchers: { [chainId: string]: BondWithdrawalWatcher }
+  lastAvailableCredit: { [sourceChainId: string]: { [destinationChainId: string]: BigNumber }} = {}
 
   constructor (config: Config) {
     super({
@@ -61,7 +62,19 @@ class BondWithdrawalWatcher extends BaseWatcher {
     }
 
     const promises: Promise<any>[] = []
-    for (const { transferId } of dbTransfers) {
+    for (const dbTransfer of dbTransfers) {
+      const {
+        transferId,
+        sourceChainId,
+        destinationChainId,
+        amount
+      } = dbTransfer
+
+      const lastAvailableCredit = this.lastAvailableCredit[sourceChainId][destinationChainId]
+      if (lastAvailableCredit && lastAvailableCredit.lt(amount)) {
+        continue
+      }
+
       promises.push(this.checkTransferId(transferId).catch(err => {
         this.logger.error(`checkTransferId error: ${err.message}`)
       }))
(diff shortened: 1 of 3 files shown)


packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
```diff
@@ -27,6 +27,7 @@ class BondError extends Error {}
 
 class BondWithdrawalWatcher extends BaseWatcher {
   siblingWatchers: { [chainId: string]: BondWithdrawalWatcher }
+  lastAvailableCredit: { [sourceChain: string]: { [destinationChain: string]: BigNumber }} = {}
```
miguelmota, line 30: If the `lastAvailableCredit` `sourceChain` key is always going to be `this.bridge.chainSlug`, then it can be omitted to make the mapping simple

```js
lastAvailableCredit: { [destinationChain: string]: BigNumber } = {}
```

packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
```diff
@@ -61,7 +64,25 @@ class BondWithdrawalWatcher extends BaseWatcher {
...
+        destinationChainId,
+        amount,
+        withdrawalBondTxError
+      } = dbTransfer
+
+      const sourceChain = this.bridge.chainSlug
+      const destinationChain = this.chainIdToSlug(destinationChainId)
+      const lastAvailableCredit = this.lastAvailableCredit?.[sourceChain]?.[destinationChain]
+      if (
+        lastAvailableCredit &&
```
miguelmota, line 79: The first conditional can be omitted and the second can use `?` to simply things

```js
      if (
        lastAvailableCredit?.lt(amount) &&
        withdrawalBondTxError === TxError.NotEnoughLiquidity
      ) {
    
```

packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
```diff
@@ -129,21 +150,16 @@ class BondWithdrawalWatcher extends BaseWatcher {
       }
       availableCredit = availableCredit.sub(pendingAmounts).sub(amount)
     }
+
+    this.lastAvailableCredit[sourceChain][destinationChain] = availableCredit
```
miguelmota, line 154: I see `lastAvailableCredit` is only updated when a bond withdrawal is attempted; do you think it should also be updated when credit or debit is updated in general which have an effect on available liquidity (e.g. staking, settling, challenges)?

shanefontaine: That is a great thought & I didn't originally consider it.

My vote would be to not update in those cases & leave it as-is because I believe that it can introduce some edge cases while providing a negligible benefit. By negligible benefit, I mean that I see the benefit of adding the state updates in other watchers as being a user who is waiting will have their tx bonded faster (seconds or minutes). I believe, based on our current usage, this will be negligible because other user's send txs every few mins that will update this value anyway.

If we were to add the updates everywhere, one complex case I am imagining is "race conditions". Imagine we have two watchers who update it w/in milliseconds of each other. Imagine, in this case, that the one who updates it last actually did the `amount` calcs 1s before, which means it is overwriting w/ bad data. I think this can almost be guaranteed to not happen if we do the calc on the line before the state update, but then I _think_ we need to introduce the `includePendingAmount` in all watchers too.

Open for discussion if you think differently!

miguelmota: I don't think there would have to be updates everywhere; I think the available liquidity logic would be in a single place and the update gets triggered after every resync (it can be in the postSyncHandler). This makes it so that the available liquidity amount is always up to date after receiving new events

shanefontaine: I like that. I'll take a whack at it and send over a commit with this change.

PR #156, opened 2021-09-20
hop-node: Track gas prices and use nearest gas price for bond withdrawal bonder fee comparision

miguelmota:


packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
```diff
@@ -262,6 +262,16 @@ class BondWithdrawalWatcher extends BaseWatcher {
     logger.debug('recipient:', recipient)
     logger.debug('transferNonce:', transferNonce)
     logger.debug('bonderFee:', this.bridge.formatUnits(bonderFee))
+
+    let gasPrice : BigNumber
+    const dbTransfer = await this.db.transfers.getByTransferId(transferId)
+    if (dbTransfer?.transferSentTimestamp) {
+      const item = await this.db.gasPrices.getNearest(this.chainSlug, dbTransfer?.transferSentTimestamp)
```
shanefontaine, line 269: I believe the `chain` needs to be the `destinationChain` since, at the time of this call, we are in the context of the source chain watcher but need the `gasPrice` of the destination chain.

I do not know if `this.db.gasPrices` is aware of the other watcher's `gasPrices` db. If not, then I think `this.db.gasPrices` needs to be able to read from the other DBs.

packages/hop-node/src/db/GasPricesDb.ts
```diff
@@ -0,0 +1,36 @@
...
+  timestamp?: number
+  gasPrice?: BigNumber
+}
+
+class GasPricesDb extends BaseDb {
+  async update (key: string, data: Partial<GasPrice>) {
+    return super.update(key, data)
+  }
+
+  async getNearest (chain: string, timestampSeconds: number): Promise<GasPrice> {
```
shanefontaine, line 18: `chain` is unused. I would expect to filter by chain here.

packages/hop-node/src/db/GasPricesDb.ts
```diff
@@ -0,0 +1,36 @@
...
+    const keys = await this.getKeys()
+    const items = await Promise.all(
+      keys.map((key: string) => {
+        return this.getById(key)
+      })
+    )
+
+    const target = DateTime.fromSeconds(timestampSeconds)
+    const dates = items.map(item => DateTime.fromSeconds(item.timestamp).toJSDate())
```
shanefontaine, line 27: 1. As we have it, `timestampSeconds` is in seconds and `item.timestamp` is in ms. I think they should be the same unit since we process them with the same logic (`DateTime.fromSeconds(x).toJSDate()`).
2. I don't know the best way, but I think we should be using `new DateTime()` instead of `DateTime.fromSeconds()`. Not 100% sure though.


```
var luxon = require("luxon")

const timeSec = 1632174913
const timeMs = 1632174913000

console.log(new luxon.DateTime(timeSec))
console.log(new luxon.DateTime(timeMs))

console.log(luxon.DateTime.fromSeconds(timeSec))
console.log(luxon.DateTime.fromSeconds(timeMs))

console.log(luxon.DateTime.fromSeconds(timeSec).toJSDate())
console.log(luxon.DateTime.fromSeconds(timeMs).toJSDate())


DateTime {ts: 1632178046986, _zone: SystemZone {}, …}
DateTime {ts: 1632178047018, _zone: SystemZone {}, …}
DateTime {ts: 1632174913000, _zone: SystemZone {}, …}
DateTime {ts: 1632174913000000, _zone: SystemZone {}, …}
Mon Sep 20 2021 14:55:13 GMT-0700 (Pacific Daylight Time)
Wed Jul 25 53691 01:16:40 GMT-0700 (Pacific Daylight Time)
```

shanefontaine (changes requested):
@miguelmota back to you

packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
miguelmota: 1. you're correct, should be destinationChain
2. this.db.gasPrices is a global db and not watcher specific

packages/hop-node/src/db/GasPricesDb.ts
miguelmota: you're correct, it should filter here

packages/hop-node/src/db/GasPricesDb.ts
miguelmota: Good point, will make it be the time format

miguelmota:
Feedback issues resolved

shanefontaine (approved):
LGTM

Only other non-critical suggestion. It looks like `nearest-date` works with unix time, so I think we can remove `toJSDate()` to simplify code if you want. 

```
var nearestDate = require("nearest-date")

const dates = [
    1632174113,
    1632174513,
    1632174913
]

console.log(nearestDate(dates, 1632174813))

> 2
```

miguelmota:
True, updated now to use unix timestamp

PR #157, opened 2021-09-20
Bugfix for a "bug" that changes the WMATIC ticker to WWMATIC

ramuta:
If a user goes to the Pool section of the Hop Exchange page and opens the wrap/unwrap section (for MATIC on Polygon), it shows a WWMATIC ticker instead of WMATIC. This can confuse people.

It's a small fix in hop/packages/frontend/src/pages/Pools/TokenWrapper/TokenWrapper.tsx ([line 84](https://github.com/hop-protocol/hop/blob/3501350629bc09c831056d05e06dfcda3c9a0865/packages/frontend/src/pages/Pools/TokenWrapper/TokenWrapper.tsx#L84)):

```javascript
balanceLabel={`W${wrappedToken?.symbol}:`}
```

I only removed "W" from this line 😉


miguelmota:
Thanks for the fix @ramuta!

PR #159, opened 2021-09-22
Feature/available liquidity worker

miguelmota:
This PR is ongoing so it's in draft mode (still need the frontend portion) but wanted to get some feedback to make sure the calculations are correct. 

The availableCredit calcuation is [here](https://github.com/hop-protocol/hop/pull/159/files#diff-d7cca2a07e0034090a9c7c5ec85ba2e1a0dedd759b6535a4a8847ed937e0ac0fR712) and commitPendingAmounts is [here](https://github.com/hop-protocol/hop/pull/159/files#diff-d7cca2a07e0034090a9c7c5ec85ba2e1a0dedd759b6535a4a8847ed937e0ac0fR787).  commitPendingAmounts is basically the total amounts of committed transfers from the last hour that haven't yet been bonded.

This is what the structure of the json file it writes to S3 will look like once the worker is running (format is `{ token: { source: { destination: amount }}}`). The update will happen after every sync interval.

```bash
$ curl https://assets.hop.exchange/v1-available-liquidity.json
```

```json
{
  "timestamp": 1632272374050,
  "data": {
    "MATIC": {
      "availableCredit": {
        "xdai": {
          "ethereum": "0",
          "polygon": "0"
        },
        "polygon": {
          "ethereum": "0",
          "xdai": "0"
        }
      },
      "commitPendingAmounts": {
        "xdai": {
          "ethereum": "0",
          "polygon": "0"
        },
        "polygon": {
          "ethereum": "0",
          "xdai": "0"
        }
      }
    },
    "USDT": {
      "availableCredit": {
        "optimism": {
          "ethereum": "0",
          "arbitrum": "0",
          "xdai": "283039925",
          "polygon": "0"
        },
        "xdai": {
          "ethereum": "12103195010",
          "arbitrum": "0",
          "optimism": "8192000000",
          "polygon": "0"
        },
        "polygon": {
          "ethereum": "12103195010",
          "arbitrum": "0",
          "optimism": "8192000000",
          "xdai": "283039925"
        },
        "arbitrum": {
          "ethereum": "0",
          "optimism": "8192000000",
          "xdai": "283039925",
          "polygon": "0"
        }
      },
      "commitPendingAmounts": {
        "optimism": {
          "ethereum": "0",
          "arbitrum": "0",
          "xdai": "0",
          "polygon": "0"
        },
        "xdai": {
          "ethereum": "0",
          "arbitrum": "0",
          "optimism": "0",
          "polygon": "0"
        },
        "polygon": {
          "ethereum": "0",
          "arbitrum": "0",
          "optimism": "0",
          "xdai": "0"
        },
        "arbitrum": {
          "ethereum": "0",
          "optimism": "0",
          "xdai": "0",
          "polygon": "0"
        }
      }
    },
    "USDC": {
      "availableCredit": {
        "optimism": {
          "ethereum": "0",
          "arbitrum": "1260000000",
          "xdai": "4702000000",
          "polygon": "7347821164"
        },
        "xdai": {
          "ethereum": "72602271513",
          "arbitrum": "1260000000",
          "optimism": "5809000000",
          "polygon": "7347821164"
        },
        "polygon": {
          "ethereum": "72602271513",
          "arbitrum": "1260000000",
          "optimism": "5809000000",
          "xdai": "4702000000"
        },
        "arbitrum": {
          "ethereum": "0",
          "optimism": "5809000000",
          "xdai": "4702000000",
          "polygon": "7347821164"
        }
      },
      "commitPendingAmounts": {
        "optimism": {
          "ethereum": "0",
          "arbitrum": "0",
          "xdai": "0",
          "polygon": "0"
        },
        "xdai": {
          "ethereum": "0",
          "arbitrum": "0",
          "optimism": "0",
          "polygon": "0"
        },
        "polygon": {
          "ethereum": "0",
          "arbitrum": "0",
          "optimism": "0",
          "xdai": "0"
        },
        "arbitrum": {
          "ethereum": "0",
          "optimism": "0",
          "xdai": "0",
          "polygon": "0"
        }
      }
    }
  }
}
```

Let me know if this is what you had in mind in terms of how things should be calculated 

@shanefontaine 

I also need to update it to use the all bonder addresses in order to get available liquidity for assets other than what the currently running bonder has enabled (which is why matic show 0 available credit in example above)


packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -667,6 +708,148 @@ class SyncWatcher extends BaseWatcher {
...
+    }
+
+    if (availableCredit.lt(0)) {
+      return BigNumber.from(0)
+    }
+
+    return availableCredit
+  }
+
+  public async calculateRecentCommitPendingAmounts (destinationChainId: number) {
```
shanefontaine, line 787: Since we use the word "pending" to describe un-committed amounts, I would suggest we use a different name for all of these functions that have to do with the "in-flight" commits.

My vote would be "unsetCommits", so `calculateUnsetCommitAmounts`, for example.

shanefontaine (commented):
The calcs all look good. Left some comments on architecture. In general, this is pretty much what we've been needing 🙏 

Re: the equation in Slack, I think this code is good to go except for the `/2` at the end of the equation in Slack. I don't fully grasp that part and we can nail it down on a call, but I think this code will need that `/2` mentioned in the equations discussed earlier. I may be missing something here.

> I also need to update it to use the all bonder addresses in order to get available liquidity for assets other than what the currently running bonder has enabled (which is why matic show 0 available credit in example above)

Is the intention that we run a single worker that pushes the data to S3? Or would it be that all running bonders push data to S3?

If it is the latter, I would not expect what you said to be true but I see why you said it. Let's discuss this, but I wouldn't expect any bonder running this software to write any data that has to do with other bonders. I would expect, for example, the MATIC bonder to only write MATIC data somewhere while the USDC/USDT bonder only writes USDC/USDT data somewhere. If we want a single endpoint to call the data, I would expect we would aggregate all of the individual bonder's data upstream, not within their running process.

> `$ curl https://assets.hop.exchange/v1-available-liquidity.json`

There is a part of the response that doesn't make sense to me. I would expect that the `availableCredit` for all `USDT` chains with the `destinationChain` of `ethereum` to be the exact same. I see that `xdai` and `polygon` have `12103195010`, but `opt` and `arb` have `0`. Would you expect opt/arb to have the same `12103...` number?

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -667,6 +708,148 @@ class SyncWatcher extends BaseWatcher {
...
+      if (shouldSkip) {
+        continue
+      }
+      await this.updateAvailableCredit(destinationChainId)
+      const lastAvailableCredit = await this.getLastAvailableCredit(destinationChainId)
+      this.logger.debug(`lastAvailableCredit (${this.tokenSymbol} ${sourceChain}→${destinationChain}): ${this.bridge.formatUnits(lastAvailableCredit)}`)
+    }
+  }
+
+  public async getLastAvailableCredit (destinationChainId: number, amount: BigNumber = BigNumber.from(0)) {
```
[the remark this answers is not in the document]
miguelmota: So the relevant changes would be to add a `amountToCompare` method that takes in amount (like in your recent commit) and scope it to BondWithdrawalWatcher. I agree with all that

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -213,6 +213,32 @@ class TransferRootsDb extends BaseDb {
       )
     })
   }
+
+  async getRecentlyCommittedTransferRoots (filter: Partial<TransferRoot>): Promise<TransferRoot[]> {
```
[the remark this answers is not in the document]
miguelmota: that's a great suggestion, will update

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -213,6 +213,32 @@ class TransferRootsDb extends BaseDb {
...
+          return false
+        }
+      }
+      if (filter?.destinationChainId) {
+        if (filter.destinationChainId !== item.destinationChainId) {
+          return false
+        }
+      }
+
+      const timestampOk = item.committedAt && (item.committedAt >= Math.floor((Date.now() - OneHourMs) / 1000))
```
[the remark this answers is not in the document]
miguelmota: good to know

packages/hop-node/src/watchers/SyncWatcher.ts
miguelmota: I like the unsetCommits terminology, will update

miguelmota:
> Is the intention that we run a single worker that pushes the data to S3? Or would it be that all running bonders push data to S3?

It would only be a single instance that we run that pushes to s3. It's not expected that bonders would push to s3. This aggregate worker would be enabled by either a flag we run the bonder with or a separate command entirely (open to suggestions)

> I would expect that the availableCredit for all USDT chains with the destinationChain of ethereum to be the exact same.

I think you're right and need to verify the output

packages/hop-node/src/watchers/SyncWatcher.ts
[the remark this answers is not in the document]
shanefontaine: Yes exactly

miguelmota:
@shanefontaine

miguelmota:
PR updated based on latest discussion

- ORU -> L1: (`credit` - `debit` - `OruToL1PendingAmount` - `OruToAllUnbondedTransferRoots`) / `2`
  - divide by `2` because `amount` gets added to OruToL1PendingAmount
- nonORU -> L1: (`credit` - `debit` - `OruToL1PendingAmount` - `OruToAllUnbondedTransferRoots`)
- L2 -> L2: (`credit` - `debit`)

note

- OruToL1PendingAmount is [`getOruToL1PendingAmount()`](https://github.com/hop-protocol/hop/pull/159/files#diff-d7cca2a07e0034090a9c7c5ec85ba2e1a0dedd759b6535a4a8847ed937e0ac0fR865)
- OruToAllUnbondedTransferRoots is [`getOruToAllUnsetCommitAmount()`](https://github.com/hop-protocol/hop/pull/159/files#diff-d7cca2a07e0034090a9c7c5ec85ba2e1a0dedd759b6535a4a8847ed937e0ac0fR865)

@shanefontaine

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -662,6 +711,212 @@ class SyncWatcher extends BaseWatcher {
...
+  }
+
+  isNonOruToL1 (destinationChainId: number) {
+    const sourceChain = this.chainSlug
+    const destinationChain = this.chainIdToSlug(destinationChainId)
+    return destinationChain === Chain.Ethereum && !bondableChains.includes(sourceChain)
+  }
+
+  // ORU -> L1: (credit - debit - OruToL1PendingAmount - OruToAllUnbondedTransferRoots) / 2
+  //    - divide by 2 because `amount` gets added to OruToL1PendingAmount
```
shanefontaine, line 728: nitpick: we divide by 2 in the comments but we don't implement the /2 here, we do it upstream. I think it makes sense to keep the whole equation here in the comments, but it might be worth adding another comment saying that the /2 happens upstream.

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -662,6 +711,212 @@ class SyncWatcher extends BaseWatcher {
...
+
+  async syncUnsetCommitAmounts () {
+    const chains = await this.bridge.getChainIds()
+    for (const destinationChainId of chains) {
+      const sourceChain = this.chainSlug
+      const destinationChain = this.chainIdToSlug(destinationChainId)
+      const shouldSkip = (
+        sourceChain === Chain.Ethereum ||
+        sourceChain === destinationChain ||
+        !this.hasSiblingWatcher(destinationChainId)
```
shanefontaine, line 821: why does this function have the sibling watcher check but `syncPendingAmounts()` does not?

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -662,6 +711,212 @@ class SyncWatcher extends BaseWatcher {
...
+  }
+
+  async uploadToS3 () {
+    if (!this.s3Upload) {
+      return
+    }
+    if (!this.isAllSiblingWatchersInitialSyncCompleted()) {
+      return
+    }
+    const { syncIndex } = this.getSiblingWatcherByChainSlug(Chain.Ethereum)
```
shanefontaine, line 892: See comment above about the `syncIndex`

shanefontaine (commented):
* I describe it in an issue below, but I was wrong about using the term `unset` for the transfer root. We should be using `unbonded` everywhere instead of `unset`.
* I think the "stale" data check included in the SDK is really good. It looks like we don't have a similar check in the bonder. Do you think we should add an `if (stale)` check to the bonder so that the bonder does not get blocked by a weird bug?
  * A non-happy path example would be that a large transfer is sent, `calculateUnsetCommitAmounts()` is very high, then the available credit worker dies for some reason, then the bonder gets stuck thinking it has 0 liquidity available.
* I think it makes sense to come up with a different naming terminology for `availableCredit` to avoid confusion as to what the return data is supposed to represent. As an example, `getAvailableCredit()` seems to mean different things in different places:
  * used as a function for on-chain available credit
    * represents `credit-debit`
  * used as a function in BondWithdrawalWatcher
    * represents `credit - debit - ORUcalcs`
  * used as a function in SyncWatcher
    * represents `credit - debit - ORUcalcs`

I don't have a good suggestion here. Only thing I can come up with is to change the on-chain function name to `getOnChainAvaiableCredit`. Any thoughts here?

packages/hop-node/src/watchers/BondTransferRootWatcher.ts
```diff
@@ -108,6 +108,9 @@ class BondTransferRootWatcher extends BaseWatcher {
         `transferRootHash ${transferRootHash} already bonded. skipping.`
       )
       const event = await l1Bridge.getTransferRootBondedEvent(transferRootHash)
+      if (!event) {
```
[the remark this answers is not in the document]
miguelmota: I was seeing this particular event be null randomly when syncing so destructuring would fail so wanted to make the error more friendly. The other places already have null event handling

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -213,6 +213,31 @@ class TransferRootsDb extends BaseDb {
       )
     })
   }
+
+  async getUnsetCommittedTransferRoots (filter: Partial<TransferRoot>): Promise<TransferRoot[]> {
```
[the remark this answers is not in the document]
miguelmota: That makes sense, will updated to use `getUnbondedTransferRoots()`.

The reason that currently getUnbondedTransferRoots doesn't take in source/destination is because there's only one instance of BondTransferRootWatcher (only L1) so it's not concerned about filtering

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -207,6 +244,12 @@ class SyncWatcher extends BaseWatcher {
     )
 
     await Promise.all(promises)
+
+    // these must come after db is done syncing,
+    // and syncAvailableCredit must be last
```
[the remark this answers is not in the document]
miguelmota: great point, will use .then()

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -268,11 +311,17 @@ class SyncWatcher extends BaseWatcher {
   }
 
   async handleWithdrawalBondedEvent (event: Event) {
+    if (!event) {
```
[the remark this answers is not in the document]
miguelmota: I think this check is an artifact from my debugging, will remove

packages/hop-node/src/watchers/SyncWatcher.ts
miguelmota: good point, will do

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -104,6 +133,14 @@ class SyncWatcher extends BaseWatcher {
     )
   }
 
+  isAllSiblingWatchersSyncCompleted (syncIndex: number = this.syncIndex): boolean {
+    return Object.values(this.siblingWatchers).every(
+      (siblingWatcher: SyncWatcher) => {
+        return siblingWatcher.syncIndex === syncIndex
```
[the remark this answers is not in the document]
miguelmota: I believe you're right. The problem is that isAllSiblingWatchersInitialSyncCompleted is only set to true once after initial sync but I need a way to know when every re-sync finishes to trigger the re-calculations and upload to s3. I'll think about how to do it better since it won't work based on what you said which I think you're correct

packages/hop-node/src/watchers/SyncWatcher.ts
miguelmota: This method has a sibling watcher call but not syncPendingAmounts

miguelmota:
1. good idea, it's been updated now to use 'unbonded' terms
2. makes sense, a stale check has been added now
3. this is what I came up with (we can always change this later)
    - `getBaseAvailableCredit` - on-chain available credit
    - `getEffectiveAvailableCredit` - SyncWatcher method that takes account pending amounts
    - `getAvailableCreditForTransfer` - BondWithdrawalWatcher method that takes account transfer amount

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -674,6 +712,213 @@ class SyncWatcher extends BaseWatcher {
...
+    }
+
+    return pendingAmounts
+  }
+
+  async getOruToAllUnbondedTransferRootAmounts () {
+    let totalAmount = BigNumber.from(0)
+    for (const destinationChain in this.unbondedTransferRootAmounts) {
+      if (this.lastCalculated[this.lastCalculated[destinationChain]]) {
+        const tenMinutes = 10 * 60 * 1000
```
shanefontaine, line 872: We have a `TenMinutesMs` in constants.ts. It probably makes sense to use that here.

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -674,6 +712,213 @@ class SyncWatcher extends BaseWatcher {
...
+      pendingAmounts = pendingAmounts.add(pendingAmount)
+    }
+
+    return pendingAmounts
+  }
+
+  async getOruToAllUnbondedTransferRootAmounts () {
+    let totalAmount = BigNumber.from(0)
+    for (const destinationChain in this.unbondedTransferRootAmounts) {
+      if (this.lastCalculated[this.lastCalculated[destinationChain]]) {
```
shanefontaine, line 871: I believe this is supposed to not be nested. I think it should just be `this.lastCalculated[destinationChain]`?

packages/sdk/src/HopBridge.ts
```diff
@@ -787,11 +788,15 @@ class HopBridge extends Base {
 
     let availableLiquidity = credit.sub(debit)
 
-    const isBondableChain = bondableChains.includes(sourceChain.slug)
-    const includePendingAmount =
-      isBondableChain && destinationChain.equals(Chain.Ethereum)
+    const unsetCommitAmount = await this.getUnsetCommitAmount(
```
shanefontaine, line 791: unset -> unbonded

packages/sdk/src/HopBridge.ts
```diff
@@ -817,10 +821,14 @@ class HopBridge extends Base {
         .div(tokenPriceBn)
         .mul(precision)
 
-      const liquidityMinusAmounts = availableLiquidity.sub(pendingAmounts)
-      availableLiquidity = liquidityMinusAmounts
-        .div(2)
-        .sub(bufferAmountTokensBn) // account for transfer root bonds
+      availableLiquidity = availableLiquidity
+        .sub(pendingAmounts)
+        .sub(unsetCommitAmount)
```
shanefontaine, line 826: unset -> unbonded

packages/sdk/src/HopBridge.ts
```diff
@@ -830,6 +838,56 @@ class HopBridge extends Base {
...
+    const tenMinutes = 10 * 60 * 1000
+    const isOutdated = Date.now() - timestamp > tenMinutes
+    if (isOutdated) {
+      return
+    }
+
+    return data
+  }
+
+  async getUnsetCommitAmount (sourceChain: Chain, destinationChain: Chain) {
```
shanefontaine, line 872: unset -> unbonded

packages/sdk/src/HopBridge.ts
```diff
@@ -830,6 +838,56 @@ class HopBridge extends Base {
...
+    try {
+      const data = await this.getBonderAvailableLiquidityData()
+      if (data) {
+        const _unsetCommitAmount =
```
shanefontaine, line 876: unset -> unbonded

packages/sdk/src/HopBridge.ts
```diff
@@ -830,6 +838,56 @@ class HopBridge extends Base {
...
+          data?.[this.tokenSymbol]?.unsetCommitAmounts?.[sourceChain.slug]?.[
```
shanefontaine, line 877: unset -> unbonded

packages/sdk/src/HopBridge.ts
```diff
@@ -830,6 +838,56 @@ class HopBridge extends Base {
...
+            destinationChain.slug
+          ]
+        if (_unsetCommitAmount) {
```
shanefontaine, line 880: unset -> unbonded

packages/sdk/src/HopBridge.ts
```diff
@@ -830,6 +838,56 @@ class HopBridge extends Base {
...
+          return BigNumber.from(_unsetCommitAmount)
```
shanefontaine, line 881: unset -> unbonded

packages/hop-node/src/config/config.ts
```diff
@@ -33,13 +33,17 @@ export const minPriorityFeePerGas = normalizeEnvVarNumber(process.env.MIN_PRIORI
...
+export const awsProfile = process.env.AWS_PROFILE
 const envNetwork = process.env.NETWORK || Network.Kovan
 const isTestMode = !!process.env.TEST_MODE
 const bonderPrivateKey = process.env.BONDER_PRIVATE_KEY
 
-export const bondableChains: string[] = [Chain.Optimism, Chain.Arbitrum]
+export const oruChains: string[] = [Chain.Optimism, Chain.Arbitrum]
 export const rateLimitMaxRetries = 5
-export const rpcTimeoutSeconds = 10 * 60
+export const rpcTimeoutSeconds = 90
```
shanefontaine, line 46: It looks like this is the only place we use this variable is here. I think we should remove the `* 1000` in this line.

https://github.com/hop-protocol/hop/blob/develop/packages/hop-node/src/decorators/rateLimitRetry.ts#L28

shanefontaine (commented):
@miguelmota back to you

LGTM. Left some final comments but I'd feel good merging after you make these final changes.  A few comments:

1. I think the new naming for the available credit is perfect
2. For my understanding, every running bonder will run the code to update the data, but only a bonder with s3 credentials will do anything with that data, correct? 
3. I _think_ there are a few more places to change `unset -> unbonded`. I left comments at each.

packages/hop-node/src/config/config.ts
shanefontaine: I think I am incorrect. I am going to resolve this.

packages/sdk/src/HopBridge.ts
miguelmota: forgot to update sdk, thanks for catching that

packages/hop-node/src/watchers/SyncWatcher.ts
miguelmota: nice catch, copy paste error. Fixed

packages/hop-node/src/watchers/SyncWatcher.ts
miguelmota: good call, fixed

miguelmota:
2. that's right, the data is only uploaded to s3 if the bonder is ran with the `--s3-upload` flag and has AWS credentials set

Issue #162, opened 2021-09-22
Meta: update/upgrade package management


kangarang:
currently we git-ignore package lock files and rely on each individual machine to correctly install dependencies for each individual (lerna) package. we should use lock files to keep our `node_modules` synced across local and remote environments. additionally, i propose we adopt yarn package manager throughout the monorepo -- installations are quicker, error messages are clearer, and the cli is cleaner (imo).

steps to update/upgrade package management:
- [x] rm `package-lock.json` and `yarn.lock` from `.gitignore` (and local workspace if present)
- [x] simplify nested `.gitgnore` files (rm extraneous lines)
- [x] update/set node/yarn versions: add `engines.node: "^12.22.1"`/`engines.yarn: "^1.22.11"` to `package.json`
- [ ] (optional) allow minor package diffs using `^` (e.g. `"react": "^17.0.2",`)
- [x] upgrade dependencies (run `yarn install`, which creates `yarn.lock`)

for completeness, replace instances where we use the command `npm` -> `yarn` in the following:
- [x] npm scripts (`package.json`)
- [x] readme instructions
- [x] docs

miguelmota:
This all sounds great @kangarang!

Does it make sense to not check in package-lock.json if we use yarn and yarn.lock? Thinking of ways that we can minimize confusion on which package installer to use

kangarang:
@miguelmota yep, we should delete all instances of `package-lock.json` and avoid using it from now on. removing it from `.gitignore` will ensure that none of our environments have it. if we forget to use yarn and it gets created by npm, it'll be obvious during code review

PR #164, opened 2021-09-24
Meta: update/upgrade package management

kangarang:
addresses #162 

note: this pr includes changes to all 5 lerna packages. i'm curious to see if these changes break `hop-node` and `stats-worker`, as i haven't run them yet (the other 3 pkgs build/run successfully)


miguelmota:
Nice job on the first PR @kangarang!

My feedback below

**.nvmrc**

We have a `.nvmrc` file that makes it so we can automatically change to the right node version when doing `nvm use`

DeepinScreenshot_select-area_20210923182645

Since it is removed now, how do you go about quickly changing to the correct node version?

DeepinScreenshot_select-area_20210923182651


**.node-version**

Some CI services, like netlify, use `.node-version` or `.nvmrc` to determine what node version to use ([docs](https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-environment)), otherwise it uses node `10` by default. I think we'll need to keep one of these files for CI services

**`Dockerfile`**

The `Dockerfile` in `hop-node` and `stats-workers` needs to have the node version updated as well ([line](https://github.com/hop-protocol/hop/blob/develop/packages/hop-node/Dockerfile#L1)). It can be tested by running `make docker-build`

kangarang:
@miguelmota thank you for the insight! will re-add these files and set node version to 14

miguelmota (approved):

Issue #150, opened 2021-09-17
UI: Better Error message when user does not have enough ETH in their wallet


cwhinfrey:
Steps to reproduce:

1. Send funds to a wallet on Arbitrum that doesn't have ETH
2. Try to send from Arbitrum with that wallet

Current behavior:

A nondescript error from MetaMask is shown

Expected behavior:
An error telling the user to add ETH to their wallet for gas should be shown.

miguelmota:
Adding more info to this issue;
some users have mentioned they see a `Internal JSON-RPC error` when they try adding liquidity on Arbitrum with very little or no ETH. We should check if ETH balance is greater than 0 before attempting any transactions.

PR #165, opened 2021-09-24
UI: display insufficient funds error msg (approve tx)

kangarang:
addresses #150

Screen Shot 2021-09-23 at 8 57 14 PM

additionally, this pr includes a more detailed warning when the bonder fee > estimated received

Screen Shot 2021-09-23 at 8 40 03 PM


miguelmota:
Looks good 👍 @kangarang 

There's a similar util `prettifyErrorMessage` ([file](https://github.com/hop-protocol/hop/blob/develop/packages/frontend/src/utils/prettifyErrorMessage.ts)), maybe they can be combined?

packages/frontend/src/utils/format.ts
```diff
@@ -0,0 +1,28 @@
...
+export function formatError(error: Error & ErrorData) {
+  if (typeof error === 'string') {
+    return error
+  }
+
+  const { data } = error
+
+  // TODO: handle custom error messages elsewhere (and better)
+  if (data?.message === 'not enough funds for gas') {
+    return 'Insufficient funds. Please add ETH to pay for tx gas.'
```
shanefontaine, line 16: The native (fee) token on Polygon is MATIC and the native token on xDai id xDai, so technically "ETH" will not always be true here.

PR #169, opened 2021-09-25
sdk: Update optimism gas estimate call

miguelmota:


shanefontaine:
@miguelmota LGTM 👍 

I think we can move the whole estimation outside of the `if` statement and remove the concept of `BondTransferGasLimit` so that we use it for every chain, if we can trust that this estimation works constantly. 

I'm down to merge this and revisit that later if it will require additional testing and verification.

miguelmota:
@shanefontaine that's a good idea. Going to merge it in and then we can do the dynamic estimation for all chains in another commit since that requires more testing

shanefontaine:
Actually I take back the comment about removing `BondTransferGasLimit `. I think this might behave oddly at the edge cases. For example, if the estimation fails, I'm not sure if the function throws (and we `catch` it) or if it will return a very large `gasLimit` (10m+).

If someone uses a huge `amount`, for example, the est

PR #170, opened 2021-09-25
Feature/bug bash

shanefontaine:
Some house keeping commits:

* Change `CommitTransferWatcher` to `CommitTransfersWatcher` where appropriate.
* Only log if the bonder fee is too low. Do not throw an error.
* Wait 7 days before attempting an ORU exit.
  * I believe this will fix some (most) of our exit watcher errors, as we were trying to process exits ~10 minutes after a commit. I think.


miguelmota:
Everything looks good to me

Ready for merge

@shanefontaine

Issue #167, opened 2021-09-24
UI: Use query params across routes to set default networks


kangarang:
`sourceNetwork`

`destNetwork`

`token`

PR #176, opened 2021-09-28
UI: use query params to set initial network/token state

kangarang:
addresses #167 

this pr enables the gui to persist a user selected network and token when changing routes

### DIFF
--- a/packages/frontend/src/contexts/AppContext/AppContext.tsx
+++ b/packages/frontend/src/contexts/AppContext/AppContext.tsx
@@ -23,7 +23,7 @@ type AppContextProps = {
   user: User | undefined
   networks: Network[]
   l2Networks: Network[]
-  DEFAULT_L2_NETWORK: Network | undefined
+  defaultL2Network: Network | undefined
   l1Network: Network | undefined
   contracts: Contracts | undefined
   tokens: Token[]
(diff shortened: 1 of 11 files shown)


kangarang:
set base to prettier for sanity for now

miguelmota:
Works pretty good! @kangarang 

There's on small issue where the source and destination on the send page end up being the same 

**Steps to reproduce**

Step 1: Select xDai as source and Polygon as destination

DeepinScreenshot_select-area_20210929131251

Step 2: Go to convert page and select Polygon 

DeepinScreenshot_select-area_20210929131413

Step 3: Go back to send page and notice Polygon is set to as both source and destination

DeepinScreenshot_select-area_20210929131431

**Expected behavior:**

Destination chain to be another selection (not the same as source) or unselected

packages/frontend/src/contexts/AppContext/AppContext.tsx
```diff
@@ -22,6 +22,8 @@ type AppContextProps = {
   setSelectedBridge: (bridge: HopBridge) => void
   user: User | undefined
   networks: Network[]
+  l2Networks: Network[]
+  DEFAULT_L2_NETWORK: Network | undefined
```
miguelmota, line 26: I think this should be treated as a regular read only property (`defaultL2Network`) instead of an uppercased constant since this value is computed and is not a hard coded constant

miguelmota (approved):

PR #177, opened 2021-09-28
Format frontend dir with prettier

kangarang:
this pr includes changes resulting from running `prettier` with the new settings. also included is a minor routing change which is further developed in #167 (can be ignored here for now)


miguelmota:
Looks good @kangarang. Can be merged

PR #178, opened 2021-09-29
hop-node: Fix unhandled rejection error and nonce too low

miguelmota:


packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
```diff
@@ -213,9 +213,6 @@ class BondWithdrawalWatcher extends BaseWatcher {
             this.bridge.formatUnits(bondedAmount)
           )
         })
-        .catch(async (err: Error) => {
```
shanefontaine, line 216: Out of curiosity, why was this removed -- is it redundant? 

My understanding is that our new change will not throw anymore, but what do you expect to happen if another part of the code throws and this does not exist?

packages/hop-node/src/gasboost/GasBoostTransaction.ts
```diff
@@ -566,7 +566,18 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
       } catch (err) {
         const isAlreadyKnown = /AlreadyKnown/gi.test(err.message)
         const isFeeTooLow = /FeeTooLowToCompete/gi.test(err.message)
+        const nonceTooLow = /(nonce.*too low|already been used|NONCE_EXPIRED)/gi.test(err.message)
         const shouldRetry = (isAlreadyKnown || isFeeTooLow) && i < maxRetries
```
shanefontaine, line 570: If a transaction consistently uses a nonce that is too low (for whatever reason) it will try every 10s and ignore `maxRetries`. I would imagine that it might be ideal to check `maxRetries` for the nonce handler as well. I might be not fully thinking through something.

packages/hop-node/src/gasboost/GasBoostTransaction.ts
```diff
@@ -566,7 +566,18 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
...
+        if (nonceTooLow) {
+          if (!this.nonce) {
```
shanefontaine, line 572: My understanding is that there is no `this.nonce` for the first tx, but if it is boosted then it will exist because it will be passed into the constrcutor.

If that is correct, I think this stops us handling the issue for non-boosted transactions. I would imagine we want to handle those. Do I have a bad assumption somewhere?

packages/hop-node/src/gasboost/GasBoostTransaction.ts
```diff
@@ -566,7 +566,18 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
...
+            // wait a bit before attempting again so it re-fetches latest nonce
+            const warnMsg = `nonce too low. Waiting a moment before retrying again. Error: ${err.message}`
+            this.logger.warn(warnMsg)
+            this.notifier.warn(warnMsg, { channel: gasBoostWarnSlackChannel })
+            await wait(10 * 1000)
+            continue
```
shanefontaine, line 578: I think we discussed only trying 5 times and then erroring or throwing away the transfer. If I am reading this correctly, I think it would constantly try a single tx every 10s forever until it works. This would block all other transactions, I believe.

I think we need to throw it away or do something different so we don't get stuck in an infinite loop. I'd expect an RPC endpoint to fix itself eventually, but if we have a bug in our code that would cause this to fail every time, I'd imagine we need to not loop infinitely.

shanefontaine (commented):
@miguelmota back to you. Higher level comments:

1. I see that the `nonce` in `send()` function in the GasBoostTransaction.ts is unused. Is this just an artifact that should be deleted or was it meant to be used somewhere?
2. Please don't merge after fixes. I want to take a fresh look at this tomorrow before merging 👍

packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
miguelmota: yes it's redundant. Removing catch that throws same error produces same result

```
async function p() {
  throw new Error('not ok error')
}

async function main() {
  try {
    await p().then(() => console.log('ok'))
    // removing this has same effect
    .catch(err => {
      throw err
    })
  } catch (err) {
    console.log('got error:', err)
  }
}

main()
```

miguelmota:
1. `GasBoostSigner.sendTransaction()` calls `GasBoostTransaction.send()` ([line](https://github.com/hop-protocol/hop/blob/bd29d2efaa91a585b18445022f43b2986636588b/packages/hop-node/src/gasboost/GasBoostSigner.ts#L50))
2. 👍

packages/hop-node/src/gasboost/GasBoostTransaction.ts
miguelmota: good catch, fixed with https://github.com/hop-protocol/hop/pull/178/commits/c3439f45339a4c385de5cfffbaf617c8f9873441

packages/hop-node/src/gasboost/GasBoostTransaction.ts
miguelmota: I believe it is fixed with https://github.com/hop-protocol/hop/pull/178/commits/c3439f45339a4c385de5cfffbaf617c8f9873441

packages/hop-node/src/gasboost/GasBoostTransaction.ts
miguelmota: Correct, if it is boosted then the nonce would be set and not reach this part of code. The thinking was that nonce should not be overridden if it has been set because it's possible that another transaction could already have the same nonce and we don't want to override another transaction or have the current transaction be replaced by the other transaction. Do you think the nonce check if statement here should be removed?

miguelmota:
This simple example shows that infura sometimes returns the same pending nonce even after waiting 5s after broadcasting transaction.

```js
require('dotenv').config()
const wait = require('wait')
const ethers = require('ethers')
const privateKey = process.env.PRIVATE_KEY
async function main() {
  const url = 'https://kovan.rpc.hop.exchange'
  const provider = new ethers.providers.StaticJsonRpcProvider(url)
  const signer = new ethers.Wallet(privateKey, provider)
  for (let i = 0; i < 10; i++) {
      const nonce = await signer.getTransactionCount('pending')
      console.log(i, nonce)
      const tx = await signer.sendTransaction({
        to: await signer.getAddress(),
        value: ethers.utils.hexlify(i),
      })
      console.log(i, tx.nonce, tx.hash)
      await wait(5 * 1000)
  }
}
main()
```

DeepinScreenshot_select-area_20210928233400

DeepinScreenshot_select-area_20210928233421

@shanefontaine

shanefontaine:
> GasBoostSigner.sendTransaction() calls GasBoostTransaction.send() (line)

I should have linked the line. It looks like [this](https://github.com/hop-protocol/hop/blob/64b5ca64f95bd52c79007db987154442e823b4c1/packages/hop-node/src/gasboost/GasBoostTransaction.ts#L254) `nonce` is unused. Is this expected or is it an artifact?

> This simple example shows that infura sometimes returns the same pending nonce even after waiting 5s after broadcasting transaction.

Really great find 🙏  Do you think it is worth adding a `wait(5*1000)` to the first line of `GasBoostTransaction.send()` with a comment mentioning that we need to wait for the RPC endpoint to be in sync?

The reason for 5s (instead of 10s) is that it looks like this is not happening to us too often, so it probably takes a few seconds for our program to process a tx, which is usually enough time for Infura to catch up & use the correct nonce. 5s would be a buffer on top of that processing time, which I would imagine would help reduce the number of nonce errors to basically 0. I think the downside is that we have to wait 5s extra per tx but the upside is (1) no (theoretical) errors and (2) a tx that errors out would have to wait 10s more after erroring out.

I'm pretty up in the air here because I hate arbitrary things in code -- I could go either way. What do you think?

shanefontaine:
@miguelmota back to you. those are my last two comments. The rest of the code other than those comments LGTM.

miguelmota:
1. yes it was an artifact. Removed now
2. instead of forcing a delay, I added a delay that only waits a maximum delta between transaction sends. I used 7 seconds since that seemed to be enough time to use a unique nonce every time. For example, if tx1 is sent at T-0 and tx2 is sent at T+20s, then there's no need to wait a delay. If tx1 is sent at T-0 and tx2 is sent at T+2s, then it waits 5s so there's exactly 7s between tx sends. https://github.com/hop-protocol/hop/pull/178/commits/350154f606834fe486049c0ec3265ed462f74ae2

shanefontaine:
@miguelmota LGTM. Ready to be merged IMO

PR #186, opened 2021-10-02
hop-node: change fee check

shanefontaine:


packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
```diff
@@ -259,15 +259,11 @@ class BondWithdrawalWatcher extends BaseWatcher {
...
-    const {
-      gasPrice,
-      tokenUsdPrice,
-      chainNativeTokenUsdPrice
-    } = await this.getPricesNearTransferEvent(dbTransfer)
-    logger.debug('gasPrice:', gasPrice.toString())
-    logger.debug('tokenUsdPrice:', tokenUsdPrice)
-    logger.debug('chainNativeTokenUsdPrice:', chainNativeTokenUsdPrice)
+    const minFee = BigNumber.from('1000000')
+    if (bonderFee.lt(minFee)) {
```
miguelmota, line 263: I think this bonder fee check can be moved to the db getter so it doesn't try to process it at all

shanefontaine:
Closing this since it was just an example.

PR #187, opened 2021-10-03
hop-node: Db: Add key stream filter to filter within range

miguelmota:
Adds leveldb createKeyStream filter [options]( https://www.npmjs.com/package/levelup#dbcreatereadstreamoptions) to narrow lookup. Drastically improves lookup speed by having leveldb narrow the scope instead of filtering a constantly increasing list of entries.

Filter options added to `GasPricesDb` + `TokenPricesDb` `getNearest` method and `TransfersDb` db getters.

gas prices lookup example (finding an entry closest to timestamp):

without key filter

```
items: 19947
elapsed: 2430.229ms
```

DeepinScreenshot_select-area_20211002220816

with key filter

```
items: 237
elapsed: 24.927ms
```

DeepinScreenshot_select-area_20211002220744

both examples return expected entry.

Transfers db lookup example (finding entries between two dates):

without key filter

```
items: 7
elapsed: 546.952ms
```

DeepinScreenshot_select-area_20211004010248

with key filter

```
items: 7
elapsed: 14.137ms
```

DeepinScreenshot_select-area_20211004010155

both example return the same expected set of entries.

This [medium article](https://kevinsimper.medium.com/how-to-get-range-of-keys-in-leveldb-and-how-gt-and-lt-works-29a8f1e11782) explains filters in more detail (and the reason for using tilda).

@shanefontaine

### DIFF
--- a/packages/hop-node/src/db/GasPricesDb.ts
+++ b/packages/hop-node/src/db/GasPricesDb.ts
@@ -1,8 +1,8 @@
-import BaseDb, { BaseItem } from './BaseDb'
+import BaseDb, { BaseItem, KeyFilter } from './BaseDb'
 import nearest from 'nearest-date'
 import wait from 'src/utils/wait'
 import { BigNumber } from 'ethers'
-import { OneHourMs, OneWeekMs } from 'src/constants'
+import { OneHourMs, OneHourSeconds, OneWeekMs } from 'src/constants'
 import { normalizeDbItem } from './utils'
 
 export const varianceSeconds = 10 * 60
@@ -40,8 +40,8 @@ class GasPricesDb extends BaseDb {
     return this.update(key, data)
   }
 
-  async getItems ():Promise<GasPrice[]> {
-    const keys = await this.getKeys()
+  async getItems (filter?: KeyFilter):Promise<GasPrice[]> {
+    const keys = await this.getKeys(filter)
     const items: GasPrice[] = (await Promise.all(
       keys.map((key: string) => {
         return this.getById(key)
@@ -51,7 +51,13 @@ class GasPricesDb extends BaseDb {
   }
 
   async getNearest (chain: string, targetTimestamp: number, staleCheck: boolean = true): Promise<GasPrice | null> {
-    const items : GasPrice[] = (await this.getItems()).filter((item: GasPrice) => item.chain === chain && item.timestamp)
+    const startTimestamp = targetTimestamp - OneHourSeconds
+    const endTimestamp = targetTimestamp + OneHourSeconds
+    const filter = {
+      gte: `${chain}:${startTimestamp}`,
+      lte: `${chain}:${endTimestamp}`
+    }
+    const items : GasPrice[] = (await this.getItems(filter)).filter((item: GasPrice) => item.chain === chain && item.timestamp)
 
     const dates = items.map((item: GasPrice) => item.timestamp)
     const index = nearest(dates, targetTimestamp)
--- a/packages/hop-node/src/db/GasPricesDb.ts
+++ b/packages/hop-node/src/db/GasPricesDb.ts
@@ -55,7 +55,7 @@ class GasPricesDb extends BaseDb {
     const endTimestamp = targetTimestamp + OneHourSeconds
     const filter = {
       gte: `${chain}:${startTimestamp}`,
-      lte: `${chain}:${endTimestamp}`
+      lte: `${chain}:${endTimestamp}~`
     }
     const items : GasPrice[] = (await this.getItems(filter)).filter((item: GasPrice) => item.chain === chain && item.timestamp)
 
--- a/packages/hop-node/src/db/TokenPricesDb.ts
+++ b/packages/hop-node/src/db/TokenPricesDb.ts
@@ -53,7 +53,7 @@ class TokenPricesDb extends BaseDb {
     const endTimestamp = targetTimestamp + OneHourSeconds
     const filter = {
       gte: `${token}:${startTimestamp}`,
-      lte: `${token}:${endTimestamp}`
+      lte: `${token}:${endTimestamp}~`
     }
     const items : TokenPrice[] = (await this.getItems(filter)).filter((item: TokenPrice) => item.token === token && item.timestamp)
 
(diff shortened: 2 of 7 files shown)


packages/hop-node/src/db/GasPricesDb.ts
```diff
@@ -51,7 +51,13 @@ class GasPricesDb extends BaseDb {
...
 
   async getNearest (chain: string, targetTimestamp: number, staleCheck: boolean = true): Promise<GasPrice | null> {
-    const items : GasPrice[] = (await this.getItems()).filter((item: GasPrice) => item.chain === chain && item.timestamp)
+    const startTimestamp = targetTimestamp - OneHourSeconds
+    const endTimestamp = targetTimestamp + OneHourSeconds
+    const filter = {
+      gte: `${chain}:${startTimestamp}`,
+      lte: `${chain}:${endTimestamp}~`
+    }
+    const items : GasPrice[] = (await this.getItems(filter)).filter((item: GasPrice) => item.chain === chain && item.timestamp)
```
shanefontaine, line 60: Does it make sense to explicitly handle the case where `items` is `null` after this line? This might happen, for example, when, for some reason, there were no `gasPrice` db writes within +- 1 hour of this call. Another time may be when calling from the CLI.

packages/hop-node/src/db/TokenPricesDb.ts
```diff
@@ -49,7 +49,13 @@ class TokenPricesDb extends BaseDb {
...
 
   async getNearest (token: string, targetTimestamp: number, staleCheck: boolean = true): Promise<TokenPrice | null> {
-    const items : TokenPrice[] = (await this.getItems()).filter((item: TokenPrice) => item.token === token && item.timestamp)
+    const startTimestamp = targetTimestamp - OneHourSeconds
+    const endTimestamp = targetTimestamp + OneHourSeconds
+    const filter = {
+      gte: `${token}:${startTimestamp}`,
+      lte: `${token}:${endTimestamp}~`
+    }
+    const items : TokenPrice[] = (await this.getItems(filter)).filter((item: TokenPrice) => item.token === token && item.timestamp)
```
shanefontaine, line 58: Same comment about explicit check as above.

packages/hop-node/src/db/GasPricesDb.ts
```diff
@@ -51,7 +51,13 @@ class GasPricesDb extends BaseDb {
   }
 
   async getNearest (chain: string, targetTimestamp: number, staleCheck: boolean = true): Promise<GasPrice | null> {
-    const items : GasPrice[] = (await this.getItems()).filter((item: GasPrice) => item.chain === chain && item.timestamp)
+    const startTimestamp = targetTimestamp - OneHourSeconds
+    const endTimestamp = targetTimestamp + OneHourSeconds
+    const filter = {
+      gte: `${chain}:${startTimestamp}`,
+      lte: `${chain}:${endTimestamp}~`
```
shanefontaine, line 58: Is the tilde here intentional?

packages/hop-node/src/db/TransfersDb.ts
```diff
@@ -36,8 +42,61 @@ export type Transfer = {
...
+      return
+    }
+    const exists = await this.getById(key)
+    if (!exists) {
+      await super.update(key, { transferId })
+    }
+  }
+
+  getTimestampedKey (transfer: Partial<Transfer>) {
+    if (transfer?.transferSentTimestamp && transfer?.transferSentIndex !== undefined) {
```
shanefontaine, line 91: With our latest `pure` event update PR, I believe this data will not exist immediately. In that case, is the timestamp ever added to the entry?

packages/hop-node/src/db/TransfersDb.ts
```diff
@@ -36,8 +42,61 @@ export type Transfer = {
...
+
+  async trackTimestampedKey (transfer: Partial<Transfer>) {
+    const transferId = transfer.transferId
+    const key = this.getTimestampedKey(transfer)
+    if (!key || !transferId) {
+      return
+    }
+    const exists = await this.getById(key)
+    if (!exists) {
+      await super.update(key, { transferId })
```
shanefontaine, line 86: Will we have two entries per transfer (for all old txs) -- one with a timestamp and one without?

shanefontaine (commented):
@miguelmota back to you

1. It looks like this PR only updated the `transfers` db to get items within a week and did not touch the `transferRoots` db. Is there a reason for that? I believe we can look back two weeks with no real degradation of service (other than missing some settlements for manual bonds), as there should be no action in the happy-path case where the bonder is performing any action past two weeks old. Any thoughts?

packages/hop-node/src/db/GasPricesDb.ts
miguelmota: yes it's intentional (mentioned above)

DeepinScreenshot_select-area_20211004124338

https://kevinsimper.medium.com/how-to-get-range-of-keys-in-leveldb-and-how-gt-and-lt-works-29a8f1e11782

packages/hop-node/src/db/GasPricesDb.ts
miguelmota: getItems will always return an array `[]` should it won't ever be null. `Promise.all` always returns an array

miguelmota:
1. Yes I'm going to add the changes to transferRootsDb as well. The number of transfer roots is really small compared to transfers so want to take a step at a time

packages/hop-node/src/db/TransfersDb.ts
miguelmota: yes it's checked and added during and `.update()`

DeepinScreenshot_select-area_20211004124930

packages/hop-node/src/db/TransfersDb.ts
miguelmota: correct. it's filtered here

DeepinScreenshot_select-area_20211004125040

I'm looking into ways to use another subleveldb for the timestamped keys only

shanefontaine: understood 👍

PR #197, opened 2021-10-06
UI: track destination transaction status

kangarang:
WIP


miguelmota:
Minor style bug

I know you mentioned it but just documenting the minor alignment issue here

DeepinScreenshot_select-area_20211006194314

kangarang:
> Bug
> 
> Expected: The destination tx link to point to destination chain explorer Actual: The destination tx link points to the source chain explorer

@miguelmota nice catch! hm, it seems we do indeed need the destination tx hash for this feature to fully work. for now, we can watch for the `sdk.Event.DestinationTxReceipt`, but this will not work for transactions that are old or if the user reloads the app. at this point, we could optionally query the graph as @cwhinfrey suggested.

thoughts?

miguelmota:
@kangarang works good 👍 Merging.

PR #201, opened 2021-10-06
Fix/optimized transaction timing

shanefontaine:
This PR:

1. Removes `@delay` decorator completely and removes `@queue` decorator from functions that call the GasBoostSigner
2. Adds logging
3. Tracks a local `nonce` and uses it when sending transactions
  * This allows us to remove the 7s delay.

The goal is to remove all waiting when sending a tx.


packages/hop-node/src/gasboost/GasBoostSigner.ts
```diff
@@ -47,26 +56,34 @@ class GasBoostSigner extends Wallet {
...
-      await wait(delta)
+  private async getNonce () {
+    if (!this.nonce) {
+      this.nonce = await this.signer.getTransactionCount()
+    }
+
+    const timeSinceLastTxMs = Date.now() - this.lastTxSentTimestamp
+    if (this.lastTxSentTimestamp && timeSinceLastTxMs > TenMinutesMs) {
+      this.logger.info(`checking on-chain nonce. timeSinceLastTxMs ${timeSinceLastTxMs}`)
+      const onChainNonce = await this.signer.getTransactionCount()
```
miguelmota, line 78: I believe this should be `this.signer.getTransactionCount('pending')`

miguelmota:
Regarding your Slack question:

the GasBoostTransaction keeps tracking of broadcasted ethers transactions and the GasBoostSigner keeps track of GasBoostTransactions. So they each only track things that concern them (for example, the GasBoostSigner doesn't care about tx replacements; that is the concern of GasBoostTransaction.

packages/hop-node/src/gasboost/GasBoostSigner.ts
```diff
@@ -47,26 +56,34 @@ class GasBoostSigner extends Wallet {
...
+      if (onChainNonce !== this.nonce) {
+        this.logger.error(
```
miguelmota, line 80: sending a notifier notification here could be useful

PR #202, opened 2021-10-06
hop-node: TransferRootsDb: Add timestamped keys sublevel db

miguelmota:
Ready for review but I don't want to merge this in yet until the latest updates have been running in production for a while.


shanefontaine:
Changes LGTM 👍

miguelmota:
> would it be possible to DRY this up with the similar logic in ..

Should be possible, will think about it/implement it

PR #204, opened 2021-10-07
hop-node: Remove decorators and use regular functions instead

miguelmota:


miguelmota:
@shanefontaine @cwhinfrey ready for review

PR #211, opened 2021-10-08
Fix/tx block

shanefontaine:


miguelmota:
@shanefontaine left one or two review questions

PR #235, opened 2021-10-15
Prometheus metrics server scaffolding

gsgalloway:
Sets up scaffolding for a metrics server and adds a metric for the bonder's various balances.

StakeWatcher is probably not the right place to update that metric since it seems it only ever loads up the bonder's balances once at startup. Should probably plumb all the watchers or all the bridge contracts directly into the metrics server and let it poll on its own.

Test by running the node as a bonder and then 

```bash
$ curl localhost:8080/metrics

# HELP bonder_balance Balance of the bonder for a particular token and chain
# TYPE bonder_balance gauge
bonder_balance{chain="ethereum",token="USDC"} 1000
bonder_balance{chain="ethereum",token="ETH"} 0.1
```

### DIFF
--- a/packages/hop-node/package.json
+++ b/packages/hop-node/package.json
@@ -79,6 +79,7 @@
     "ethereumjs-util": "^7.0.9",
     "ethers": "5.4.5",
     "expect": "^26.6.2",
+    "express": "^4.17.1",
     "fast-memoize": "^2.5.2",
     "js-yaml": "^4.1.0",
     "keythereum": "1.2.0",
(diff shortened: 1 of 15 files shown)


miguelmota:
Hey @gsgalloway! This is really cool; thanks for kicking this off! It works great with prometheus and grafana. Besides balances, any particular metrics that you think would be good to start monitoring?

miguelmota:
@gsgalloway those are some good suggestions! We'll add in separate commits. Merging this one in. Thanks 👍

PR #251, opened 2021-10-27
Fix/async syncing

shanefontaine:
Some event handlers had dependencies on other event handlers. This PR fixes that by removing all dependencies. This allows for:

1. Faster sync because it is fully async
2. Guaranteed DB state because `allSettled` is now written only after all DB items exist


shanefontaine:
@miguelmota Thanks for the review. Part of the reason for the PR was to fix the DB overwrite issue & I believe that this fix does not address that. Because of this, I am going to close this PR without merging.

The remainder of this PR is an optimization that I believe should be handled down the line.

miguelmota:
👍

PR #259, opened 2021-10-29
UI: fix `/convert` numerical input field

kangarang:
previously, `handleSourceTokenAmountChange` would call `fixedDecimals()`, setting the input to a decimal value on each change.

this is now addressed by calling `sanitizeNumericalString()` only. the appropriate conversion already happens when the tx method is executed (`convertTokens()`).

misc:
- rm `normalizeNumberInput` util (essentially a duplicate of `sanitizeNumericalString`)


miguelmota (approved):
👍

Issue #260, opened 2021-10-30
estimation of sending max value native token is limited to `Send` component


kangarang:
there are 2 `AmountSelectorCard` components -- the one for the `Send` component is able to properly estimate/evaluate the total gas cost of sending transfer txs. however, the other `AmountSelectorCard` is not, and it is used for the `ConvertContent`, `Pools`, `StakeWidget`, and `TokenWrapper` components.

resolution: add a reusable hook `useSendNativeTokenMaxValue` that accepts a `method` and returns an estimated max value a user can send based on the type of intended transaction.

note: it may be simpler and results may be more consistent to use a hardcoded dictionary of `method: gasCost`, and subtract the intended tx's gas cost value (multiplied by the current `gasPrice`) from the user's balance.

kangarang:
only txs sent via `ConvertContent` and `TokenWrapper` need to be adjusted. methods: `convertToken` and `wrapToken`.

PR #263, opened 2021-10-31
ui/sdk: enable max value estimations for `wrapToken` and `convertToken`

kangarang:
this pr enables max value native token tx estimations for methods: `wrapToken` and `convertToken`.

addresses #260. see issue for more details


packages/frontend/src/hooks/useNativeTokenMaxValue.ts
```diff
@@ -0,0 +1,243 @@
...
+    arbitrum: 124_069,
+    optimism: 289_937,
+    polygon: 72_039,
+    xdai: 78_450,
+  },
+  wrapToken: {
+    // destNetwork
+    arbitrum: 964_242,
+    optimism: 65_000_000,
+    polygon: 27_941,
```
kangarang, line 28: i think this value is incorrect

packages/frontend/src/hooks/useNativeTokenMaxValue.ts
```diff
@@ -0,0 +1,243 @@
...
+import { useApp } from 'src/contexts/AppContext'
+import Network from 'src/models/Network'
+import { ZERO_ADDRESS } from 'src/constants'
+
+export const METHOD_NAMES = {
+  convertTokens: 'convertTokens',
+  wrapToken: 'wrapToken',
+}
+
+const estimatedGasLimits = {
```
miguelmota, line 16: I don't think the hard coded gas limits will be reliable because they are dynamic on rollups, since gas limits on ORUs are a function of current L1 gas prices. 

From [optimism docs](https://community.optimism.io/docs/developers/fees.html#transaction-fees-on-l2):

> You should always use the gas limit suggested by `eth_estimateGas`.

Is the idea that the hardcoded limits are only for estimates when a wallet is _not_ connected?

kangarang:
@miguelmota ready for another review!

packages/frontend/src/hooks/useNativeTokenMaxValue.ts
```diff
@@ -0,0 +1,157 @@
...
+import { BigNumber } from 'ethers'
+import { useWeb3Context } from 'src/contexts/Web3Context'
+import logger from 'src/logger'
+import Transaction from 'src/models/Transaction'
+import { Token, TProvider } from '@hop-protocol/sdk'
+import { useApp } from 'src/contexts/AppContext'
+import Network from 'src/models/Network'
+import { ZERO_ADDRESS } from 'src/constants'
+
+export const METHOD_NAMES = {
```
miguelmota, line 11: This would be better as an enum

```ts
enum MethodNames {
    convertTokens = 'convertTokens',
    wrapToken = 'wrapToken'
}
```

miguelmota:
Hey @kangarang, I encountered some issues testing it out

- Send page max button is not working
DeepinScreenshot_select-area_20211103203309
- Add liquidity max button is not working
DeepinScreenshot_select-area_20211103203352
- Stake page max button is not working
DeepinScreenshot_select-area_20211103204300

The wrap/unwrap component max button works as expected

I made sure to rebuild the sdk and lerna linking before running the frontend. 

Happy to provide more info

kangarang:
@miguelmota resolved and ready for another review please!

miguelmota:
Hey @kangarang things are looking good!

There's a small issue when no destination network is selected on send page;

Expected: clicking source amount max button when asset is not native to the source chain should fill in the max token amount.
Actual: the source amount max button has no effect when source chain is selected and the transfer asset is a non-native asset to the source chain.

DeepinScreenshot_select-area_20211105181110

kangarang:
@miguelmota, great catch! i've adjusted it so that if the dest network is not selected, a warning will show if the token is a native asset. if the token is not a native asset while the dest network is not selected, the field will use the source network balance. thanks for the review

miguelmota (approved):
@kangarang works good! A nice to have would be to add a loading indicator (or gray out) the max button when it's taking time to get the calculated number. Good to merge

PR #267, opened 2021-11-03
hop-node: optimize gas-cost worker for debugability

shanefontaine:
This PR will help with debugging. We can implement possible `fetch` rate limits in a future PR, if we desire.

### DIFF
--- a/packages/hop-node/src/watchers/SyncWatcher.ts
+++ b/packages/hop-node/src/watchers/SyncWatcher.ts
@@ -1276,16 +1276,16 @@ class SyncWatcher extends BaseWatcher {
 
   async pollGasCost () {
     const bridgeContract = this.bridge.bridgeContract.connect(getRpcProvider(this.chainSlug)!) as L1BridgeContract | L2BridgeContract // eslint-disable-line @typescript-eslint/no-non-null-assertion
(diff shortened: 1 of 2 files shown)


packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -1275,20 +1275,20 @@ class SyncWatcher extends BaseWatcher {
...
+    const bridgeContract = this.bridge.bridgeContract.connect(getRpcProvider(this.chainSlug)!) as L1BridgeContract | L2BridgeContract // eslint-disable-line @typescript-eslint/no-non-null-assertion
+    const amount = BigNumber.from(10)
+    const amountOutMin = BigNumber.from(0)
+    const bonderFee = BigNumber.from(1)
+    const bonder = this.bridge.getConfigBonderAddress()
+    const recipient = `0x${'1'.repeat(40)}`
+    const transferNonce = `0x${'0'.repeat(64)}`
+
     while (true) {
+      const txOverrides = await this.bridge.txOverrides()
```
miguelmota, line 1287: I think this async function should be in the try/catch

miguelmota (approved):
One minor comment about moving async call to try/catch but everything else LGTM

PR #274, opened 2021-11-10
hop-node: prep for regenesis

shanefontaine:
DO NOT MERGE.

This is an example PR to prep for the Optimism regenesis. If the code makes sense, we can cherrypick commits into the bonder from develop.

Changes are:
1. Update `bridgeDeployedBlockNumber` to `0` in `core`
2. Turn off Optimism watchers
3. On the first sync, reset the Optimism `latestBlockSynced` to `0` for just the first run
    * This is meant to be removed the next time the bonder restarts

### DIFF
--- a/packages/hop-node/src/watchers/classes/Bridge.ts
+++ b/packages/hop-node/src/watchers/classes/Bridge.ts
@@ -36,7 +36,6 @@ export default class Bridge extends ContractBase {
   bridgeDeployedBlockNumber: number
   l1CanonicalTokenAddress: string
   stateUpdateAddress: string
-  isFirstIteration: boolean = true
 
   constructor (bridgeContract: BridgeContract) {
     super(bridgeContract)
@@ -553,14 +552,6 @@ export default class Bridge extends ContractBase {
         options.cacheKey
       )
       state = await this.db.syncState.getByKey(cacheKey)
-    }
-
-    if (cacheKey && this.isFirstIteration && this.chainSlug === Chain.Optimism) {
-      await this.db.syncState.update(cacheKey, {
-        latestBlockSynced: 0,
-        timestamp: Date.now()
-      })
-      this.isFirstIteration = false
     }
 
     let {
--- a/packages/hop-node/src/watchers/classes/Bridge.ts
+++ b/packages/hop-node/src/watchers/classes/Bridge.ts
@@ -36,6 +36,7 @@ export default class Bridge extends ContractBase {
   bridgeDeployedBlockNumber: number
   l1CanonicalTokenAddress: string
   stateUpdateAddress: string
+  isFirstIteration: boolean = true
 
   constructor (bridgeContract: BridgeContract) {
     super(bridgeContract)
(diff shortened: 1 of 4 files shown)


packages/hop-node/src/watchers/classes/Bridge.ts
```diff
@@ -554,6 +555,14 @@ export default class Bridge extends ContractBase {
       state = await this.db.syncState.getByKey(cacheKey)
     }
 
+    if (cacheKey && this.isFirstIteration && this.chainSlug === Chain.Optimism) {
```
miguelmota, line 558: I think the db can be reset on start instead of on each eventsBatch, so the code would look like this to reset all entries that have optimism as the key prefix

```js
class SyncStateDb extends BaseDb {
  ready: boolean = false
  constructor (prefix: string, _namespace?: string) {
    super(prefix, _namespace)
    this.migrations()
      .then(() => {this.ready = true})
      .catch(this.logger.error)
  }

  async migrations () {
    const items = await this.getKeyValues()
    for (let { key, value } of items) {
      if (key?.startsWith(`${chainSlugToId(Chain.Optimism)}`)) {
        await this._update(key, Object.assign({}, value, {
          latestBlockSynced: 0,
          timestamp: Date.now()
        }))
        this.logger.debug(`${key} updated latestBlockSynced to 0`)
      }
    }
  }

  async update (key: string, data: Partial<State>) {
    await this.tilReady()
    ...
  }

  async getItems (): Promise<State[]> {
    await this.tilReady()
    ...
  }

 ...
}
```

miguelmota (commented):
The sdk will have to get bumped as well

packages/hop-node/src/watchers/classes/Bridge.ts
shanefontaine: I understand and think this makes a ton of sense. Update is 87dddef4922c0967b065b1a316abfb91c8a5faa3. Two slight differences to just bring to your attention:

1. `startsWith` for Optimism is chain `10` & for xDai is `100`, so I added a `:` to the end of the `startsWith()` so that xDai doesn't get picked up.
2. `tilReady` is not in `BaseDb` but rather in `TimestampedKeysDb`. I duplicated it & added to SyncWatcher since this is just a temp fix. I _think_ it makes sense to eventually move it to `BaseDb`, but I'm not sure of the implications of it right now. Can do it soon if you think it makes sense to be in base.

miguelmota: 1. nice, good looking out for that 
2. there shouldn't be any implications; it's safe to move it to BaseDb and remove it from TimestampedKeysDb

shanefontaine: Perfect. Thanks 👍

miguelmota: One last thing, the `getByKey` is missing the `tilReady` check (forgot to add it in example)

```js
  async getByKey (key: string): Promise<State> {
    await this.tilReady()
    ...
  }

shanefontaine: Nice catch. Will add 👍

shanefontaine:
This PR was replaced by #277

PR #275, opened 2021-11-10
frontend: Disable Optimism during regenesis

miguelmota:
Do not merge until regenesis is about to occur (Nov 11 9am PT)

DeepinScreenshot_select-area_20211110124840

- Disables Optimism completely as an option
- Adds explainer banner


shanefontaine (approved):
@miguelmota Looks good to me 👍

PR #278, opened 2021-11-12
UI: smarter button states

kangarang:
this pr prevents accidental double clicks for the modal components in `TxConfirm`.

additionally, readability is improved for the `Send` component, and tx confirm component names which were previously named the same as others are renamed with the prefix `Confirm-`.


miguelmota (approved):
Nice clean up 👍 LGTM

PR #282, opened 2021-11-16
UI: tx page

kangarang:
this pr adds a `/tx/:hash` route where a user can see a data dump for a specified tx hash.

- added some utils
- refactored some methods
- shuffled around some components for better organization

Screen Shot 2021-11-15 at 9 14 26 PM


packages/frontend/package.json
```diff
@@ -17,6 +17,7 @@
     "classnames": "2.2.6",
     "clsx": "1.1.1",
     "copy-to-clipboard": "^3.3.1",
+    "dayjs": "^1.10.7",
```
kangarang, line 20: not actually using this yet, but have plans to

packages/frontend/src/pages/Transaction/TransactionPage.tsx
```diff
@@ -0,0 +1,232 @@
...
+    console.log(`txObj:`, txObj)
+  }, [txObj])
+
+  async function handleSetNetwork(network) {
+    if (txHash && network) {
+      setSourceNetwork(network)
+    }
+  }
+
+  async function addToken() {
```
kangarang, line 54: this only works sometimes :/

will eventually turn into a hook

packages/frontend/src/hooks/useTransaction.ts
```diff
@@ -0,0 +1,238 @@
...
+  methodName?: string
+  tokenSymbol?: string
+  params?: any[]
+  eventValues?: any
+  destNetworkName?: string
+  destTxHash?: string
+  destExplorerLink?: string
+}
+
+const useTransaction = (txHash?: string, slug?: string) => {
```
kangarang, line 67: another iteration of this hook is what i imagine would be a nice complement to the `Transaction` model

packages/frontend/src/utils/transactions.ts
```diff
@@ -7,3 +7,14 @@ export const sortByRecentTimestamp = (txs: Transaction[]) => {
 export function filterByHash(txs: Transaction[] = [], hash: string = '') {
   return txs.filter(tx => tx.hash !== hash)
 }
+
+export async function queryFilterTransferFromL1CompletedEvents(bridge, networkName) {
+  const destL2Bridge = await bridge.getL2Bridge(networkName)
+  const bln = await destL2Bridge.provider.getBlockNumber()
+  const evs = await destL2Bridge.queryFilter(
+    destL2Bridge.filters.TransferFromL1Completed(),
+    bln - 9999,
```
miguelmota, line 16: We can't assume that a 10k block range will always work because certain rpc endpoints only allow range look ups of `1k` blocks, so pagination would have to happen here

packages/frontend/src/pages/Transaction/TransactionPage.tsx
```diff
@@ -0,0 +1,232 @@
...
+      const { tokenSymbol } = txObj
+      const tokenImageUrl = metadata.tokens[tokenSymbol].image
+      const tokenModel = sdk.toTokenModel(txObj.tokenSymbol)
+      const networkName = networkIdToSlug(connectedNetworkId)
+      if (networkName === 'ethereum') {
+        return
+      }
+      console.log(`tokenImageUrl:`, tokenImageUrl)
+      const addr = tokens[tokenSymbol][networkName].l2CanonicalToken.address
+      ;(window as any).ethereum.request({
```
miguelmota, line 66: are we assuming here that `;(window as any).ethereum` will always be available?

packages/frontend/src/pages/Transaction/TransactionPage.tsx
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  async function addToken() {
+    if (txObj?.tokenSymbol) {
+      console.log(`tokens['DAI']:`, (tokens as any).DAI)
+      const { tokenSymbol } = txObj
+      const tokenImageUrl = metadata.tokens[tokenSymbol].image
+      const tokenModel = sdk.toTokenModel(txObj.tokenSymbol)
+      const networkName = networkIdToSlug(connectedNetworkId)
+      if (networkName === 'ethereum') {
```
miguelmota, line 61: I know we check for string network slugs in multiple places but an improvement would be use an enum

```typescript
if (networkName === Chain.Ethereum) {
```

enum type

```typescript
export enum Chain {
  Ethereum = 'ethereum',
  Optimism = 'optimism',
  Arbitrum = 'arbitrum',
  Polygon = 'polygon',
  xDai = 'xdai'
}
```

packages/frontend/src/pages/Transaction/TransactionPage.tsx
```diff
@@ -0,0 +1,232 @@
...
+    if (txHash && network) {
+      setSourceNetwork(network)
+    }
+  }
+
+  async function addToken() {
+    if (txObj?.tokenSymbol) {
+      console.log(`tokens['DAI']:`, (tokens as any).DAI)
+      const { tokenSymbol } = txObj
+      const tokenImageUrl = metadata.tokens[tokenSymbol].image
```
miguelmota, line 58: The metadata info can be abstracted into a util getter 

```js
const tokenImageUrl = getTokenImage(tokenSymbol)
```

packages/frontend/src/utils/queries.ts
```diff
@@ -5,6 +5,10 @@ export function getUrl(chain) {
   if (chain === 'ethereum') {
     chain = 'mainnet'
   }
+
+  if (chain === 'optimism') {
+    chain = 'optimism1'
```
miguelmota, line 10: we can actually use `'hop-optimism` again now (no need for `hop-optimism1` anymore)

packages/frontend/src/utils/getRpcUrl.ts
```diff
@@ -1,7 +1,24 @@
...
 
+export function getRpcUrlByNetworkName(networkName: string) {
+  let rpcUrl = ''
+  if (networkName.startsWith('arbitrum')) {
+    rpcUrl = getRpcUrl('arbitrum')
+  } else if (networkName.startsWith('optimism')) {
+    rpcUrl = getRpcUrl('optimism')
+  } else if (networkName.startsWith('xdai')) {
+    rpcUrl = getRpcUrl('xdai')
+  } else if (networkName.startsWith('matic') || networkName.startsWith('polygon')) {
```
miguelmota, line 16: I think it's safe to remove `networkName.startsWith('matic')` and just have `networkName.startsWith('polygon')`

packages/frontend/src/pages/Transaction/TransactionPage.tsx
```diff
@@ -0,0 +1,232 @@
...
+      const tokenModel = sdk.toTokenModel(txObj.tokenSymbol)
+      const networkName = networkIdToSlug(connectedNetworkId)
+      if (networkName === 'ethereum') {
+        return
+      }
+      console.log(`tokenImageUrl:`, tokenImageUrl)
+      const addr = tokens[tokenSymbol][networkName].l2CanonicalToken.address
```
miguelmota, line 65: since the sdk is already being imported here, then we can call the sdk method to get canonical token address here instead of doing map lookup

```typescript
const addr = sdk.getL2CanonicalTokenAddress(tokenSymbol, networkName)
```

packages/frontend/src/hooks/useTransaction.ts
```diff
@@ -0,0 +1,238 @@
...
+          }
+
+          setTxObj({
+            type: txType,
+            response,
+            receipt,
+            methodName,
+            params,
+            gasCost,
+            datetime: new Date(ts * 1000).toString(),
```
miguelmota, line 207: are we assuming that `ts` is always defined here?

packages/frontend/src/hooks/useTransaction.ts
```diff
@@ -0,0 +1,238 @@
...
+          }
+
+          const { data, gasPrice } = response
+          const { gasUsed } = receipt
+          const gasCost = formatEther(gasUsed.mul(gasPrice!))
+
+          const funcSig = data.slice(0, 10)
+          const methodName = sigHashes[funcSig]
+
+          setLoading(false)
```
miguelmota, line 100: would it be better to set `setLoading(false)` at the _end_ of the function once all async calls have finished?

packages/frontend/src/models/Transaction.ts
```diff
@@ -133,6 +118,53 @@ class Transaction extends EventEmitter {
...
 
+  getRpcUrl(networkName): string {
+    let rpcUrl = ''
+    if (networkName.startsWith('arbitrum')) {
+      rpcUrl = getRpcUrl('arbitrum')
+    } else if (networkName.startsWith('optimism')) {
+      rpcUrl = getRpcUrl('optimism')
+    } else if (networkName.startsWith('xdai')) {
+      rpcUrl = getRpcUrl('xdai')
+    } else if (networkName.startsWith('matic') || networkName.startsWith('polygon')) {
```
miguelmota, line 129: I think it's safe to remove `networkName.startsWith('matic')` and just have `networkName.startsWith('polygon')`

packages/frontend/src/utils/getExplorerUrl.ts
```diff
@@ -0,0 +1,29 @@
...
+      } catch (err) {
+        return ''
+      }
+    }
+    case 'dai':
+    case 'xdai': {
+      return `${getBaseExplorerUrl('xdai')}/tx/${txHash}`
+    }
+    case 'polygon':
+    case 'matic': {
```
miguelmota, line 25: I think it's safe to remove `'matic'` and just have `'polygon'`

packages/frontend/src/utils/getExplorerUrl.ts
```diff
@@ -0,0 +1,29 @@
...
+    case 'optimism': {
+      try {
+        const url = new URL(getBaseExplorerUrl('optimism'))
+        return `${url.origin}${url.pathname}/tx/${txHash}${url.search}`
+      } catch (err) {
+        return ''
+      }
+    }
+    case 'dai':
+    case 'xdai': {
```
miguelmota, line 21: the network should always be `'xdai'` so we can remove `'dai'` here

packages/frontend/src/utils/logs.ts
```diff
@@ -3,6 +3,54 @@ import find from 'lodash/find'
...
+
+    return {
+      ...acc,
+      [arg]: argValue,
+    }
+  }, {})
+  return formatted
+}
+
+export function getLastLog(logs: any[]) {
```
miguelmota, line 51: A check for `logs.length > 0` would add additional safety here

miguelmota:
A few bugs around the data were encountered

---

arbitrum tx: `0xa27b55e2a1ef3fadd38496cd9cd6eea07e2784f1c9367dfeaa5e607ae91f6440`

Token symbol is missing

DeepinScreenshot_select-area_20211116195906

The date here is invalid

DeepinScreenshot_select-area_20211116195952

---

polygon tx: `0xde78af099466429d0f286ab8098295813ca56705679f83c3aa0e6de2424c920d`

The native token here should be `MATIC` instead of `ETH`

DeepinScreenshot_select-area_20211116200135


---

xdai tx: `0xa10cedbefc66de7e1c28c8c77bb4c818d8c562c7c10c33450c4756a688879cf1`

The native Token here should be `XDAI` instead of `ETH`

DeepinScreenshot_select-area_20211116200241

---

These call data parameters can be formatted using the transferred asset token decimals

DeepinScreenshot_select-area_20211116200605

miguelmota:
Nice to haves: 

- additional data around bond tx, ie 'bonded within', bonder address, value, gas cost, etc
- a 'search' input field at `/tx` route to allow user to input either: source tx hash, destination tx hash, or transfer ID

I think what you're building will be used by a lot of users! and hopefully reduce discord support messages

kangarang:
@miguelmota thank you for the great code review! will address your comments in the next set of commits (which also greatly cleans up the `useTransaction` hook)

packages/frontend/src/pages/Transaction/TransactionPage.tsx
kangarang: will use `sdk.Chain.Ethereum.slug`

packages/frontend/src/hooks/useTransaction.ts
kangarang: will change to `new Date(parseInt(timestamp || '0') * 1000).toString()`

packages/frontend/src/hooks/useTransaction.ts
kangarang: i experienced some delays in especially the later async calls, so i'm setting loading to false early on. additionally, if a tx was recent, or is pending being bonded, or just hasn't been mined yet, we still want to see the data that *has* been loaded so far so we can get an idea of what the current status is.

kangarang:
@miguelmota for xdai tx: `0xa10cedbefc66de7e1c28c8c77bb4c818d8c562c7c10c33450c4756a688879cf1`:

able to see conversions here:

Screen Shot 2021-11-18 at 5 14 05 PM

the bottom link will redirect to here, where we get more insight on the bond tx:

Screen Shot 2021-11-18 at 5 14 13 PM

^this is still WIP. i need to add a util to grab the transferred asset by address and configure the page for more reliable state initiation

packages/frontend/src/models/Transaction.ts
```diff
@@ -133,6 +118,53 @@ class Transaction extends EventEmitter {
     }
   }
 
+  getRpcUrl(networkName): string {
+    let rpcUrl = ''
+    if (networkName.startsWith('arbitrum')) {
```
miguelmota, line 123: for consistency, these should be changed to use chain enum

```js
import { Chain } from 'src/utils/constants'
```

miguelmota (approved):

packages/frontend/src/utils/getRpcUrl.ts
```diff
@@ -1,7 +1,35 @@
+import { L1_NETWORK } from 'src/constants'
 import { networks } from 'src/config'
+import { Chain } from './constants'
 
 export const getRpcUrl = (network: string) => {
   return networks[network]?.rpcUrl
 }
 
+export function getRpcUrlByNetworkName(networkName: string) {
```
miguelmota, line 9: What is the purpose of this method if `getRpcUrl` already exists?

kangarang: haha! good point. i copy-pasted this block of code from `Transaction.tsx`

PR #284, opened 2021-11-16
hop-node: do not block CLI bond-withdrawal command

shanefontaine:
A change we made a while ago blocked the ability to bond withdrawals with the CLI. This was because the bonder fee check needed the context of a `gasCost` DB, however the CLI should not care about that, I believe.

This PR logically reorders a few items in the `bondWithdrawalWatcher` and calls `sendBondWithdrawalTx()` directly in the CLI command.


miguelmota (approved):
Looks good 👍

PR #286, opened 2021-11-18
Feature/remove stake watcher

shanefontaine:
This PR simplifies some of the legacy staking logic. Some features/designs are slightly different than the original:

* Removes concept of stakeWatcher
* Introduces a `--skip-send-to-l2` flag to explicitly skip sending tokens to L2 instead of being implicit about it

Some considerations:
* I don't think the bridge instantiation is optimal. It works, but let me know if there is a more optimal way to do this with the current architecture (i.e., no explicit stake watcher)
  * For example, an arbitrary watcher is instantiated in order to retrieve the bridge.


packages/hop-node/src/cli/stake.ts
```diff
@@ -1,66 +1,137 @@
...
+async function stake (
+  bridge: L2Bridge | L1Bridge,
+  parsedAmount: BigNumber
 ) {
-  if (!chain) {
+  logger.debug('Staking')
+
+  const token: Token | void = await getToken(bridge)
+  const stakeTokenBalance: BigNumber = await getTokenBalance(bridge, token)
+  const formattedAmount = bridge.formatUnits(parsedAmount)
```
miguelmota, line 50: do decimals need to be passed here for `formatUnits`?

packages/hop-node/src/cli/stake.ts
```diff
@@ -80,7 +152,33 @@ program
...
+      const skipSendToL2 = source.skipSendToL2
+
+      if (!amount) {
+        throw new Error('amount is required. E.g. 100')
+      }
+      if (!chain) {
+        throw new Error('chain is required')
+      }
+      const bridge: L2Bridge | L1Bridge = await getBridge(token, chain)
+      const parsedAmount: BigNumber = bridge.parseUnits(amount)
```
miguelmota, line 164: do decimals need to be passed here for `parseUnits`?

packages/hop-node/src/cli/cli.ts
```diff
@@ -22,7 +22,6 @@ import './send'
 import './settle'
 import './showConfig'
 import './stake'
-import './stakeStatus'
```
miguelmota, line 25: why was the stake status command removed?

packages/hop-node/src/cli/stake.ts
```diff
@@ -1,66 +1,137 @@
...
+  if (!token) {
+    return bridge.getEthBalance()
   }
+  return token.getBalance()
+}
 
-  const watchers = getStakeWatchers({
+async function getBridge (token: string, chain: string): Promise<L2Bridge | L1Bridge> {
+  // Arbitrary watcher since only the bridge is needed
+  const watchers = await getWatchers({
```
miguelmota, line 122: Why not just get the bridge directly here?

```typescript
import contracts from 'src/contracts'

const tokenContracts = contracts.get(token, chain)
return chain === Chain.Ethereum ? tokenContracts.l1Bridge : tokenContracts.l2Bridge
```

packages/hop-node/src/cli/unstake.ts
```diff
@@ -24,7 +61,33 @@ program
...
+
+      if (!amount) {
+        throw new Error('amount is required. E.g. 100')
+      }
+      if (!chain) {
+        throw new Error('chain is required')
+      }
+
+      // Arbitrary watcher since only the bridge is needed
+      const watchers = await getWatchers({
```
miguelmota, line 73: Why not just get the bridge directly here?

```typescript
import contracts from 'src/contracts'

const tokenContracts = contracts.get(token, chain)
return chain === Chain.Ethereum ? tokenContracts.l1Bridge : tokenContracts.l2Bridge
```

miguelmota:
feedback comments @shanefontaine

packages/hop-node/src/cli/unstake.ts
shanefontaine: It is my understanding that there is no provider associated with the bridge when we do that. Is that correct and if so, how would the provider become associated with this instance?

packages/hop-node/src/cli/stake.ts
shanefontaine: It is my understanding that the `bridge` [does this already](https://github.com/hop-protocol/hop/blob/develop/packages/hop-node/src/watchers/classes/Bridge.ts#L506-L512) with the `token` associated with the bridge itself.

packages/hop-node/src/cli/stake.ts
shanefontaine: It is my understanding that the `bridge` [does this already](https://github.com/hop-protocol/hop/blob/develop/packages/hop-node/src/watchers/classes/Bridge.ts#L506-L512) with the `token` associated with the bridge itself.

miguelmota: you're right. Disregard comment

packages/hop-node/src/cli/stake.ts
miguelmota: you're right. Disregard comment

packages/hop-node/src/cli/unstake.ts
miguelmota: I was confused thinking it was referencing the bridge contract and not the bridge class. You can disregard comment

packages/hop-node/src/cli/stake.ts
miguelmota: I was confused thinking it was referencing the bridge contract and not the bridge class. You can disregard comment

packages/hop-node/src/cli/cli.ts
shanefontaine: Reintroduced with adf38fdc40b6963969631da597912c2148487588

PR #287, opened 2021-11-18
Use OVM_GasPriceOracle for Optimism tx fee estimation

miguelmota:


packages/hop-node/src/watchers/classes/Bridge.ts
```diff
@@ -752,7 +755,24 @@ export default class Bridge extends ContractBase {
...
+        const ovmGasPriceOracle = getContractFactory('OVM_GasPriceOracle')
+          .attach(predeploys.OVM_GasPriceOracle).connect(getRpcProvider(this.chainSlug))
+        const serializedTx = serializeTransaction({
+          value: parseEther('0'),
+          gasPrice,
+          gasLimit,
+          to,
+          data
+        })
+        gasCost = await ovmGasPriceOracle.getL1Fee(serializedTx)
```
shanefontaine, line 771: I think this value should be added to the existing `gasCost`. IIUC, this value provides the L1 cost, in Wei and needs to be added to our existing calc, `gasCost = gasLimit.mul(gasPrice)`, which is the L2 cost.

packages/sdk/src/HopBridge.ts
```diff
@@ -665,13 +666,21 @@ class HopBridge extends Base {
...
+        const ovmGasPriceOracle = getContractFactory('OVM_GasPriceOracle')
+          .attach(predeploys.OVM_GasPriceOracle).connect(destinationChain.provider)
+        const serializedTx = serializeTransaction({
+          value: parseEther('0'),
+          gasPrice,
+          gasLimit: bondTransferGasLimit,
+          to,
+          data
+        })
+        txFeeEth = await ovmGasPriceOracle.getL1Fee(serializedTx)
```
shanefontaine, line 680: Same comment as hop-node.

shanefontaine (commented):
@miguelmota back to you

packages/sdk/src/HopBridge.ts
miguelmota: Fixed with https://github.com/hop-protocol/hop/pull/287/commits/5f60e11bb197451ff9f6d15ea3edc644dab3f5b0

packages/hop-node/src/watchers/classes/Bridge.ts
miguelmota: Fixed with https://github.com/hop-protocol/hop/pull/287/commits/5f60e11bb197451ff9f6d15ea3edc644dab3f5b0

PR #288, opened 2021-11-18
sdk: fix gas estimation for L2 -> L1 txs

kangarang:
previously, `getBondWithdrawalEstimatedGas` would set `destinationBridge` as the the `L2Bridge` of the destination chain. however, if the destination chain was mainnet ethereum, no bridge would be set.

this is now addressed by checking if the destination bridge is L1, and using the `L1Bridge` as the `destinationBridge`


miguelmota:
👍

PR #292, opened 2021-11-19
UI: improve tx history

kangarang:
- remove tx if timestamp is older than 1 hour
- replace tx if a recent one is found with the same nonce from same address
- use effect to update Header TxPill pending tx count


packages/frontend/src/utils/blocks.ts
```diff
@@ -0,0 +1,29 @@
...
+export async function getRecentTransactionsByFromAddress(
+  provider: providers.Provider,
+  fromAddress: string
+) {
+  const blocks = await getRecentBlocks(provider)
+  const txs: providers.TransactionResponse[] = []
+
+  for (const block of blocks) {
+    for (const transaction of block.transactions) {
+      if (transaction.from === fromAddress) {
```
miguelmota, line 22: Should the addresses here be lowercased to make sure the casing is the same when comparing?

packages/frontend/src/models/Transaction.ts
```diff
@@ -94,7 +106,7 @@ class Transaction extends EventEmitter {
       const tsDetails = getTransferSentDetailsFromLogs(receipt.logs)
       this.blockNumber = receipt.blockNumber
       const block = await this.provider.getBlock(receipt.blockNumber)
-      this.timestamp = block.timestamp
+      this.timestamp = block.timestamp * 1000
```
miguelmota, line 109: to avoid confusion, maybe this should be called `timestampMs`?

miguelmota (approved):

packages/frontend/src/utils/blocks.ts
kangarang: it _should_ be ok, since both are checksummed, but i agree we should for best practices

PR #293, opened 2021-11-21
hop-node: normalize exit tx function sig and only require commit tx hash

shanefontaine:
This PR adds two features:

1. Allow the `exit-commit-tx` CLI command to work without the need for a local DB
2. Standardize all exit watchers with a `relayXDomainMessage()` function that only takes the commit tx hash

One other change was to only allow for a single exit on the xDai watcher per run. This allows us to standardize all the watchers. The reason this will work is because there should never be two `CommitTransfers` for the same token bridge within the range we search for (3 blocks). The contract [does not allow people to call the `commit` function more than once per few hours](https://github.com/hop-protocol/contracts/blob/v1/contracts/bridges/L2_Bridge.sol#L180). It could technically be called by sending 512 txs and auto-triggering it, but even if a chain has a block `gasLimit` of 30 million (I think xDai only has 20 million), a user could only fire off 120 txs per block (assuming 250k gas per tx), or 360 per 3 blocks if they use up the entire block, which is still not enough to cause any issues.


packages/hop-node/src/watchers/xDaiBridgeWatcher.ts
```diff
@@ -101,89 +56,110 @@ class xDaiBridgeWatcher extends BaseWatcher {
...
       }
+      const { tx } = result
+      const msg = `sent chainId ${this.bridge.chainId} confirmTransferRoot L1 exit tx ${tx.hash}`
+      logger.info(msg)
+      this.notifier.info(msg)
     } catch (err) {
-      this.logger.error('xDai bridge watcher error:', err)
-      this.quit()
+      logger.error(err)
+      throw err
```
miguelmota, line 84: if the error is being propagated up and then logged then it's going to get logged twice

- suggestion: remove try/catch and logger.error, or add additional context to the logger.error here

packages/hop-node/src/watchers/xDaiBridgeWatcher.ts
```diff
@@ -101,89 +56,110 @@ class xDaiBridgeWatcher extends BaseWatcher {
...
-    const destinationChainId = dbTransferRoot?.destinationChainId
-    const l2Amb = getL2Amb(this.tokenSymbol)
+  async relayXDomainMessage (commitTxHash: string) {
+    const token: string = this.tokenSymbol
+    const l1Amb = getL1Amb(token)
+    const l2Amb = getL2Amb(token)
+
+    const sigEvent = await this.getValidSigEvent(commitTxHash)
+    if (!sigEvent) {
+      this.logger.error(`sigEvent not found for ${commitTxHash}`)
```
miguelmota, line 95: should there a be a return statement here?

packages/hop-node/src/watchers/xDaiBridgeWatcher.ts
```diff
@@ -101,89 +56,110 @@ class xDaiBridgeWatcher extends BaseWatcher {
...
+    }
+
+    this.logger.info('found sigEvent event')
+    const message = sigEvent!.args.encodedData
+    if (!message) {
+      this.logger.error(`message not found for ${commitTxHash}`)
```
miguelmota, line 101: should there be a return statement here?

packages/hop-node/src/watchers/xDaiBridgeWatcher.ts
```diff
@@ -101,89 +56,110 @@ class xDaiBridgeWatcher extends BaseWatcher {
...
+    }
+
+    const messageId =
+      '0x' +
+      Buffer.from(strip0x(message), 'hex')
+        .slice(0, 32)
+        .toString('hex')
+    const alreadyRelayed = await l1Amb.relayedMessages(messageId)
+    if (alreadyRelayed) {
+      return
```
miguelmota, line 119: should we log that it's already been relayed here?

miguelmota (approved):

packages/hop-node/src/watchers/xDaiBridgeWatcher.ts
shanefontaine: Good call. We are using a `try...catch` for Optimism to explicitly handle certain errors. Since we are not explicitly handling errors in other watchers, I removed the `try...catch`es from them in eea40482dd494755352531db496125963da10f27

packages/hop-node/src/watchers/xDaiBridgeWatcher.ts
shanefontaine: Addressed in top level comment below.

packages/hop-node/src/watchers/xDaiBridgeWatcher.ts
```diff
@@ -101,89 +56,110 @@ class xDaiBridgeWatcher extends BaseWatcher {
...
+    const l1Amb = getL1Amb(token)
+    const l2Amb = getL2Amb(token)
+
+    const sigEvent = await this.getValidSigEvent(commitTxHash)
+    if (!sigEvent) {
+      this.logger.error(`sigEvent not found for ${commitTxHash}`)
+    }
+
+    this.logger.info('found sigEvent event')
+    const message = sigEvent!.args.encodedData
```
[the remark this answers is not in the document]
shanefontaine: Good call. Handled 58e6ced549adcdc0d1576df35875bd039f679e64

packages/hop-node/src/watchers/xDaiBridgeWatcher.ts
shanefontaine: Addressed in top level comment below.

packages/hop-node/src/watchers/xDaiBridgeWatcher.ts
shanefontaine: Updated to throw. Addressed in top level comment below.

shanefontaine:
@miguelmota Thanks for the comments. I updated all the exit watchers to `throw` in the case that something goes wrong (such as `already exited` or `exit not found`) instead of return + log. In this case, the top-level `pollHandler()` will catch it and log it.

I did this because, in the happy path case, that watcher should not produce any errors. If it does, we should throw them as such.  Do you think this makes sense?

miguelmota:
Makes sense and looks good @shanefontaine

PR #296, opened 2021-11-23
UI: netlify config

kangarang:
this pr updates the netlify config and cleans up the exported values in `addresses.ts`.

list of blocked countries is still first draft.

this _should_ just work w/o any further config changes in netlify's portal.


packages/frontend/.env.production
```diff
@@ -0,0 +1,5 @@
+REACT_APP_NETWORK="mainnet"
```
miguelmota, line 1: Why is this `.env.production` committed? Env vars should be set on netlify dashboard

miguelmota (approved):

PR #298, opened 2021-11-24
frontend: Show estimated received in send confirmation modal

miguelmota:


miguelmota:
DeepinScreenshot_select-area_20211124123019

DeepinScreenshot_select-area_20211124123008

kangarang (changes requested):
https://user-images.githubusercontent.com/18040654/143314702-2ccd0d0a-ea79-4215-a104-a7012806b741.mov

the alignment isn't a real issue (it happens because of the scroll bar), but we should reduce the font size and align-right for smaller screens. other than that, i think this is a great change

shanefontaine:
Those images LGTM!

miguelmota:
@kangarang good suggestion

PR #299, opened 2021-11-27
hop-node: introduce a per-asset per-chain fee structure

shanefontaine:
Allows the SDK and the bonder to define fees on a per-asset, per-chain basis


miguelmota (approved):
LGTM

PR #300, opened 2021-11-27
hop-node: handle ETH in token class

shanefontaine:
Handles ETH in the token class. Fixes some CLI issues. Nothing major.


packages/hop-node/src/watchers/classes/Token.ts
```diff
@@ -5,34 +5,47 @@ import { formatUnits, parseUnits } from 'ethers/lib/utils'
 
 export default class Token extends ContractBase {
   tokenContract: ERC20
+  isEth: boolean
```
miguelmota, line 8: minor nitpick

the condition is already a boolean so the assignment can be shortened to

```js
this.isEth = (this.tokenContract.address === constants.AddressZero)
```

miguelmota (approved):
LGTM

PR #302, opened 2021-11-30
Feature/ignore invalid transfers and roots

shanefontaine:
1. Standardize known incomplete items into a single array.
2. Reintroduces `isNotFound` DB key transfers and transferRoots. Marks items as `isNotFound` if they were not found after an explicit lookup for the item.
3. Ignores incomplete items or `isNotFound` items when trying to fill in incomplete items.
4. Removes arbitrary-chain lookups for certain events (`populateTransferSentEvent()` and `populateTransferRootCommittedEvent()`). These existed to handle the case where DB overwrites were happening, which has since been fixed.
5. Mark any `populate...()` function item as `isNotFound` if the entry does not exist on the first lookup.
6. When the contract state for a transfer or root did not match our DB, we historically tried to backfill it ([BondWithdrawalWatcher here](https://github.com/hop-protocol/hop/blob/b5c563ce8628ed6191ed649696d8d7fc5e372302/packages/hop-node/src/watchers/BondWithdrawalWatcher.ts#L129) and in the [BondTransferRootWatcher here](https://github.com/hop-protocol/hop/blob/b5c563ce8628ed6191ed649696d8d7fc5e372302/packages/hop-node/src/watchers/BondTransferRootWatcher.ts#L110)). Now, we just mark the item as `isNotFound`.

Items 5 & 6 are the biggest change. They assume that we will find the on-chain data (timestamp, tx) on the first try (or retry x5) or else we will ignore it going forward. 

 I think this is a valid update because it tremendously simplifies the whole process and the flow becomes more logical (sync -> fill in incomplete -> process). If something `isNotFound` we should investigate it and understand why, as that should never be the case in the happy-path.

### DIFF
--- a/packages/hop-node/src/watchers/SyncWatcher.ts
+++ b/packages/hop-node/src/watchers/SyncWatcher.ts
@@ -120,9 +120,9 @@ class SyncWatcher extends BaseWatcher {
         }
         await this.preSyncHandler()
         await this.syncHandler()
-        this.logger.debug('done syncing pure handlers. index:', this.syncIndex)
+        this.logger.debug(`done syncing pure handlers. index: ${this.syncIndex}`)
         await this.incompletePollSync()
-        this.logger.debug('done syncing incomplete items. index:', this.syncIndex)
+        this.logger.debug(`done syncing incomplete items. index: ${this.syncIndex}`)
         await this.postSyncHandler()
       } catch (err) {
         this.notifier.error(`pollSync error: ${err.message}`)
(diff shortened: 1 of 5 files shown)


packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -794,13 +688,15 @@ class SyncWatcher extends BaseWatcher {
...
+      logger.warn(`populateTransferSentTimestamp marking item not found: sourceChainId. dbItem: ${JSON.stringify(dbTransfer)}`)
+      await this.db.transfers.update(transferId, { isNotFound: true })
       return
     }
     const sourceBridge = this.getSiblingWatcherByChainId(sourceChainId).bridge // eslint-disable-line @typescript-eslint/no-non-null-assertion
     const timestamp = await sourceBridge.getBlockTimestamp(transferSentBlockNumber)
     if (!timestamp) {
-      logger.warn(`timestamp not found for block number ${transferSentBlockNumber} on sourceChainId ${sourceChainId}`)
+      logger.warn(`populateTransferSentTimestamp marking item not found: timestamp for block number ${transferSentBlockNumber} on sourceChainId ${sourceChainId}. dbItem: ${JSON.stringify(dbTransfer)}`)
+      await this.db.transfers.update(transferId, { isNotFound: true })
```
miguelmota, line 699: Do you think it would be possible for getBlockTimestamp here to return null if the rpc provider hasn't yet indexed the block yet? basically would it be a possibility to prematurely set this item as not found because it couldn't find the block info but it eventually does exist (I think we saw some endpoints in the past not return info right away)

miguelmota (approved):
@shanefontaine I think the changes look good. Left a small question

packages/hop-node/src/watchers/SyncWatcher.ts
shanefontaine: This was true when there was a `0` block delay for ORUs AND we were using the ORU RPC provider instead of an aggregated node service (Infura or Alchemy). There is now a delay and the node services we use aggregate data.

Because of both of those facts, I believe we will not ever prematurely set this as not found.

PR #303, opened 2021-12-01
sdk: fix gas estimation

kangarang:
previously, if `sourceChain` was optimism, gas estimations would fail. setting the `from` address resolved this. i'm guessing the optimism rpc requires a `from` field when calling `eth_estimateGas`.


miguelmota:
I think this PR can be closed in favor of https://github.com/hop-protocol/hop/commit/a564c534e7b6104515a421284c0409daa10345d3

PR #304, opened 2021-12-02
hop-node: Add config option for specifying routes bonder wants to bond

miguelmota:
Add config option for uni-direction routes

Example config for bonding withdrawals (and roots) going source→destination:

```json
  "routes": {
    "optimism": {
      "arbitrum": true,
      "ethereum": true
    },
    "arbitrum": {
      "optimism": true
    },
    "xdai": {
      "polygon": true
    }
  }
```

routes:
optimism→arbitrum
arbitrum→optimism
optimism→ethereum
xdai→polygon


packages/hop-node/src/watchers/BondTransferRootWatcher.ts
```diff
@@ -41,7 +43,24 @@ class BondTransferRootWatcher extends BaseWatcher {
   }
 
   async checkTransfersCommittedFromDb () {
-    const dbTransferRoots = await this.db.transferRoots.getUnbondedTransferRoots()
+    let filterSourceChainId: number | undefined
```
shanefontaine, line 46: Does it make sense to DRY these functions up and put them in BaseWatcher.ts? I think the destination ones make sense since they are identical, AFAICT. Something like:

`const filterSourceChainId = getFilterSourceChainId()`

&

`const filterDestinationChainIds = getFilterDestinationChainIds()`

shanefontaine (commented):
1. Sanity check -- if no routes are specified in the config, what would you expect the bonder to do (if anything)?
2. How are you thinking about the more general bonder actions? Specifically, who performs exits, who performs bonding of transfer roots, and who performs settlements? I think this code needs to be updated slightly based on that answer, but want to be on the same page before making any suggestions.

miguelmota:
1. It will work as it currently behaves (meaning it'll bond all the enabled chains)

2. Answers based on the discussion the other day and what Chris mentioned:

  > Who performs exits?

  We'll be performing all exits and then be compensated on a scheduled basis (every 2 weeks, or month, etc)  by requesting payment from bonders 

  > Who performs bonding of roots?

  The bonder associated with the route will be performing the bonding of that route's roots 

  > Who performs settlements?

  Each bonder is responsible for settling their own roots, since each root is tied to a route 
 
@shanefontaine

packages/hop-node/src/watchers/BondTransferRootWatcher.ts
miguelmota: Will do

shanefontaine:
Nice, exits make sense.

If I understand the code correct, I believe any bonder running a route on any asset would try to settle another bonder's root. I think this is a bug?

Regardless of the line above, I want to say the route filters should be more generalized and apply to all watchers. IIUC, they currently only exist on the `BondTransferRootWatcher`, `BondWithdrawalWatcher`, and `CommitTransfersWatcher`. I think it makes sense to have every watcher being run by a bonder filter for only the routes they are in charge of. This reduces mental overhead of trying to have to understand which watchers should have destination filters and why. Plus, it makes the route filtering standard throughout the codebase.

If I am thinking about it correctly (and I may be missing something), then I _think_ it should act in the same way that we filter everything by the `sourceChainId` in the getters:

```
      if (filter.sourceChainId) {
        if (filter.sourceChainId !== item.sourceChainId) {
          return false
        }
      }
```

Curious to hear your thoughts.

packages/hop-node/src/watchers/BondTransferRootWatcher.ts
miguelmota: Resolved with https://github.com/hop-protocol/hop/commit/581ed82661c82a7586482afc5f7a631dab7ed18f

miguelmota:
I believe the latest changes now in the PR should address the comments (settlement filter fix and running only specific watchers based on config routes) 

Open to feedback

@shanefontaine

shanefontaine:
After sleeping on it, I think the route should be defined in the watcher constructor (since it won't change) and every watcher should only ever get DB entries for their respective routes. In that case, it might be something as follows. In this case, every watcher is only watching for their route and is never concerned about anything else, which _I think_ should be the way the system works.

`BaseWatcher.ts`
```
 constructor() {
    this.route = this.getFilterRoute()
 }

....

  async getFilterRoute () {
    const sourceChainId = await this.getFilterDestinationChainIds()
    const destinationChainIds = await this.getFilterDestinationChainIds()
    return {
      sourceChainId,
      destinationChainIds
    }
  }
```

`xxxWatcher.ts` (this would be in every watcher)
```
  async checkTransfersCommittedFromDb () {
    const dbTransferRoots = await this.db.transferRoots.getUnbondedTransferRoots(
      this.route
    )

  ...
```

miguelmota:
Implementing the db class wrapper for cleaning the watcher class db getters (something I suggested while discussing this) is a little more complicated than I expected, so I dried up the code as you suggested which is good enough for now in (https://github.com/hop-protocol/hop/pull/304/commits/72892fc148fd0905cbc51e63c16c2476c7369902) and the watchers can be further dried in later PRs

Let me know if you have any other suggestions

@shanefontaine

packages/hop-node/src/watchers/CommitTransfersWatcher.ts
```diff
@@ -78,26 +79,23 @@ class CommitTransfersWatcher extends BaseWatcher {
...
     )
-    const destinationChainIds: number[] = []
+
+    const seen: Record<number, boolean> = {}
     for (const dbTransfer of dbTransfers) {
       const { destinationChainId } = dbTransfer
-      if (!destinationChainIds.includes(destinationChainId!)) { // eslint-disable-line @typescript-eslint/no-non-null-assertion
-        destinationChainIds.push(destinationChainId!) // eslint-disable-line @typescript-eslint/no-non-null-assertion
+      if (!seen[destinationChainId!]) { // eslint-disable-line @typescript-eslint/no-non-null-assertion
+        seen[destinationChainId!] = true // eslint-disable-line @typescript-eslint/no-non-null-assertion
```
shanefontaine, line 95: If `this.checkIfShouldCommit()` fails for some reason, then I think this logic prevents it from being retried again. What is the purpose of this `seen` logic?

packages/hop-node/src/watchers/SettleBondedWithdrawalWatcher.ts
```diff
@@ -116,6 +111,15 @@ class SettleBondedWithdrawalWatcher extends BaseWatcher {
       }
 
       for (const bonder of bonderSet.values()) {
+        // if bonder has custom routes configured, then it should only check it's own
+        // bonded withdrawals to see if they need settlement
+        if (this.hasCustomRoutes) {
```
shanefontaine, line 116: Given our new understanding, I believe all bonders should have custom routes, so the concept of `hasCustomRoutes` can be removed. If a bonder wants all routes enabled (the `availableLiquidity` worker, for example), then they can just explicitly state every route in the config.

packages/hop-node/src/watchers/classes/BaseWatcher.ts
```diff
@@ -236,6 +236,41 @@ class BaseWatcher extends EventEmitter implements IBaseWatcher {
     }
   }
 
+  get hasCustomRoutes () {
```
shanefontaine, line 239: See comment above about this being unnecessary, I believe.

shanefontaine:
Left a few more comments. A few more higher-level items below.

For this iteration, I think it still makes sense to filter _every_ getter by route so that (a) it matches the functionality of the next upgrade and (b) consistency / less mental overhead. This means adding `if (!this.isRouteOk(filter, item))` to all `getXxx()` functions, such as `getExitableTransferRoots()` and a I believe a few others. I think point (a) is the important one because it will make the next upgrade super seamless.

I also believe I mistyped earlier. I think the following should be taken out of `isRouteOk()` and left inside the single function it is used now (`getUnbondedTransferRoots()`). From a logical perspective, this does not necessarily have to do with the routes available to the watcher (like the others in this PR), but rather that one function and how it is called now.

```
      if (filter.destinationChainId) {
          if (filter.destinationChainId !== item.destinationChainId) {
              return false
          }
      }

```

shanefontaine (changes requested):

packages/hop-node/src/watchers/SettleBondedWithdrawalWatcher.ts
miguelmota: Removed in https://github.com/hop-protocol/hop/commit/4d5b5a96e8c1b000bda20ca710aaeaaab5ce568e

Going forward, should the bonder config be explicitly set for other things, like fee bps, enabled networks, rpc urls, etc? basically have everything set to off/0 values unless explicitly set

packages/hop-node/src/watchers/classes/BaseWatcher.ts
miguelmota: Resolved by https://github.com/hop-protocol/hop/commit/4d5b5a96e8c1b000bda20ca710aaeaaab5ce568e

packages/hop-node/src/watchers/CommitTransfersWatcher.ts
miguelmota: I was trying to clean it up by using a map and one loop but I see what you mean now. 

Changes reverted in https://github.com/hop-protocol/hop/commit/4d5b5a96e8c1b000bda20ca710aaeaaab5ce568e

miguelmota:
> Left a few more comments. A few more higher-level items below.
> 
> For this iteration, I think it still makes sense to filter _every_ getter by route so that (a) it matches the functionality of the next upgrade and (b) consistency / less mental overhead. This means adding `if (!this.isRouteOk(filter, item))` to all `getXxx()` functions, such as `getExitableTransferRoots()` and a I believe a few others. I think point (a) is the important one because it will make the next upgrade super seamless.
> 
> I also believe I mistyped earlier. I think the following should be taken out of `isRouteOk()` and left inside the single function it is used now (`getUnbondedTransferRoots()`). From a logical perspective, this does not necessarily have to do with the routes available to the watcher (like the others in this PR), but rather that one function and how it is called now.
> 
> ```
>       if (filter.destinationChainId) {
>           if (filter.destinationChainId !== item.destinationChainId) {
>               return false
>           }
>       }
> ```

I agree with this

Resolved with https://github.com/hop-protocol/hop/pull/304/commits/4d5b5a96e8c1b000bda20ca710aaeaaab5ce568e (also removed some unused db getters)

packages/hop-node/src/watchers/SettleBondedWithdrawalWatcher.ts
shanefontaine: > basically have everything set to off/0 values unless explicitly set

I think that makes a ton of sense now that we have routes. Only consideration is that we wouldn't want to accidentally default to `0` if not set and then accept, for example, fees of `0`. Maybe upon startup there is an "if route check that values are non-`0`" or something like that, but I think, in general, this makes a ton of sense.

shanefontaine (approved):
PR LGTM! Ready to merge 👍

PR #305, opened 2021-12-02
hop-node: add settlement cost to destination tx fee

shanefontaine:
The cost to settle a tx should be included in the destination tx fee, not the bonder fee. This is because the volume of the transfer has no effect on the cost of the settlement. This PR shifts the settlement cost away from the `bonderFee` and into the `destinationTxFee`.


packages/hop-node/src/constants/constants.ts
```diff
@@ -24,6 +24,14 @@ export const AvgBlockTimeSeconds = {
   xDai: 5
 }
 
+export const SettlementCostPerTx: Record<string, number> = {
```
miguelmota, line 27: > `SettlementCostPerTx`

If I understand correctly, the values here are gas limits, so I think a variable name like `SettlementGasLimitPerTx` would be better since Cost implies gasLimit*gasPrice

miguelmota (approved):
Left minor comment

LGTM 👍

packages/hop-node/src/constants/constants.ts
shanefontaine: You are correct. Good call, updated with 9991eb7f7ee2efce6b373f6ded989f8779617319

PR #306, opened 2021-12-03
core: Update bonder address structure

miguelmota:


packages/core/src/addresses/mainnet.ts
```diff
@@ -2,12 +2,198 @@ import { Addresses } from './types'
...
+      polygon: {
+        ethereum: '0xa6a688F107851131F0E1dce493EbBebFAf99203e',
+        arbitrum: '0xa6a688F107851131F0E1dce493EbBebFAf99203e',
+        xdai: '0xa6a688F107851131F0E1dce493EbBebFAf99203e',
+        optimism: '0xa6a688F107851131F0E1dce493EbBebFAf99203e'
+      }
+    },
+    MATIC: {
+      ethereum: {
+        optimism: '0xd8781ca9163e9f132a4d8392332e64115688013a',
```
shanefontaine, line 71: There are no Optimism and Arbitrum routes for MATIC. It might make sense to remove them. If it doesn't make sense to remove them (for example, if the typings get annoying b/c of it), maybe it makes sense to use the `ZERO_ADDRESS` here.

packages/core/src/addresses/types.ts
```diff
@@ -62,7 +62,7 @@ export type Bridges = {
 }
 
 export type Bonders = {
-  [token: string]: string[]
+  [token: string]: Record<string, Record<string, string>>
```
shanefontaine, line 65: Up for discussion, as I don't have too strong of an opinion here.

It might make sense to explicitly lay out the structure here. Things look good now, but I could see an annoying bug in the future where copy/paste goes wrong and a `source` doesn't have the correct `destination`s. For example, we could accidentally do:

```
polygon: {
  ethereum: ...
  optimism: ...
  arbitrum: ...
  polygon: ... // <------------ supposed to be xDai
}
```

So maybe it is something like

```
type PolygonRoute = {
  ethereum: ...
  optimism: ...
  arbitrum: ...
  xdai: ...
}

...

type Routes = {
  PolygonRoutes,
  OptimismRoutes,
  ...
}


export type Bonders = {
  [token: string]: Routes
```

packages/hop-node/src/watchers/classes/Bridge.ts
```diff
@@ -829,6 +829,7 @@ export default class Bridge extends ContractBase {
   }
 
   getConfigBonderAddress (): string {
-    return globalConfig?.bonders?.[this.tokenSymbol]?.[0]
+    const sourceBonders = globalConfig?.bonders?.[this.tokenSymbol]?.[this.chainSlug]
+    return sourceBonders?.[Object.keys(sourceBonders)[0]]
```
shanefontaine, line 833: This is used to estimate gas for transactions, so we need to ensure that the correct bonder is used for the estimations. So I believe this needs to return:

`return sourceBonders?.[this.destination]` or something like that.

With dedicated routes, we cannot default to `[0]` when

packages/sdk/src/Base.ts
```diff
@@ -429,12 +429,18 @@ class Base {
...
   }
 
-  public getBonderAddress (token: TToken): string {
+  protected _getBonderAddress (token: TToken, sourceChain: TChain, destinationChain?: TChain): string {
     token = this.toTokenModel(token)
-    return this.bonders?.[this.network]?.[token.canonicalSymbol]?.[0]
+    sourceChain = this.toChainModel(sourceChain)
+    destinationChain = this.toChainModel(destinationChain)
+    const sourceBonders = this.bonders?.[this.network]?.[token.canonicalSymbol]?.[sourceChain.slug]
+    if (destinationChain) {
```
shanefontaine, line 437: With the introduction of bonder routes, I _think_ we now need to always have the context of source and destination for stuff like this. Thinking out loud, this is probably the biggest mental model update.

packages/sdk/src/Base.ts
```diff
@@ -429,12 +429,18 @@ class Base {
...
+      return sourceBonders?.[destinationChain.slug]
+    }
+    return sourceBonders?.[Object.keys(sourceBonders)[0]]
```
shanefontaine, line 440: See comment above.

shanefontaine (changes requested):
@miguelmota Back to you.

With the introduction of bonder routes, I _think_ we now need to always have the context of source and destination when getting a bonder. Thinking out loud, this is probably the biggest mental model update.

In practice, all the `getBonderAddress()` calls in this PR need to have the context of the source and the destination chain, or else I think this doesn't work for multiple bonders. For example, if each route is its own bonder then `getAvailableLiquidity()` will need to know the exact route.

I'm not sure what else might need to change because of this kind of thinking. I can't come up with anything off the top of my head.

packages/core/src/addresses/mainnet.ts
miguelmota: Resolved by https://github.com/hop-protocol/hop/commit/31b6b58f9d3edd92481619bfbd066287e0b74d94

packages/core/src/addresses/types.ts
miguelmota: Resolved by https://github.com/hop-protocol/hop/commit/31b6b58f9d3edd92481619bfbd066287e0b74d94

packages/hop-node/src/watchers/classes/Bridge.ts
miguelmota: Resolved by https://github.com/hop-protocol/hop/commit/98c910a15e4b6e3537df362b44283f62f2164323

packages/sdk/src/Base.ts
miguelmota: Resolved by https://github.com/hop-protocol/hop/commit/98c910a15e4b6e3537df362b44283f62f2164323

packages/sdk/src/Base.ts
miguelmota: Resolved by https://github.com/hop-protocol/hop/commit/98c910a15e4b6e3537df362b44283f62f2164323

shanefontaine:
👍  LGTM

packages/sdk/src/Base.ts
```diff
@@ -437,12 +437,14 @@ class Base {
...
+  protected _getBonderAddress (token: TToken, sourceChain: TChain, destinationChain: TChain): string {
     token = this.toTokenModel(token)
-    return this.bonders?.[this.network]?.[token.canonicalSymbol]?.[0]
+    sourceChain = this.toChainModel(sourceChain)
+    destinationChain = this.toChainModel(destinationChain)
+    return this.bonders?.[this.network]?.[token.canonicalSymbol]?.[sourceChain.slug]?.[destinationChain.slug]
   }
 
-  public getBonderAddresses (token: TToken): string[] {
+  public getBonderAddresses (token: TToken): Routes {
```
shanefontaine, line 447: I loosely believe this function should be removed, as it is not used anywhere and it doesn't necessarily make sense in the context of per-route bonder, I believe. Happy to discuss otherwise.

packages/frontend/src/pages/Stats/StatsContext.tsx
```diff
@@ -168,10 +169,10 @@ const StatsContextProvider: FC = ({ children }) => {
       return
     }
     const [credit, debit, totalDebit, availableLiquidity, eth] = await Promise.all([
-      bridge.getCredit(selectedNetwork.slug, bonder),
-      bridge.getDebit(selectedNetwork.slug, bonder),
-      bridge.getTotalDebit(selectedNetwork.slug, bonder),
-      bridge.getAvailableLiquidity(selectedNetwork.slug, bonder),
+      bridge.getCredit(selectedNetwork.slug, Chain.Ethereum, bonder),
```
shanefontaine, line 172: See top level comment about these functions

packages/sdk/src/HopBridge.ts
```diff
@@ -901,12 +903,13 @@ class HopBridge extends Base {
   }
 
   public async getAvailableLiquidity (
```
shanefontaine, line 905: See top level comment.

shanefontaine (changes requested):
I believe there are other places that need to be addressed in order for this to work as intended. The ones I came up with are as follows:

* [FE stats context 1](https://github.com/hop-protocol/hop/blob/0bad45032eec85811d7f91f432bda12cb1c149a9/packages/frontend/src/pages/Stats/StatsContext.tsx#L215)
* [FE stats context 2](https://github.com/hop-protocol/hop/blob/0bad45032eec85811d7f91f432bda12cb1c149a9/packages/frontend/src/pages/Stats/StatsContext.tsx#L402)
* [SDK base](https://github.com/hop-protocol/hop/blob/0bad45032eec85811d7f91f432bda12cb1c149a9/packages/sdk/src/Base.ts#L116) 

Please let me know if you can think of any others, as I think there may be even more given how broad this change is. I will think through once more as well, but things are looking good!

---

The following functions in the SDK are unintuitive now. If both a `bonder` and `destination` chain is passed in, then destination is completely ignored, even if the caller expects to get the value at the specific destination.

* `getCredit()`
* `getDebit()`
* `getTotalDebit()`
* `getAvailableLiquidity()`

I think it makes sense to update the function signature of those functions to `getXxx(chain: TChain, bonder: string)`. This would do a few things:
1. The bonder would be explicitly passed in, as to avoid any implicitness and confusion
2. The destination chain would not be used during the call, which I believe is expected since those functions should only care about a single chain and be unconcerned with source/destination.

packages/sdk/src/HopBridge.ts
miguelmota: Resolved with https://github.com/hop-protocol/hop/commit/460cd62d1f1399a6d44661a8b355b6c688114bb2

packages/frontend/src/pages/Stats/StatsContext.tsx
miguelmota: Resolved with https://github.com/hop-protocol/hop/commit/460cd62d1f1399a6d44661a8b355b6c688114bb2

packages/sdk/src/Base.ts
miguelmota: Agree

Resolved with https://github.com/hop-protocol/hop/commit/460cd62d1f1399a6d44661a8b355b6c688114bb2

miguelmota:
> I believe there are other places that need to be addressed in order for this to work as intended. The ones I came up with are as follows:
> 
>     * [FE stats context 1](https://github.com/hop-protocol/hop/blob/0bad45032eec85811d7f91f432bda12cb1c149a9/packages/frontend/src/pages/Stats/StatsContext.tsx#L215)
> 
>     * [FE stats context 2](https://github.com/hop-protocol/hop/blob/0bad45032eec85811d7f91f432bda12cb1c149a9/packages/frontend/src/pages/Stats/StatsContext.tsx#L402)
> 
>     * [SDK base](https://github.com/hop-protocol/hop/blob/0bad45032eec85811d7f91f432bda12cb1c149a9/packages/sdk/src/Base.ts#L116)
> 
> 
> Please let me know if you can think of any others, as I think there may be even more given how broad this change is. I will think through once more as well, but things are looking good!

This is a great find. I'll get the frontend updated

> The following functions in the SDK are unintuitive now. If both a `bonder` and `destination` chain is passed in, then destination is completely ignored, even if the caller expects to get the value at the specific destination.
> 
>     * `getCredit()`
> 
>     * `getDebit()`
> 
>     * `getTotalDebit()`
> 
>     * `getAvailableLiquidity()`
> 
> 
> I think it makes sense to update the function signature of those functions to `getXxx(chain: TChain, bonder: string)`. This would do a few things:
> 
>     1. The bonder would be explicitly passed in, as to avoid any implicitness and confusion
> 
>     2. The destination chain would not be used during the call, which I believe is expected since those functions should only care about a single chain and be unconcerned with source/destination.

I agree and meant to bring it up friday. Resolved with https://github.com/hop-protocol/hop/commit/460cd62d1f1399a6d44661a8b355b6c688114bb2

shanefontaine:
I think [`getFrontendAvailableLiquidity()`](https://github.com/hop-protocol/hop/blob/feature/core-pkg-bonders/packages/sdk/src/HopBridge.ts#L924) should be updated as well, since it is subject to the same issues WRT mismatched destination/bonder.

I think I was wrong about changing [setConfigAddresses()](https://github.com/hop-protocol/hop/blob/0bad45032eec85811d7f91f432bda12cb1c149a9/packages/sdk/src/Base.ts#L116) in my last comment. You did not update it, which I believe is correct. Just to be explicit, you believe that this line is good as-is, correct?


Other than that, changes look great to me. Ready for merge IMO after addressing those final comments 🎉

@miguelmota

shanefontaine (approved):
👍

miguelmota:
There's 2 reasons I think `getFrontendAvailableLiquidity` should be left as is

1. This method requires both source and destination chain already so it can read the bonder for the route easily already
1. It won't require a manual look up for bonders in our fronend and also third parties that use our sdk (like polymarket which is using this method)

I didn't update `setConfigAddresses` because I think it's good the way it is (it takes in the whole object)

shanefontaine:
I see. Do you think it makes sense to take `bonder` out of the function signature and define it within the function itself?

@miguelmota

miguelmota:
@shanefontaine yeah I think so because it's not being manually passed in anywhere

shanefontaine:
Perfect 👍

PR #307, opened 2021-12-03
Feature/gas price and limit updates

shanefontaine:
Not all Polygon nodes enforce the 30 Gwei minimum, as [suggested by the Polygon team](https://forum.matic.network/t/recommended-min-gas-price-setting/2531). Because of this, `eth_gasPrice` calls sometimes return a low `gasPrice`, depending on the validator your request hits. This PR enforces a 30 Gwei min for Polygon txs.

Considerations about this PR: this cannot get in the way of any boosting of txs or else it would break. For example, we cannot do `max(gasPrice(), 30)` to boost a tx because the `30` might not be `>10%` of the previously submitted tx. I believe we are covered by this change, but would appreciate an extra pair of eyes on that fact.

From Alchemy:

> The polygon team actually mentioned that although they recommended validators set their minimum to 30 gwei, not all validators have done it yet so it is technically possible that some validators will accept lower prices which is why the nodes behave this way. Since its not yet something they will change at the node level, they told us it’s up to the user to handle the gas price as needed


miguelmota:
Changes look perfect. Don't see any issues

@shanefontaine

PR #308, opened 2021-12-04
UI: reskin

kangarang:
this pr updates component styles and adds darkmode to the GUI.

See [notion document](https://www.notion.so/authereum/Frontend-Refactor-Dev-Experience-a53c1933909544b5b1faee6507a3899a) for more details.


miguelmota (approved):

miguelmota (changes requested):
Looks really good @kangarang!

I noticed some minor cosmetic issues in a few places

The info button is hard to see in dark mode on convert page

DeepinScreenshot_select-area_20211206092040

The account modal buttons are squared but should be round or not have a hard edge 

DeepinScreenshot_select-area_20211206092007

The disabled state button has a box shadow but the active button does not. I feel like the disabled button should not appear to be clickable so should not have a box shadow

DeepinScreenshot_select-area_20211206091905

The arrow on the convert page has a hard outline while in the send page does not
DeepinScreenshot_select-area_20211206091945

I think there could be better contrast between the black text and dark purple background

DeepinScreenshot_select-area_20211206091809

PR #309, opened 2021-12-04
Feature/correct recipient

shanefontaine:
Update legacy code to explicitly pass in recipients to various function calls. Legacy code defaulted to bonder, which will not always be correct given our updated codebase. 

Other small changes as well.


miguelmota (approved):
Changes look good 👍

PR #313, opened 2021-12-08
hop-node: Remove dead code

miguelmota:


shanefontaine (commented):
I love this PR.

* `polyonBridge` and `xdaiBridge` should be removed from `cli.ts`
* Do you think it also makes sense to remove `loadTest.ts`? I think it is a useful tool but I almost think it was written so long ago when we had a different understanding of the system that it will be hard to use & iterate on if it does still work. Loose opinion here so LMK what you're feeling.

miguelmota:
1. Thanks for catching that. Removed https://github.com/hop-protocol/hop/pull/313/commits/8da37ccb7eb4fde61c0f9b7f9a902ae6b1846596
2. Agree, removed https://github.com/hop-protocol/hop/pull/313/commits/8da37ccb7eb4fde61c0f9b7f9a902ae6b1846596
3. The test suite needs an overhaul and the helpers file has some useful methods so I'm not going to touch the test files for now. The test suite can be cleaned up in a later PR

PR #315, opened 2021-12-09
Feature/rpc optimizations

shanefontaine:
* Removes concept of `s3AvailableCredit`
* Remove some “nonce out of sync” logic. We only see this when things are seriously wrong and we will have already noticed this by the fact that transactions are blocked. This was using a surprising number of `getTransactionCount` calls, which are expensive on Alchemy
* Does not sync unbonded TransferRoot amounts when the source is a non-ORU (since they will not have unbonded transfer roots)
* Only sync a local bonder's `availableCredit` every few polls. The only place this is used is in edge-cases in the `BondeWithdrawalWatcher` ([1](https://github.com/hop-protocol/hop/blob/484a47d5ceae875774415e88477b739cfb3bc0c8/packages/hop-node/src/watchers/BondWithdrawalWatcher.ts#L73) [2](https://github.com/hop-protocol/hop/blob/484a47d5ceae875774415e88477b739cfb3bc0c8/packages/hop-node/src/watchers/BondWithdrawalWatcher.ts#L131)). Reduces number of these types of calls by 90%.


packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -51,13 +51,13 @@ class SyncWatcher extends BaseWatcher {
   syncFromDate: string
   customStartBlockNumber: number
   ready: boolean = false
-  private s3AvailableCredit: { [destinationChain: string]: BigNumber } = {} // bonder from core package config
   private availableCredit: { [destinationChain: string]: BigNumber } = {} // own bonder
```
miguelmota, line 54: > // own bonder

we should remove this comment now

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -357,7 +357,22 @@ class SyncWatcher extends BaseWatcher {
...
+  }
+
+  async syncBonderCredit () {
+    this.bonderCreditPollerIncrementer++
+    const bonderCreditSyncInterval = 10
+    // Don't check the 0 remainder so that the bonder has a valid credit immediately on startup
+    const shouldSync = this.bonderCreditPollerIncrementer % bonderCreditSyncInterval === 1
+
+    // When not uploading to S3, only sync on certain poll intervals
+    if (!this.s3Upload && !shouldSync) {
```
miguelmota, line 370: since most bonders will have `s3Upload` be `false` then will it ever get past this `return`? should be an `||` instead of `&&`?

miguelmota (changes requested):

packages/hop-node/src/watchers/SyncWatcher.ts
shanefontaine: Resolved with c4586e6

packages/hop-node/src/watchers/SyncWatcher.ts
shanefontaine: I believe it will get past the `return` on every 10th poll. I believe it should be `&&`.

The goal was to only sync every 10 polls. For a typical bonder, I think 9 out of 10 polls will be `(true && true)`, but I think every 10th one will be `(true && false)`, which would then sync. Do you believe that to be correct?

miguelmota: sorry you are correct, I had it inverted in my head. PR looks good to merge

miguelmota (approved):

PR #316, opened 2021-12-10
UI: fix timing bugs

kangarang:
this pr fixes 2 bugs: (1) the flashing balance bug and (2) the "insufficient liquidity" bug

the former bug was due to a conditional block of code which compared `token._symbol === token.symbol`. so if ethereum mainnet was selected as a network for sending tokens, the `useBalance` hook would be re-triggered every specified interval (8 seconds).

the latter bug was due to inconsistent polling for 2 values (`availableLiquidity` and `requiredLiquidity`) which were subsequently compared to determine sufficient liquidity. the bug would appear if the token (i.e. `selectedBridge`) was changed within the timing difference, resulting in liquidity values for 2 different tokens.

(2) is addressed **not** by changing the interval timings, but by calling the `useAvailableLiquidity` inside the `useSendData` hook, where the `selectedBridge` will always be the correct bridge for the token in that hook

- call `useAvailableLiquidity` hook -> `useSendData`
- replace custom `useInterval` hook with `useInterval` hook from `react-use` (our custom hook was intermittently logging console warnings. this seems to be fixed now)


packages/frontend/src/pages/Send/useSendData.tsx
```diff
@@ -1,18 +1,20 @@
...
+import useAvailableLiquidity from './useAvailableLiquidity'
 
 const useSendData = (
   token: Token | undefined,
   slippageTolerance: number,
   fromNetwork: Network | undefined,
   toNetwork: Network | undefined,
-  fromAmount: BigNumber | undefined
+  fromAmount: BigNumber | undefined,
+  selectedBridge?: HopBridge
```
kangarang, line 15: this can be removed, as we can call `sdk.bridge(token.symbol)` for better readability and consistency and less overhead

miguelmota (approved):
LGTM, works as expected 👍

PR #318, opened 2021-12-11
hop-node: TransferRootsDb: Index by transfer root id instead of by transfer root hash

miguelmota:


shanefontaine (commented):
From a top-level perspective, is the goal to make a subDB for `transferRootIds` or is it to replace the current db with one that is indexed by `transferRootIds`?

My original thought was to make the top-level DB indexed by `transferRootId`s and every update and lookup would be based on the `transferRootId` and not the `transferRootHash`. But I now am trying to think through the advantages (if any) of doing all the work for that vs. creating a separate subDb. Curious if that is how you envisioned it.

miguelmota:
I think the subdb would achieve the goal of doing a more efficient look up by transfer root id without introducing additional complexity (ie computing transferRootId on the handlers). The changes for migrating to using transferRootId as the main index for everything wouldn't be hard but I'm not sure it's worth the trade off at the moment, unless it provides additional safety guarantees to existing watchers (eg prevents bonder from acting on invalid data and losing funds) then I think it'd be worth making the changes

miguelmota:
The current PR switches over to using the transfer root ID is the primary index

The things that need to be finished before merging:

- finish `migration` function for `TransferRootsDb`
- figure out best way to get transferRootId in `handleMultipleWithdrawalsSettledEvent` handler
- verify challenge watcher has no regression bugs

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -75,57 +73,124 @@ const invalidTransferRoots: Record<string, boolean> = {
...
+    // re-index timestamped db keys by transfer root id
+    let items = await this.subDbTimestamps.getKeyValues()
+    let promises: Array<Promise<any>> = []
+    for (const { key, value } of items) {
+      promises.push(new Promise(async (resolve) => {
+        if (!value?.transferRootHash) {
+          resolve(null)
+          return
+        }
+        const item = await this.getById(value.transferRootHash)
```
shanefontaine, line 108: Does this (and some of the same types of calls on subsequent lines) need to `getById` from the specific DB? Such as `this.subDbTimetamps.getById()`?

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -75,57 +73,124 @@ const invalidTransferRoots: Record<string, boolean> = {
...
-
-  async trackTimestampedKeyByTransferRootHash (transferRootHash: string) {
-    const transferRoot = await this.getByTransferRootHash(transferRootHash)
-    return await this.trackTimestampedKey(transferRoot)
-  }
-
   getTimestampedKey (transferRoot: Partial<TransferRoot>) {
-    if (transferRoot.committedAt && transferRoot.transferRootHash) {
-      const key = `transferRoot:${transferRoot.committedAt}:${transferRoot.transferRootHash}`
+    if (transferRoot.committedAt && transferRoot.transferRootId) {
```
shanefontaine, line 192: (I think this should be addressed in another PR)

It is not guaranteed that a transferRoot has a `committedAt` entry at this point in its lifecycle. I think there is an edge case here where a bonder calls `bondTransferRoot` (maliciously, for example) on a non-existent `transferId`. In this case `transferRoot.committedAt` will be undefined. I believe we _should_ handle this somehow, but I'm not sure how and I think it might be worth taking care of after this PR.

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -135,38 +200,69 @@ class TransferRootsDb extends TimestampedKeysDb<TransferRoot> {
...
+  }
+
+  async update (transferRootId: string, transferRoot: Partial<TransferRoot>) {
+    await this.tilReady()
+    const logger = this.logger.create({ root: transferRootId })
     logger.debug('update called')
-    transferRoot.transferRootHash = transferRootHash
+    transferRoot.transferRootId = transferRootId
     const timestampedKv = await this.getTimestampedKeyValueForUpdate(transferRoot)
     const promises: Array<Promise<any>> = []
```
shanefontaine, line 250: probably my biggest nitpick ever :) -- Swap the order of the lines `const timestampedKv = ...` and `const promises: ...` since now the promises are used in more than just that `if (timestampedKv)` line.

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -180,58 +276,57 @@ class TransferRootsDb extends TimestampedKeysDb<TransferRoot> {
...
   filterTimestampedKeyValues = (x: any) => {
-    return x?.value?.transferRootHash
+    return x?.value?.transferRootId
   }
 
-  async getTransferRootHashes (dateFilter?: TransferRootsDateFilter): Promise<string[]> {
+  async getTransferRootIds (dateFilter?: TransferRootsDateFilter): Promise<string[]> {
+    await this.tilReady()
     // return only transfer-root keys that are within specified range (filter by timestamped keys)
+    const filter: KeyFilter = {
```
shanefontaine, line 306: Is there a reason why these filters are defined here now instead of only within the `if` statements if the `dateFilter` exists? If there is a good reason, why did this not exist before or why does it not exist in the Transfers DB?

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -508,29 +505,28 @@ class SyncWatcher extends BaseWatcher {
...
       logger.debug(`bondAmount: ${this.bridge.formatUnits(amount)}`)
-      logger.debug(`bondTransferRootId: ${bondTransferRootId}`)
+      logger.debug(`transferRootId: ${transferRootId}`)
 
-      await this.db.transferRoots.update(root, {
-        transferRootHash: root,
+      await this.db.transferRoots.update(transferRootId, {
+        transferRootHash,
         bonded: true,
         bondTotalAmount: amount,
```
shanefontaine, line 527: I believe it now makes sense to completely get rid of the concept of `bondTotalAmount`, since only one `totalAmount` can exist per `transferId` now. Based on what I'm seeing (and after one comment below is addressed), I think it should be super easy to completely get rid of the concept of `bondTotalAmount`.

**This DB entry does need to now add `totalAmount: amount`**, however, because we need that available everywhere. See one of the bullets on the top level comment on the PR for my thoughts there.

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -390,11 +485,12 @@ class TransferRootsDb extends TimestampedKeysDb<TransferRoot> {
       }
 
       return (
-        item.bondTransferRootId &&
+        item.transferRootId &&
         item.transferRootHash &&
+        !item.committed &&
+        item.totalAmount &&
         item.bonded &&
         item.destinationChainId &&
```
shanefontaine, line 493: `item.destinationChainId` needs to be removed from this db getter since it won't always be available.

(Sorry, I misspoke earlier and should have called this out when I OKed 12f3b0d61a843e8e8eb07335a13cde0f826929ca)

packages/hop-node/src/watchers/SettleBondedWithdrawalWatcher.ts
```diff
@@ -131,7 +131,7 @@ class SettleBondedWithdrawalWatcher extends BaseWatcher {
 
         // check settle-able transfer root
         promises.push(
-          this.checkTransferRootHash(transferRootHash!, bonder) // eslint-disable-line @typescript-eslint/no-non-null-assertion
+          this.checkTransferRootId(transferRootId!, bonder) // eslint-disable-line @typescript-eslint/no-non-null-assertion
             .catch((err: Error) => {
               this.logger.error('checkTransferRootHash error:', err.message)
```
shanefontaine, line 136: This should say `checkTransferRootId error`

shanefontaine (changes requested):
* I believe `transferRootId` should be passed in [here](https://github.com/hop-protocol/hop/blob/feature/index-transfer-root-id/packages/hop-node/src/watchers/ChallengeWatcher.ts#L70-L75) and the associated code in the above `for` loop can be removed. The logger should also be updated to use the `transferRootId` instead of the hash.
* (Somewhat related to this PR and to a comment on another line below) I believe [these](https://github.com/hop-protocol/hop/blob/601583548ac557e9fff701534640de48e5b61222/packages/hop-node/src/watchers/SyncWatcher.ts#L392-L397) `throw`s should be deleted since, (1) that data should always exist at that point in the code and (2) these handlers are now `pure` and I think throwing makes them un`pure`
* If it hasn’t already been done, it might make sense to run this on a copied DB and manually observe some of the DB changes to make sure it does what is expected. There are a lot of DB migration that are hard to visualize.
* As a mental model, I think it makes sense to think of a `transferId` being “created” by _either_ a `TransfersCommitted` event _or_  a `TransferRootBonded` event now (as opposed to just `TransfersCommitted`). The code handles this, so no change is needed, but it is probably worth thinking of it this way going forward so that we don’t make assumptions of what data might exist in the DB at any given time.

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -359,6 +458,7 @@ class TransferRootsDb extends TimestampedKeysDb<TransferRoot> {
   async getChallengeableTransferRoots (
     filter: GetItemsFilter = {}
   ): Promise<TransferRoot[]> {
+    await this.tilReady()
     const transferRoots: TransferRoot[] = await this.getTransferRootsFromTwoWeeks()
     return transferRoots.filter(item => {
       // Do not check if a rootHash has been committed. A rootHash can be committed and bonded,
```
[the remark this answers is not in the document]
miguelmota: Thanks, removed in https://github.com/hop-protocol/hop/pull/318/commits/48bcdaf9078daa59da9e74ca6af291aec5d1489c

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -153,14 +153,9 @@ class SyncWatcher extends BaseWatcher {
         const allChunks = chunk(incompleteTransferRoots, chunkSize)
         for (const chunks of allChunks) {
           await Promise.all(chunks.map(async (transferRoot: TransferRoot) => {
-            const { transferRootHash } = transferRoot
-            this.logger.info(`populating transferRoot: ${transferRootHash}`)
-            return this.populateTransferRootDbItem(transferRootHash!) // eslint-disable-line @typescript-eslint/no-non-null-assertion
-              .then(async () => {
-                this.logger.info(`populated transferRoot: ${transferRootHash}`)
-                // fill in missing db timestamped keys
-                return this.db.transferRoots.trackTimestampedKeyByTransferRootHash(transferRootHash!) // eslint-disable-line @typescript-eslint/no-non-null-assertion
```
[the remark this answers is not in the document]
miguelmota: Forgot to remove at line. Removed now in https://github.com/hop-protocol/hop/pull/318/commits/48bcdaf9078daa59da9e74ca6af291aec5d1489c

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -487,14 +482,16 @@ class SyncWatcher extends BaseWatcher {
 
   async handleTransferRootConfirmedEvent (event: TransferRootConfirmedEvent) {
     const {
-      rootHash: transferRootHash
+      rootHash: transferRootHash,
+      totalAmount
     } = event.args
-    const logger = this.logger.create({ root: transferRootHash })
+    const transferRootId = this.bridge.getTransferRootId(transferRootHash, totalAmount)
```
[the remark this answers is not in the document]
miguelmota: Updated in https://github.com/hop-protocol/hop/pull/318/commits/48bcdaf9078daa59da9e74ca6af291aec5d1489c

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -508,29 +505,28 @@ class SyncWatcher extends BaseWatcher {
...
+      logger.debug(`transferRootHash from event: ${transferRootHash}`)
       logger.debug(`event transactionHash: ${transactionHash}`)
       logger.debug(`event blockNumber: ${blockNumber}`)
       logger.debug(`bondAmount: ${this.bridge.formatUnits(amount)}`)
-      logger.debug(`bondTransferRootId: ${bondTransferRootId}`)
+      logger.debug(`transferRootId: ${transferRootId}`)
 
-      await this.db.transferRoots.update(root, {
-        transferRootHash: root,
+      await this.db.transferRoots.update(transferRootId, {
```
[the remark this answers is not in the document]
miguelmota: The key name is [added as property](https://github.com/hop-protocol/hop/blob/12f3b0d61a843e8e8eb07335a13cde0f826929ca/packages/hop-node/src/db/TransferRootsDb.ts#L248) on `.update()` so in this case it will also write the `transferRootId` to the entry

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -663,7 +667,8 @@ class SyncWatcher extends BaseWatcher {
       })
     const allSettled = rootAmountAllSettled || allBondableTransfersSettled
     logger.debug(`all settled: ${allSettled}`)
-    await this.db.transferRoots.update(transferRootHash, {
+    await this.db.transferRoots.update(transferRootId, {
+      transferRootId,
```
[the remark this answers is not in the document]
miguelmota: It is redundant and not needed there (for reasons explained in [above comment](https://github.com/hop-protocol/hop/pull/318#discussion_r769214135)). Removed in https://github.com/hop-protocol/hop/pull/318/commits/48bcdaf9078daa59da9e74ca6af291aec5d1489c

packages/hop-node/src/db/TransferRootsDb.ts
miguelmota: No necessarily; `this.getById` refers to the main db with all the entries and we need to query the entry by transferRootHash (current key) to find the transferRootId in order to use that as the new key.

packages/hop-node/src/watchers/SettleBondedWithdrawalWatcher.ts
miguelmota: Updated in https://github.com/hop-protocol/hop/pull/318/commits/48bcdaf9078daa59da9e74ca6af291aec5d1489c

packages/hop-node/src/db/TransferRootsDb.ts
miguelmota: Removed in https://github.com/hop-protocol/hop/pull/318/commits/48bcdaf9078daa59da9e74ca6af291aec5d1489c

packages/hop-node/src/db/TransferRootsDb.ts
miguelmota: Updated in https://github.com/hop-protocol/hop/pull/318/commits/48bcdaf9078daa59da9e74ca6af291aec5d1489c

packages/hop-node/src/watchers/SyncWatcher.ts
miguelmota: I believe you are correct. The concept of `bondTotalAmount` is removed in https://github.com/hop-protocol/hop/pull/318/commits/43a9da3c4f59b6788f81be76eb08d43b82098aca as well as using `transferRootId` only in challenge watcher

packages/hop-node/src/db/TransferRootsDb.ts
miguelmota: That's a great point. Maybe we track transferRoots with no commitedAt in separate subdb and then delete it from this subdb when the entry does have a commitedAt?

packages/hop-node/src/db/TransferRootsDb.ts
miguelmota: The reason is to only use the timestamped keys db for returning the ids, with or without filter timestamps, instead of using the base db or timestamped keys db depending whether there's a filter or not. It simplifies the function. The same logic has been applied to the TransfersDb now in https://github.com/hop-protocol/hop/pull/318/commits/ee413eece6af209e454ded15ba8fa64822376447

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -1035,25 +1035,29 @@ class SyncWatcher extends BaseWatcher {
       rootHash: transferRootHash,
       totalBondsSettled
     } = event.args
-    const logger = this.logger.create({ root: transferRootHash })
+    const dbTransferRoot = await this.db.transferRoots.getByTransferRootHash(transferRootHash)
+    if (!dbTransferRoot?.transferRootId) {
+      throw new Error(`expected db item for transfer root hash "${transferRootHash}"`)
```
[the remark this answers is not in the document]
miguelmota: I think you're right. This commit https://github.com/hop-protocol/hop/pull/318/commits/2dc16b0aab176160462830eb90f4ca1c1a45f0de should serialize the `handleMultipleWithdrawalsSettledEvent` promises to be processed after the `handleTransfersCommitted` promises, and the `handleTransfersCommitted` promises are processed after the `transferSpentPromises`. Let me know what you think about the changes

packages/hop-node/src/watchers/SyncWatcher.ts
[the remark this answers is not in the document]
shanefontaine: Nice. I believe `trackTimestampedKeyByTransferId()`and subsequently `trackTimestampedKey()` should be removed from `TransfersDb.ts` now as well.

packages/hop-node/src/db/TransferRootsDb.ts
shanefontaine: Not a bad idea. Let's discuss on standup.

packages/hop-node/src/watchers/SyncWatcher.ts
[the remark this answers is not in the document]
shanefontaine: I believe the changes will only run `handleMultipleWithdrawalsSettledEvent()` when the chain is not an L1. I believe this is incorrect, since we need it to run on all chains.

Since `handleTransfersCommittedEvent()` does not rely on the completion of the `transferSpentPromises()`, what do you think about putting it back where it was and then adjusting the way we work with promises. Maybe we can define another promise, `transfersCommittedPromsies` and do the following:

```
transfersCommittedPromsies.push(
        l2Bridge.mapTransfersCommittedEvents(
          async (event: TransfersCommittedEvent) => {
            return await Promise.all([
              this.handleTransfersCommittedEvent(event)
            ])
          },
          getOptions(l2Bridge.TransfersCommitted)
        )
      )
```

Then, later in the code, we could do:

```
    promises.push(
      Promise.all(transferSpentPromises)
        .then(async () => Promise.all(transfersCommittedPromsies))
        .then(async () => {
          return await this.bridge.mapMultipleWithdrawalsSettledEvents(
            async (event: MultipleWithdrawalsSettledEvent) => {
              return await this.handleMultipleWithdrawalsSettledEvent(event)
            },
            getOptions(this.bridge.MultipleWithdrawalsSettled)
          )
        })
    )
````

Does that make sense to you? Open to suggestions here.

File: packages/hop-node/src/watchers/SyncWatcher.ts (hunk 5 above)
Comment (shanefontaine) [reply packages/hop-node/src/watchers/SyncWatcher.ts]: This change should also add the following to the DB during this `update`, since it is possible that this is the first instance of a `transferRoot` entry being added to the DB:

```
      await this.db.transferRoots.update(transferRootId, {
        ...
        totalAmount: amount,
        ...
      })
```

File: packages/hop-node/src/db/TransferRootsDb.ts (hunk 4 above)
Comment (shanefontaine) [reply packages/hop-node/src/db/TransferRootsDb.ts:305]: Oh nice that is an awesome update! I really like that.

Only suggestion now is to remove the newly-unused `filterOutTimestampedKeys()`.

**Comment by shanefontaine** [comment]
@miguelmota Back to you. Left some more comments on the fixes. Also curious what you think about the second bullet in the top-level review [here](https://github.com/hop-protocol/hop/pull/318#pullrequestreview-832105547).

File: packages/hop-node/src/watchers/SyncWatcher.ts (hunk 9 above)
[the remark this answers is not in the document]
Comment (miguelmota) [reply packages/hop-node/src/watchers/SyncWatcher.ts:162]: Removed in https://github.com/hop-protocol/hop/pull/318/commits/75db07adbd7005fadbda27102c80d24358b9789b

File: packages/hop-node/src/db/TransferRootsDb.ts (hunk 4 above)
Comment (miguelmota) [reply packages/hop-node/src/db/TransferRootsDb.ts:305]: Removed in https://github.com/hop-protocol/hop/pull/318/commits/75db07adbd7005fadbda27102c80d24358b9789b

File: packages/hop-node/src/watchers/SyncWatcher.ts (hunk 5 above)
Comment (miguelmota) [reply packages/hop-node/src/watchers/SyncWatcher.ts]: Great catch, added in https://github.com/hop-protocol/hop/pull/318/commits/75db07adbd7005fadbda27102c80d24358b9789b

File: packages/hop-node/src/watchers/SyncWatcher.ts (hunk 13 above)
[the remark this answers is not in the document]
Comment (miguelmota) [reply packages/hop-node/src/watchers/SyncWatcher.ts]: Alright I think commit https://github.com/hop-protocol/hop/pull/318/commits/ab26f1c6823f50a0c041515c52e10a94be8611f1 should achieve the expected order now

**Comment by miguelmota** [comment]
> * I believe `transferRootId` should be passed in [here](https://github.com/hop-protocol/hop/blob/feature/index-transfer-root-id/packages/hop-node/src/watchers/ChallengeWatcher.ts#L70-L75) and the associated code in the above `for` loop can be removed. The logger should also be updated to use the `transferRootId` instead of the hash.

Updated with https://github.com/hop-protocol/hop/pull/318/commits/43a9da3c4f59b6788f81be76eb08d43b82098aca

>  * (Somewhat related to this PR and to a comment on another line below) I believe [these](https://github.com/hop-protocol/hop/blob/601583548ac557e9fff701534640de48e5b61222/packages/hop-node/src/watchers/SyncWatcher.ts#L392-L397) `throw`s should be deleted since, (1) that data should always exist at that point in the code and (2) these handlers are now `pure` and I think throwing makes them un`pure`

Updated with https://github.com/hop-protocol/hop/pull/318/commits/75db07adbd7005fadbda27102c80d24358b9789b

**Comment by miguelmota** [comment]
Conflicts have been resolve so this should be good now for review @shanefontaine

File: packages/hop-node/src/watchers/classes/L1Bridge.ts [hunk 14]
```diff
@@ -274,4 +274,22 @@ export default class L1Bridge extends Bridge {
   getBondForTransferAmount = async (amount: BigNumber): Promise<BigNumber> => {
     return await this.l1BridgeContract.getBondForTransferAmount(amount)
   }
+
+  async decodeBondTransferRootCalldata (data: string): Promise<any> {
+    if (!data) {
+      throw new Error('data to decode is required')
+    }
+    const decoded = await this.l1BridgeContract.interface.decodeFunctionData(
```
shanefontaine, line 282: nitpick: This does not need to be awaited.

shanefontaine (approved):
Looks wonderful to me. Left a tiny nitpick, but everything else looks ready to merge, IMO. Great work.

@miguelmota

PR #319, opened 2021-12-12
Use ENS name as account identifier if it exists

jackmpcollins:
Show ENS name instead of truncated address in transactions pill if one exists for the address.

Before
Screen Shot 2021-12-12 at 12 04 05 AM

After
Screen Shot 2021-12-12 at 12 03 50 AM

### DIFF
--- a/packages/frontend/src/components/header/TxPill.tsx
+++ b/packages/frontend/src/components/header/TxPill.tsx
@@ -1,5 +1,4 @@
 import React, { useEffect, useState } from 'react'
-import { ethers } from 'ethers';
 import { makeStyles } from '@material-ui/core/styles'
 import CircularProgress from '@material-ui/core/CircularProgress'
 import { useApp } from 'src/contexts/AppContext'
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/header/TxPill.tsx
```diff
@@ -1,4 +1,5 @@
 import React, { useEffect, useState } from 'react'
+import { ethers } from 'ethers';
```
kangarang, line 2: i believe this line can be removed (extraneous)

packages/frontend/src/components/header/TxPill.tsx
```diff
@@ -44,6 +46,12 @@ const TxPill = () => {
     }
   }, [transactions])
 
+  useEffect(() => {
+    if (address) {
+      provider?.lookupAddress(address.toString()).then(setENSName);
+    }
+  })
```
kangarang, line 53: add useEffect dependencies `[address, provider]`

kangarang (changes requested):
hi @jackmpcollins, thanks for this pr. currently, these changes will only work if the user is connected to ethereum mainnet. ideally, we should display the user's ens name regardless of connected network. instead of using the connected provider, i recommend grabbing the mainnet provider via the util [`getProviderByNetworkName('ethereum')`](https://github.com/hop-protocol/hop/blob/develop/packages/frontend/src/utils/getProvider.ts#L17) and using that to resolve the ens name.

jackmpcollins:
Thanks for the review @kangarang . Updated now.

kangarang (approved):
LGTM. thanks again for your contribution @jackmpcollins

PR #321, opened 2021-12-14
frontend: display native token amount necessary for tx fee

### LINKED ISSUE #253: Show more detailed warning message when account eth is too low for fee
Expected: Show how much more ETH is required for transaction fee if account already has ETH
Actual: Message states to add ETH to account even though account already has ETH

DeepinScreenshot_select-area_20211027135335

lukerhoads:
Fixes #253

### DIFF
--- a/packages/frontend/src/pages/Send/Send.tsx
+++ b/packages/frontend/src/pages/Send/Send.tsx
@@ -287,33 +287,39 @@ const Send: FC = () => {
   const needsTokenForFee = useNeedsTokenForFee(fromNetwork)
 
   useEffect(() => {
+    let isSubscribed = true
     const update = async () => {
-      let warning = ''
       try {
-        if (needsTokenForFee && fromNetwork && toNetwork && fromTokenAmountBN && fromBalance && deadline) {
+        if (needsTokenForFee && fromNetwork && toNetwork && fromTokenAmountBN && fromBalance && deadline && sourceToken?.isNativeToken) {
           const options = {
-            balance: fromBalance,
             token: sourceToken,
             fromNetwork,
             toNetwork,
             deadline,
           }
     
-          var estimatedGasCost = await estimateSend(options) 
+          const estimatedGasCost = await estimateSend(options) 
 
           if (estimatedGasCost && fromBalance?.lt(estimatedGasCost)) {
-            var shortBalance = estimatedGasCost.sub(fromBalance)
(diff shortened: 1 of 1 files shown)


miguelmota:
Thanks for the PR @lukerhoads!

miguelmota:
@kangarang it would be good if you can review this PR when you get a chance

packages/frontend/src/pages/Send/Send.tsx
```diff
@@ -287,14 +287,34 @@ const Send: FC = () => {
   const needsTokenForFee = useNeedsTokenForFee(fromNetwork)
 
   useEffect(() => {
-    if (needsTokenForFee && fromNetwork) {
-      setNeedsNativeTokenWarning(
-        `Add ${fromNetwork.nativeTokenSymbol} to your account on ${fromNetwork.name} for the transaction fee.`
-      )
-    } else {
-      setNeedsNativeTokenWarning('')
+    const update = async () => {
```
kangarang, line 290: since this hook makes async calls, we want to avoid any timing issues if this component gets unmounted for any reason. simple fix: follow the `isSubscribed` pattern that is [implemented here](https://github.com/hop-protocol/hop/blob/develop/packages/frontend/src/pages/Send/Send.tsx#L172)

[more info on cancelling hook promises](https://juliangaramendy.dev/blog/use-promise-subscription)

thanks, @miguelmota for the insight

packages/frontend/src/pages/Send/Send.tsx
```diff
@@ -287,14 +287,34 @@ const Send: FC = () => {
...
+            balance: fromBalance,
+            token: sourceToken,
+            fromNetwork,
+            toNetwork,
+            deadline,
+          }
+    
+          var estimatedGasCost = await estimateSend(options) 
+
+          if (estimatedGasCost && fromBalance?.lt(estimatedGasCost)) {
```
kangarang, line 304: if this block of code should **only** run if the user is trying to send a native token asset, let's explicitly state it in the `if` condition -- `sourceToken.isNativeToken`.

kangarang: matter of fact, we can be just as explicit while exiting this hook quicker if we check `sourceToken.isNativeToken` in the earlier `if` condition

kangarang (changes requested):
hi @lukerhoads, thank you for this pr. generally looks good! please see my review comments for some small suggestions.

lukerhoads:
Thanks for the review @kangarang , learned something new with the isSubscribed pattern 🙌

kangarang (approved):
LGTM, will merge soon. thank you for your contribution @lukerhoads

PR #324, opened 2021-12-15
feat: add token to metamask

kangarang:
this pr adds a feature that allows users to add a transferred asset to their wallet. users will have to switch networks in order to add the asset to the destination network of the transfer. this is an additional step, but we feel this feature would not be as useful if the user could only add the asset to the connected network.

additionally, clear `sourceTokenAmount` in `/convert` if `tx` exists (similar to 01e237b)


miguelmota:
This is a really nice feature. The functionality works as expected. Should we should show the add button as soon as the modal is shown rather than waiting for both transfers to complete? I think if user closes the modal then there's no way to access the button to add asset

kangarang:
@miguelmota i think that's a good idea. this is totally out scope, but i also would like to allow users to re-open the modal once they close it. still thinking how best to accomplish this though, as there is a lot of context. it might just be better to use the new refactored `useTransaction` hook.

miguelmota (approved):

PR #325, opened 2021-12-16
Feature/post pr updates

shanefontaine:
Some clean up after resettle PR and transferRootId PR have been merged. Two bigger items:

1. Added `mapTransferRootBondedEvents()` to the array of promises that need to be completed prior to running `mapMultipleWithdrawalsSettledEvents()`. This is needed because, technically, the `TransferRootBonded` event can be seen prior to the `TransfersCommitted` event, which would not work without serializing as seen below.
2. Removed return statements from `handleTransferSentEvent()`. It definitely makes sense here because the two items we were `return`ing on will always exist in the event data that we are reading, so those were never going to be reached. 
3. Re-introduced a `throw` in the `handleMultipleWithdrawalsSettledEvent()` function (accidentally pushed it to develop [here](https://github.com/hop-protocol/hop/commit/d814bb716ff0d5550c1b3729304f10dcae615474)). I think you were correct originally and this is a good throw. Ideally, this never happens because we either see the `TransfersCommitted` event or the `TransferRootBonded` event first. However, if we were to completely miss one of those events for some reason (i.e. RPC doesn't give it to us, or we have a bug in our code), then I believe we should explicitly fail in order to investigate instead of silently "missing" the event. Happy to hear any thoughts you have.

### DIFF
--- a/packages/hop-node/src/watchers/ChallengeWatcher.ts
+++ b/packages/hop-node/src/watchers/ChallengeWatcher.ts
@@ -80,12 +80,12 @@ class ChallengeWatcher extends BaseWatcher {
     if (!tx) {
       throw new Error('expected transaction object')
     }
+
     const { destinationChainId } = await l1Bridge.decodeBondTransferRootCalldata(tx.data)
-    const transferRootCommittedAt = await l1Bridge.getTransferRootCommittedAt(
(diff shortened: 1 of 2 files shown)


miguelmota (approved):
Looks good, don't see any issues :+1:

PR #326, opened 2021-12-16
core/sdk: Move bonder fee bps info to core package and read config from s3

miguelmota:
This PR does the following
- Moves sdk bonder fee bps to `core` package
- Moves sdk gas price price multiplier to `core` package
- Adds script `npm run upload` to upload `core` config as JSON file to s3
- Updates `sdk` to use bonder fee bps from `core` package
- Updates `sdk` to use gas price multiplier from `core` package
- Updates `sdk` to fetch and read S3 JSON config file on initialization and set bonder addresses, bonder fee bps, and gas price multiplier based on values from S3 JSON files
- If it can't fetch the S3 file, it defaults to the values of the `core` package config


packages/sdk/src/Base.ts
```diff
@@ -75,8 +76,9 @@ class Base {
   private addresses = config.addresses
   private chains = config.chains
   private bonders = config.bonders
-  private fees = config.fees
-  gasPriceMultiplier: number = 0
+  fees : { [token: string]: Record<string, number>}
+  gasPriceMultiplier: number = 1
```
shanefontaine, line 80: What is the difference between `gasPriceMultiplier` and `destinationFeeGasPriceMultiplier`?

packages/sdk/src/Base.ts
```diff
@@ -106,6 +108,28 @@ class Base {
...
+    try {
+      const data = await this.getS3ConfigData()
+      if (data.bonders) {
+        this.bonders = data.bonders
+      }
+      if (data.bonderFeeBps) {
+        this.fees = data.bonderFeeBps
+      }
+      if (data.gasPriceMultiplier) {
+        this.destinationFeeGasPriceMultiplier = data.gasPriceMultiplier
```
shanefontaine, line 128: * I don't believe `this.destinationFeeGasPriceMultiplier` should be set to `gasPriceMultiplier`. Logically, if these are intended to accomplish different things, then setting them as equal is rather confusing. I may be missing something here though.
* I think this will happen unintentionally since `gasPriceMultiplier` defaults to `1`.

packages/sdk/src/HopBridge.ts
```diff
@@ -2013,7 +2012,7 @@ class HopBridge extends Base {
       sourceChain
     )
 
-    const feeBps = this.getFeeBps(this.tokenSymbol, destinationChain)
+    const feeBps = await this.getFeeBps(this.tokenSymbol, destinationChain)
```
shanefontaine, line 2015: I do not believe this needs to be `await`ed. I think the `async` keyword can be removed from `getFeeBps`.

shanefontaine (changes requested):
* I recognize that this was already the case because of `bridgeDeployedBlockNumber`, but I believe the directory `addresses` should be renamed, as now it encompasses fees and multipliers.

* Based on the result of the first bullet point, I think the types should accurately reflect the change. For example, `bonderFeeBps` and `gasPriceMultiplier` shouldn’t be under the `Addresses` type.

* IMO `bonderFeeBps` and `gasPriceMultipler` types shouldn’t be optional, as we have run into issues when we forget to define config options in various places. Subsequently, I’d imagine all of the testnet files define these values as well (even if they are just `0` for the fees and `1` for the multiplier.

@miguelmota back to you

packages/sdk/src/Base.ts
miguelmota: `gasPriceMultiplier` is the multiplier the user specifies for when sending transactions using sdk. It's different than `destinationFeeGasPriceMultiplier` because `destinationFeeGasPriceMultiplier` is only for estimating gas for destination tx if chain is L1, or ORU

packages/sdk/src/Base.ts
miguelmota: I'll update the naming which is causing confusion

packages/sdk/src/HopBridge.ts
miguelmota: Resolved by https://github.com/hop-protocol/hop/pull/326/commits/79f100a3e97a6a2fb07c9b2dcab5a6ad405d85de

packages/sdk/src/Base.ts
miguelmota: The naming should be consistent now (https://github.com/hop-protocol/hop/pull/326/commits/79f100a3e97a6a2fb07c9b2dcab5a6ad405d85de)

`gasPriceMultiplier` is different than `destinationFeeGasPriceMultiplier`

- `destinationFeeGasPriceMultiplier` is what used be in `src/constants` as `GasPriceMultiplier` which was _only_ used for estimating destination gas cost fee if destination is L1, or ORU
- `gasPriceMultiplier` is the multiplier for `gasPrice` when sending transactions and this is used by developers

miguelmota:
> I recognize that this was already the case because of `bridgeDeployedBlockNumber`, but I believe the directory `addresses` should be renamed or refactored, as now it encompasses fees and multipliers & not just addresses.

The bonder fee and multiplier config has been moved to a `config` directory under `core/`. Updated with https://github.com/hop-protocol/hop/pull/326/commits/855639874ebccf77773d1de83875b5577bad4665

> Based on the result of the first bullet point, I think the types should accurately reflect the change. For example, `bonderFeeBps` and `gasPriceMultiplier` shouldn’t be under the `Addresses` type.

The types have been updated now that the config is outside of `addresses/`. Updated with https://github.com/hop-protocol/hop/pull/326/commits/855639874ebccf77773d1de83875b5577bad4665

> IMO `bonderFeeBps` and `gasPriceMultipler` types shouldn’t be optional, as we have run into issues when we forget to define config options in various places. Subsequently, I’d imagine all of the testnet files define these values as well (even if they are just `0` for the fees and `1` for the multiplier.
 
Agree, each network as a config now. Updated with https://github.com/hop-protocol/hop/pull/326/commits/855639874ebccf77773d1de83875b5577bad4665

---

Let me know if you any comments about the above

shanefontaine (approved):
LGTM 👍 

@miguelmota

PR #327, opened 2021-12-17
UI: add react query

kangarang:
this pr adds `react-query` as a frontend dependency. as an example implementation, the `useBalance` hook is refactored to follow the react-query pattern for fetching async data.

[react-query docs](https://react-query.tanstack.com/guides/queries)

### DIFF
--- a/packages/frontend/src/hooks/useBalance.tsx
+++ b/packages/frontend/src/hooks/useBalance.tsx
@@ -1,64 +1,28 @@
-import { useState, useRef, useEffect, useCallback } from 'react'
-import { BigNumber } from 'ethers'
+import { useQuery } from 'react-query'
 import { Token } from '@hop-protocol/sdk'
 import Network from 'src/models/Network'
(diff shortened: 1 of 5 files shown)


packages/frontend/src/hooks/useBalance.tsx
```diff
@@ -1,64 +1,28 @@
...
+  const { isLoading, isError, data, error } = useQuery(
+    ['balance', token?.symbol, network?.slug, address?.toString],
+    () => {
+      if (token && address) {
+        return fetchBalance(token, address)
       }
+    },
+    {
+      enabled: !!token?.symbol && !!address?.toString,
+      refetchInterval: 8e3,
```
miguelmota, line 20: I think we can update this to 5 seconds (`5 * 1000`) to be consistent with other pollers

miguelmota (approved):
Very nice PR, keeps the code clean

Issue #216, opened 2021-10-10
Check the available credit for the BondTransferRoot watcher in the DB getter, similar to the BondWithdrawal watcher


shanefontaine:
The BondWithdrawal watcher has an `availableCredit` check that looks at the local state in the DB getter in order to avoid on-chain calls every poll. The BondTransferRoot watcher can use this logic as well in order to avoid the repeated on-chain calls.

miguelmota:
@shanefontaine is this still an issue?

shanefontaine:
@miguelmota Yes it is. The BondTransferRoot watcher should have some logic similar to [this](https://github.com/hop-protocol/hop/blob/d5b56ad8d23f17d8989587bce5d868ea3729495d/packages/hop-node/src/watchers/BondWithdrawalWatcher.ts#L76-L86) in order to optimally send transactions at the appropriate time.

The name/description of this issue was incorrectly worded, so I just updated it.

miguelmota:
Thanks, I'll handle these changes

miguelmota:
Closed by https://github.com/hop-protocol/hop/pull/328

PR #328, opened 2021-12-17
hop-node: Check available credit in BondTransferRootWatcher before processing db entry

miguelmota:
Addresses issue https://github.com/hop-protocol/hop/issues/216


packages/hop-node/src/watchers/BondTransferRootWatcher.ts
```diff
@@ -63,6 +63,15 @@ class BondTransferRootWatcher extends BaseWatcher {
         transferIds
       } = dbTransferRoot
 
+      const availableCredit = this.getAvailableCreditForBond(destinationChainId!) // eslint-disable-line @typescript-eslint/no-non-null-assertion
+      if (availableCredit.lt(totalAmount!)) { // eslint-disable-line @typescript-eslint/no-non-null-assertion
+        this.logger.debug(
```
shanefontaine, line 68: I think it makes sense to define the `logger` here with the `transferRootId`, similar to the [BondWithdrawalWatcher](https://github.com/hop-protocol/hop/blob/33384d0da7f1e6de16ff80fc6cbc018067c91de3/packages/hop-node/src/watchers/BondWithdrawalWatcher.ts#L74) for easier debugging.

packages/hop-node/src/watchers/BondTransferRootWatcher.ts
```diff
@@ -170,6 +179,11 @@ class BondTransferRootWatcher extends BaseWatcher {
       throw err
     }
   }
+
+  getAvailableCreditForBond (destinationChainId: number) {
+    const availableCredit = this.syncWatcher.getEffectiveAvailableCredit(destinationChainId)
```
shanefontaine, line 184: * Since the bond is occurring on L1, we need to get the available credit on L1
* The `effectiveAvailableCredit` returns `// L2 -> L1: (credit - debit - OruToL1PendingAmount - OruToAllUnbondedTransferRoots)`. For transfer roots, we just need `credit - debit`, like we do in original TransferRootWatcher [check](https://github.com/hop-protocol/hop/blob/33384d0da7f1e6de16ff80fc6cbc018067c91de3/packages/hop-node/src/watchers/BondTransferRootWatcher.ts#L138). With that said, it does not make sense to poll on-chain every 10s for both `credit` and `debit`. It would only make sense for this change if we expose a `this.rawAvailableCredit` (or something like that) that just returns an in-memory `credit - debit` that is polled in `syncBonderCredit` (or something like that). IIRC this change is a little heavier than I'd hope and is why I didn't just make this change w/o a PR. It looks like we already [call `getBaseAvailableCredit()`](https://github.com/hop-protocol/hop/blob/33384d0da7f1e6de16ff80fc6cbc018067c91de3/packages/hop-node/src/watchers/SyncWatcher.ts#L1136), which is what we need, in `calculateAvailableCredit()`, so maybe we can expose it there?
  * If that update is made, then I think it also makes sense to update the [original check](https://github.com/hop-protocol/hop/blob/33384d0da7f1e6de16ff80fc6cbc018067c91de3/packages/hop-node/src/watchers/BondTransferRootWatcher.ts#L138) to use this new in-memory check, just like we do with the BondWithdrawalWatcher ([getter lookup](https://github.com/hop-protocol/hop/blob/33384d0da7f1e6de16ff80fc6cbc018067c91de3/packages/hop-node/src/watchers/BondWithdrawalWatcher.ts#L75) is same as [check function lookup](https://github.com/hop-protocol/hop/blob/33384d0da7f1e6de16ff80fc6cbc018067c91de3/packages/hop-node/src/watchers/BondWithdrawalWatcher.ts#L144)).

shanefontaine (changes requested):
Back to you. Left a higher-level comment in one of the in-line comments.

I also pushed a commit (db98d30f74e41d8499cb99db63d3a58ed7c4b159) to move the `sentBondTxAt` update to the appropriate location (right before sending). We had to move it before the `availableCredit` check _because_ we didn't have this PR. I was planning on moving this back after a PR like this was created.

@miguelmota

packages/hop-node/src/watchers/BondTransferRootWatcher.ts
miguelmota: @shanefontaine Thanks for the clarification. I understand what you are saying about using the base available credit instead. Check out the [latest changes](https://github.com/hop-protocol/hop/pull/328/files) which cache `baseAvailableCredit` in SyncWatcher and the BondTransferRootWatcher uses now. I think this correct but let me know

packages/hop-node/src/watchers/BondTransferRootWatcher.ts
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/328/commits/5d98c91084f4ba98b4c6cb907f3a6269a380be8f

shanefontaine:
Nice that is awesome.

In the `BondWithdrawalWatcher`, we [check the available credit against the in-memory data](https://github.com/hop-protocol/hop/blob/314434d60e8ddd7f407c25019dd6283f757aab1b/packages/hop-node/src/watchers/BondWithdrawalWatcher.ts#L144) in the `checkTransferId()` function. However, we [still check in on-chain](https://github.com/hop-protocol/hop/blob/a8395e555bec35c241865b38599c13b2a40043b7/packages/hop-node/src/watchers/BondTransferRootWatcher.ts#L145) in the new `BondTransferRootWatcher`. Since the in-memory state is just a direct 1:1 with the on-chain state, my loose vote would be to check against the in-memory state here.

Other than that, everything LGTM and can be merged.

@miguelmota

miguelmota:
@shanefontaine thanks for the review. The additional change is here https://github.com/hop-protocol/hop/pull/328/commits/54c4e93a250051b96fdb68738893ba8e1f2d5122 . Going to merge it now

PR #345, opened 2021-12-21
Fix/resettlement issues

shanefontaine:
PR fixes a resettlement issue by correctly checking on-chain if the `transferRoot` has been completely settled. I had previously performed an invalid check.

There was another issue with the settlements that don’t require any code change but have caused recent issues. I [recently added](https://github.com/hop-protocol/hop/commit/e60597cbca1b50d7fc9953a1fcc174563aa79113#diff-d7cca2a07e0034090a9c7c5ec85ba2e1a0dedd759b6535a4a8847ed937e0ac0fR460) a DB item when observing a `WithdrawalBonded` event. Because of this addition, we were able to guarantee that [this check](https://github.com/hop-protocol/hop/blob/314434d60e8ddd7f407c25019dd6283f757aab1b/packages/hop-node/src/watchers/SettleBondedWithdrawalWatcher.ts#L97) will be either explicitly be `true` or `false`. The code is valid, however I did not consider the state of the DB prior to the first addition. I should have migrated the DB accordingly. The DB at that time does not have these guarantees, therefore the check is `true` and a settlement is attempted. Since all bonders are running the latest code AND they only look for DB entries w/in two weeks, all of the “old” DB entries will be phased out over the next few days, so I do not believe there is any code change required. Any


miguelmota:
I think the changes look good. What is a scenario where the `onChainTotalAmount` will be different than the db `totalAmount` for a given transferRootId?

shanefontaine:
The only case would be if there was a bug and `totalAmount` was never defined in the DB or if there was an unhandled re-org. I don't expect either of those to occur.

More generally, since we are making the on-chain call already that provides both values, then I think we should just use what is returned to us, since it reflects the _true_ on-chain state (as opposed to our DB, which _should be_ true). For the most part, I believe the `checkTransferId()` checks in all watchers should be reserved for on-chain checks (where appropriate) just to ensure we do not send an erroneous tx on-chain.

miguelmota (approved):

miguelmota:
@shanefontaine makes sense and agree with the reasoning. PR is good to merge

Issue #337, opened 2021-12-17
The bonder should not start if the config file is not correctly populated


shanefontaine:
_Up for discussion._

There have been issues where config files are not correctly populated and the bonder software defaults to a non-ideal value. The most extreme case of this is the `waitConfirmations` defaulting to `0` and putting a bonder at risk of a reorg.

I believe the bonder software should be very explicit about these values and whoever is running the bonder should input these value intentionally. I believe the bonder software should not start unless these values are explicitly defined in the config. Very rarely will any default values apply to all cases, as bonder liquidity, risk tolerance, fees, etc. are all on a per-bonder bases.

Some config options that this applies to:
_**This is not an inclusive list. Please independently review any other potential config options that might apply.**_
* `minThresholdAmount`
* `fees`
* `routes`
* `roles`
* `watchers`
* `chains`
* `waitConfirmations`

miguelmota:
Since `minThresholdAmount` is only required for the `CommitTransfersWatcher` and `fees` is `BondWithdrawalWatcher`, then does it make sense to only require those if those watchers are enabled?

shanefontaine:
Yes, good call. I think that would be ideal.

miguelmota:
Closed by https://github.com/hop-protocol/hop/pull/346

PR #346, opened 2021-12-21
hop-node: Config: Require certain config to be set before starting. #337

miguelmota:
PR updates

- require `tokens` config
- require `chains` config
- require `chains.rpcUrl` config
- require `chains.waitConfirmations` config
- require `routes` config
- require `fees` config (only if watcher `bondWithdrawal` is enabled)
- require `commitTransfers.minThresholdAmount` config (only if watcher `commitTransfers` is enabled)
- require `watchers` config
- require `roles` config
- validate `chains.rpcUrl` url config
- validate `chains.waitConfirmations` value
- validate `tokens` keys
- validate `routes` source and destination keys
- validate `fees` token keys
- validate `fees` token chain keys based on configured route destination chains
- validate `watchers` keys
- validate `roles` keys
- validate  `commitTransfers.minThresholdAmount` config based on configured tokens and routes


packages/hop-node/src/config/fileOps.ts
```diff
@@ -173,11 +203,37 @@ export async function setGlobalConfigFromConfigFile (
...
+  if (!config?.watchers) {
+    throw new Error('config for watchers is required')
+  }
+
+  setRoutesConfig(config.routes)
+  if (enabledWatchers.includes(Watchers.BondWithdrawal) && !config?.fees) {
+    throw new Error('config for fees is required')
+  }
+
+  if (config.fees != null) {
```
shanefontaine, line 227: Why does this `if` statement do nothing if it does not exist but the others throw (i.e., why does `if (!config.watchers)` throw but this does not)?

packages/hop-node/src/watchers/classes/ContractBase.ts
```diff
@@ -135,7 +135,15 @@ export default class ContractBase extends EventEmitter {
   }
 
   get waitConfirmations () {
-    return globalConfig.networks[this.chainSlug]?.waitConfirmations ?? 0
+    const chainConfig = globalConfig.networks[this.chainSlug]
+    if (!chainConfig) {
+      throw new Error(`config for chain ${this.chainSlug} not found`)
+    }
+    const { waitConfirmations } = chainConfig
+    if (typeof waitConfirmations !== 'number') {
```
shanefontaine, line 143: See comment above about checking for the value `0` here.

packages/hop-node/src/config/validation.ts
```diff
@@ -68,32 +69,76 @@ export async function validateConfig (config?: FileConfig) {
...
+    validateKeys(validNetworkKeys, enabledChains)
+    const validChainKeys = ['rpcUrl', 'waitConfirmations']
+    for (const chain in config.chains) {
+      validateKeys(validChainKeys, Object.keys(config.chains[chain]))
+      if (!config.chains[chain]) {
+        throw new Error(`RPC config for chain "${chain}" is required`)
+      }
+      const { rpcUrl, waitConfirmations } = config.chains[chain]
+      if (!rpcUrl) {
+        throw new Error(`RPC url for chain "${chain}" is required`)
```
shanefontaine, line 93: I believe all bonders, regardless of their path, need access to a mainnet RPC endpoint in order to sync `TransferRootSet` (and others, if it is an ORU). I believe we should enforce this here. Maybe with `enabledChains.push('mainnet')` on line 84?

packages/hop-node/src/config/validation.ts
```diff
@@ -68,32 +69,76 @@ export async function validateConfig (config?: FileConfig) {
...
+        if (!parsed.protocol || !parsed.host || !['http:', 'https:'].includes(parsed.protocol)) {
+          throw new URIError()
+        }
+      } catch (err) {
+        throw new Error(`rpc url "${rpcUrl}" is invalid`)
+      }
+      if (waitConfirmations == null) {
+        throw new Error(`waitConfirmations for chain "${chain}" is required`)
+      }
+      if (typeof waitConfirmations !== 'number') {
```
shanefontaine, line 106: See comment above about checking for the value `0` here.

packages/hop-node/src/config/validation.ts
```diff
@@ -68,32 +69,76 @@ export async function validateConfig (config?: FileConfig) {
...
+  ]
+
+  let enabledChains: string[] = []
   if (config.chains) {
-    const networkKeys = Object.keys(config.chains)
-    await validateKeys(validNetworkKeys, networkKeys)
+    enabledChains = Object.keys(config.chains)
+    validateKeys(validNetworkKeys, enabledChains)
+    const validChainKeys = ['rpcUrl', 'waitConfirmations']
+    for (const chain in config.chains) {
```
shanefontaine, line 86: I believe we should to enforce that the RPC urls exist for all routes that the bonder has. For example, if the bonder is bonding on xDai -> Polygon, we should enforce that they have an RPC endpoint for xDai, Polygon, and mainnet (see comment below about mainnet).

Does this code already do that? If not, do you think this is practical/possible?

shanefontaine (changes requested):
Back to you. I really like the `WatchersConfig` type and how it is used. Left a few routine comments.

packages/hop-node/src/config/config.ts
```diff
@@ -294,5 +251,14 @@ export const setRoutesConfig = (routes: Routes) => {
 
 export const chainNativeTokens = ['ETH', 'MATIC', 'DAI']
 
+export enum Watchers {
+  BondTransferRoot = 'bondTransferRoot',
+  BondWithdrawal = 'bondWithdrawal',
+  Challenge = 'challenge',
```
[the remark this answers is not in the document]
miguelmota: Good idea to remove `roles` to keep things simple. Resolved with https://github.com/hop-protocol/hop/pull/346/commits/36a93478a3aa1d9fccd0c560ea29b4a32c1059bb

packages/hop-node/src/config/fileOps.ts
```diff
@@ -22,12 +25,12 @@ import { recoverKeystore } from 'src/keystore'
...
 export const defaultEnabledWatchers: { [key: string]: boolean } = {
-  bondTransferRoot: true,
-  bondWithdrawal: true,
-  challenge: true, // only active if role.challenger is also true
-  commitTransfers: true,
-  settleBondedWithdrawals: true,
-  xDomainMessageRelay: false
+  [Watchers.BondTransferRoot]: false,
+  [Watchers.BondWithdrawal]: false,
+  [Watchers.Challenge]: false, // only active if role.challenger is also true
```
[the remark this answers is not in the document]
miguelmota: Concept of roles has been removed. Resolved with https://github.com/hop-protocol/hop/pull/346/commits/36a93478a3aa1d9fccd0c560ea29b4a32c1059bb

packages/hop-node/src/config/fileOps.ts
```diff
@@ -22,12 +25,12 @@ import { recoverKeystore } from 'src/keystore'
 const logger = new Logger('config')
 
 export const defaultEnabledWatchers: { [key: string]: boolean } = {
```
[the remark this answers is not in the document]
miguelmota: The concept of `defaultEnabledWatchers` has been removed because all watchers are disabled by default. Resolved with https://github.com/hop-protocol/hop/pull/346/commits/36a93478a3aa1d9fccd0c560ea29b4a32c1059bb

packages/hop-node/src/watchers/classes/ContractBase.ts
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/346/commits/36a93478a3aa1d9fccd0c560ea29b4a32c1059bb

packages/hop-node/src/config/validation.ts
miguelmota: It checks now if `waitConfirmations >= 0`. Resolved with https://github.com/hop-protocol/hop/pull/346/commits/36a93478a3aa1d9fccd0c560ea29b4a32c1059bb

packages/hop-node/src/config/fileOps.ts
miguelmota: It's because `fees` is only required if the `bondWithdrawal` watcher is enabled, while others, like `watchers`, is always required. The validation checks the values of `fees`

packages/hop-node/src/config/validation.ts
miguelmota: nice catch. The validation now requires that `ethereum` chain config is set. Resolved with https://github.com/hop-protocol/hop/pull/346/commits/1878d754ab9e40806a9a1748530403686306e2bb

packages/hop-node/src/cli/hopNode.ts
```diff
@@ -112,7 +100,7 @@ async function main (source: any) {
   }
   for (const k in globalConfig.networks) {
     const { waitConfirmations, rpcUrl } = globalConfig.networks[k]
-    if (!waitConfirmations) {
+    if (typeof waitConfirmations !== 'number') {
```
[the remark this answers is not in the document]
miguelmota: It checks now if `waitConfirmations >= 0` in validation function. Resolved with https://github.com/hop-protocol/hop/pull/346/commits/36a93478a3aa1d9fccd0c560ea29b4a32c1059bb

packages/hop-node/src/config/validation.ts
miguelmota: Yes, it already validates that the chain config must exist for a route [here](https://github.com/hop-protocol/hop/blob/1878d754ab9e40806a9a1748530403686306e2bb/packages/hop-node/src/config/validation.ts#L172-L175)

miguelmota:
@shanefontaine thanks for the great feedback

packages/hop-node/src/watchers/classes/ContractBase.ts
shanefontaine, line 143: Is this validation needed? IIUC, `waitConfirmations` are validated (it is a number and is `> 0`) during the watcher creation, which (I think) always occurs before this.

packages/hop-node/src/watchers/classes/Bridge.ts
```diff
@@ -645,6 +645,9 @@ export default class Bridge extends ContractBase {
     }
 
     const bonderFeeBps = fees[destinationChain]
+    if (!bonderFeeBps) {
```
shanefontaine, line 648: Is this validation needed? IIUC, this is validated in the watcher creation which occurs prior to this.

shanefontaine (changes requested):
Left a few more minor comments.

One more top level item: I think it makes sense to get rid of the concept of `order` in the config and config options now as well, since it is unused. LMK if you agree.

@miguelmota

packages/hop-node/src/watchers/classes/Bridge.ts
miguelmota: I think it can be removed but it can be bad if for whatever reasons there is no bonder fee here so I like having the assertion check in there. I think the class shouldn't make assumptions

packages/hop-node/src/watchers/classes/ContractBase.ts
miguelmota: same reason as other similar comment

miguelmota:
Agree, the concept of `order` has been removed now https://github.com/hop-protocol/hop/pull/346/commits/133d9c4784a50464a885c95533fd6f493beada12

packages/hop-node/src/watchers/classes/Bridge.ts
shanefontaine: Makes sense & SGTM.

packages/hop-node/src/watchers/classes/ContractBase.ts
shanefontaine: Makes sense & SGTM.

shanefontaine (approved):
LGTM and can be merged 👍  Nice job.

PR #347, opened 2021-12-21
UI: add support for ens avatars

kangarang:
this pr adds support for ens avatars to the TxPill component


miguelmota (approved):
Looks good!

PR #349, opened 2021-12-21
hop-node: TransferRootsDb: Add migration to add missing root hash sub db entries

miguelmota:


shanefontaine (approved):
@miguelmota LGTM. Appears to have fixed the original issue I was seeing 👍

PR #350, opened 2021-12-21
hop-node: read existing DB entry before getting timestamped key

shanefontaine:
@miguelmota Changes are what was discussed. I verified it works on the first write when `dbTransferRoot` will be `null`.

### DIFF
--- a/packages/hop-node/src/db/TransferRootsDb.ts
+++ b/packages/hop-node/src/db/TransferRootsDb.ts
@@ -312,7 +312,9 @@ class TransferRootsDb extends BaseDb {
       return
     }
     const transferRootId = transferRoot.transferRootId
-    const key = this.getTimestampedKey(transferRoot)
+    const dbTransferRoot = this.getById(transferRootId!)
+    const combinedData = Object.assign({}, dbTransferRoot, transferRoot)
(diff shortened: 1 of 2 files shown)


packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -312,7 +312,9 @@ class TransferRootsDb extends BaseDb {
       return
     }
     const transferRootId = transferRoot.transferRootId
-    const key = this.getTimestampedKey(transferRoot)
+    const dbTransferRoot = this.getById(transferRootId!)
```
miguelmota, line 315: missing `await` here

packages/hop-node/src/db/TransfersDb.ts
```diff
@@ -177,7 +177,9 @@ class TransfersDb extends BaseDb {
       return
     }
     const transferId = transfer.transferId
-    const key = this.getTimestampedKey(transfer)
+    const dbTransfer = this.getById(transferId!)
```
miguelmota, line 180: missing `await` here

miguelmota (changes requested):

miguelmota:
Fee free to merge after the comment fixes

PR #352, opened 2021-12-24
Feature/get all graph items

shanefontaine:
Apparently TheGraph only allows for `skip`ping up to 6000 items.

* [Discord conversations](https://discord.com/channels/438038660412342282/438070183794573313/862337114019463199)
* Error message when we run large queries: `The skip argument must be between 0 and 5000, but is 6000`

The [docs](https://thegraph.com/docs/en/developer/graphql-api#pagination) mention "For retrieving a large number of items, it is much better to page through entities based on an attribute as shown in the last example.". Chris' recent TheGraph queries work this way as well. This PR attempts to do this by returning all items and using the ID as the index for the query and returning all IDs instead of being capped at 6000.

Only important note is that the order always needs to be `asc` for these searches, since it is "counting" up from `0x00000...` to `0xfffff` with this method.

### DIFF
--- a/packages/hop-node/src/theGraph/getBondedWithdrawals.ts
+++ b/packages/hop-node/src/theGraph/getBondedWithdrawals.ts
@@ -34,7 +34,7 @@ export default async function getBondedWithdrawals (
     token,
     lastId
   })
-  let withdrawals  = jsonRes.withdrawalBondeds.map((x: any) => normalizeEntity(x))
+  let withdrawals = jsonRes.withdrawalBondeds.map((x: any) => normalizeEntity(x))
 
   lastId = withdrawals[withdrawals.length - 1].id
   const maxItemsLength = 1000
--- a/packages/hop-node/src/theGraph/getBondedWithdrawals.ts
+++ b/packages/hop-node/src/theGraph/getBondedWithdrawals.ts
@@ -36,9 +36,9 @@ export default async function getBondedWithdrawals (
   })
   let withdrawals = jsonRes.withdrawalBondeds.map((x: any) => normalizeEntity(x))
 
-  lastId = withdrawals[withdrawals.length - 1].id
   const maxItemsLength = 1000
   if (withdrawals.length === maxItemsLength) {
+    lastId = withdrawals[withdrawals.length - 1].id
     withdrawals = withdrawals.concat(await getBondedWithdrawals(
       chain,
       token,
--- a/packages/hop-node/src/theGraph/getBondedWithdrawals.ts
+++ b/packages/hop-node/src/theGraph/getBondedWithdrawals.ts
@@ -1,15 +1,20 @@
 import makeRequest from './makeRequest'
 import { normalizeEntity } from './shared'
 
-export default async function getBondedWithdrawals (chain: string, token: string) {
+export default async function getBondedWithdrawals (
+  chain: string,
+  token: string,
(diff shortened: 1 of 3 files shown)


packages/hop-node/src/theGraph/getBondedWithdrawals.ts
```diff
@@ -26,7 +31,20 @@ export default async function getBondedWithdrawals (chain: string, token: string
...
   `
   const jsonRes = await makeRequest(chain, query, {
-    token
+    token,
+    lastId
   })
-  return jsonRes.withdrawalBondeds.map((x: any) => normalizeEntity(x))
+  let withdrawals = jsonRes.withdrawalBondeds.map((x: any) => normalizeEntity(x))
+
+  lastId = withdrawals[withdrawals.length - 1].id
```
miguelmota, line 39: is it possible that `withdrawals.length` here can be `0` so `.id` would throw an error trying to read from `undefined`?

packages/hop-node/src/theGraph/getTransferIds.ts
```diff
@@ -46,34 +46,26 @@ export default async function getTransferIds (
...
-        token,
-        filters,
-        skip + 1000
-      ))
-    } catch (err) {
-      if (!err.message.includes('The `skip` argument must be between')) {
-        throw err
-      }
-    }
+  lastId = transfers[transfers.length - 1].id
```
miguelmota, line 60: is it possible that `transfers.length` here can be `0` so `.id` would throw an error trying to read from `undefined`?

packages/hop-node/src/theGraph/getTransferRoots.ts
```diff
@@ -30,23 +34,19 @@ export default async function getTransferRoots (chain: string, token: string, sk
...
-        chain,
-        token,
-        skip + 1000
-      ))
-    } catch (err) {
-      if (!err.message.includes('The `skip` argument must be between')) {
-        throw err
-      }
-    }
+  lastId = roots[roots.length - 1].id
```
miguelmota, line 42: is it possible that `roots.length` here can be `0` so `.id` would throw an error trying to read from `undefined`?

miguelmota (commented):
Left minor code comments. I think the changes make sense. I'll try to run it as well to make sure it works as expected

packages/hop-node/src/theGraph/getBondedWithdrawals.ts
shanefontaine: Yes, great call. Fixed with 26ffbf5f93e5a0d1cd10ed6b2e87d8b1919e4177

packages/hop-node/src/theGraph/getTransferIds.ts
shanefontaine: Yes, great call. Fixed with 26ffbf5f93e5a0d1cd10ed6b2e87d8b1919e4177

packages/hop-node/src/theGraph/getTransferRoots.ts
shanefontaine: Yes, great call. Fixed with 26ffbf5f93e5a0d1cd10ed6b2e87d8b1919e4177

shanefontaine:
Comments addressed. Back to you.

@miguelmota

miguelmota:
Changes look great and works well 👍

PR #356, opened 2021-12-29
Fix/xdai gasprice

shanefontaine:
xDai has been returning very large `gasPrice` data [recently](https://blockscout.com/xdai/mainnet/tx/0xe113e7c90b0f854f9b435707cfa086b59fc2c5232877418a22cf5e672c3df8bc/logs) which causes us to hit our 500 Gwei cap. No other chain is exhibiting this behavior. These txs aren't getting boosted, they're just sent once.

The blocks are definitely not full, but there is weird behavior with other txs in the blocks -- some send 200+ Gwei while other send <5 gwei. I think this points to some weird data being returned form our RPC provider WRT how they aggregate gasPrice data.

This PR updates the `maxGasPrice` to a lower, reasonable number for xDai.

### DIFF
--- a/packages/hop-node/src/gasboost/GasBoostTransaction.ts
+++ b/packages/hop-node/src/gasboost/GasBoostTransaction.ts
@@ -334,6 +334,11 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
   }
 
   getMaxGasPrice () {
+    // The xDai RPC endpoint sometimes returns bad data and should be capped lower.
+    if (this.chainSlug === Chain.xDai) {
+      const maxGasPriceGweiXdai = 90
+      return this.parseGwei(maxGasPriceGweiXdai)
+    }
     return this.parseGwei(this.maxGasPriceGwei)
   }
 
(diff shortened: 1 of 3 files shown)


packages/hop-node/src/gasboost/GasBoostTransaction.ts
```diff
@@ -334,6 +334,11 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
   }
 
   getMaxGasPrice () {
+    // The xDai RPC endpoint sometimes returns bad data and should be capped lower.
+    if (this.chainSlug === Chain.xDai) {
+      const maxGasPriceGweiXdai = 90
```
miguelmota, line 339: I think this should be moved to be part of `src/config`

```js
export const maxGasPriceGweiXdai = 90
```

miguelmota (approved):

packages/hop-node/src/gasboost/GasBoostTransaction.ts
shanefontaine: Good call. Fixed in 342a0a5a26ab86a06e3365b599b6e21f9031cb80

PR #360, opened 2022-01-03
hop-node: Update config validation

miguelmota:


packages/hop-node/src/cli/shared/shared.ts
```diff
@@ -23,12 +26,15 @@ export function actionHandler (fn: Function) {
       if (configFilePath) {
```
shanefontaine, line 26: nitpick: if we now validate the config values outside of the `if (configFilePath)`, does that mean that we will always throw in `validateConfigValues(globalConfig)` if a path doesn't exist?

If so, then we can probably just do `if (!configFilePath) { throw new Error }` instead of this `if...else`.

packages/hop-node/src/config/validation.ts
```diff
@@ -225,6 +199,60 @@ export async function validateConfig (config?: FileConfig) {
...
+    }
+    try {
+      const parsed = new URL(rpcUrl)
+      if (!parsed.protocol || !parsed.host || !['http:', 'https:'].includes(parsed.protocol)) {
+        throw new URIError()
+      }
+    } catch (err) {
+      throw new Error(`rpc url "${rpcUrl}" is invalid`)
+    }
+    if (waitConfirmations != null) {
```
shanefontaine, line 237: I was under the impression this was going to be removed from the config validation since it is not a part of the config file. Is this `validateConfigValues` function used for more than validating just the config file?

shanefontaine (commented):
* I think [this](https://github.com/hop-protocol/hop/blob/b6e612e4e302fc52ffe4bffed132e50fd984ebc8/packages/hop-node/src/cli/hopNode.ts#L96-L98) check should have the requirements as [this check](https://github.com/hop-protocol/hop/pull/360/files#diff-f1fbdb9b05a352d4967ac9da74fadda906b36a38726bc4242222f344edca75e5R237-R245A). Basically add the `<= 0` here.


👍  Back to you @miguelmota

packages/hop-node/src/config/validation.ts
miguelmota: > Is this validateConfigValues function used for more than validating just the config file?

That is correct, it validates the in memory [`config`](https://github.com/hop-protocol/hop/blob/c58a99169aa6c08f3fbcba11b4d8586b909d3e62/packages/hop-node/src/config/config.ts#L125) object which has some values loaded by default (like `waitConfirmations`) so it would catch any misconfiguration values from the core package or values from config file

packages/hop-node/src/cli/shared/shared.ts
miguelmota: We can't throw here because this is a global handler and some commands don't require a config file (like keystore generation). I removed the `else` and throw in the hopNode command instead https://github.com/hop-protocol/hop/pull/360/commits/427827fe97b6fda91c4541ca4e50bd89dab7ee11

miguelmota:
> I think this check should have the requirements as this check. Basically add the <= 0 here.

Thanks, that's actually an artifact check that can be removed because the validation would throw if `<=0` before getting there. Removed in https://github.com/hop-protocol/hop/pull/360/commits/5213e7fc0bfa622990e4ec26153c46b8f7f6476b

shanefontaine:
LGTM 👍

PR #361, opened 2022-01-03
Fix/correct fee for htoken send

shanefontaine:
The SDK adjusts the bonder fee based on the AMM swap output even if the user is sending h tokens and not actually performing a swap. This PR fixes that by not adjusting the fees if the consumer of the SDK is simply sending hTokens.

### DIFF
--- a/packages/sdk/src/HopBridge.ts
+++ b/packages/sdk/src/HopBridge.ts
@@ -515,11 +515,11 @@ class HopBridge extends Base {
       adjustedBonderFee = BigNumber.from(0)
       adjustedDestinationTxFee = BigNumber.from(0)
     } else if (isHTokenSend) {
-      // sending hTokens do not need to be adjusted for AMM swaps
+      // fees do not need to be adjusted for AMM slippage when sending hTokens
       adjustedBonderFee = bonderFeeRelative
       adjustedDestinationTxFee = destinationTxFee
     } else {
-      // adjusted fee is the fee in the canonical token after adjusting for the hToken price.
+      // adjusted fee is the fee in the canonical token after adjusting for the hToken price
       adjustedBonderFee = await this.calcFromHTokenAmount(
         bonderFeeRelative,
         destinationChain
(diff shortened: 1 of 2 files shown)


packages/sdk/src/HopBridge.ts
```diff
@@ -520,7 +529,12 @@ class HopBridge extends Base {
         destinationTxFee,
         destinationChain
       )
+    }
 
+    let totalFee
+    if (isSourceL1) {
+      totalFee = BigNumber.from(0)
```
miguelmota, line 536: I think these duplicated conditional checks can be combined

```ts
let adjustedBonderFee
let adjustedDestinationTxFee
let totalFee
if (isSourceL1) {
  // there are no Hop fees when the source is L1
  adjustedBonderFee = BigNumber.from(0)
  adjustedDestinationTxFee = BigNumber.from(0)
  totalFee = BigNumber.from(0)
} else {
  if (isHTokenSend) {
    // fees do not need to be adjusted for AMM slippage when sending hTokens
    adjustedBonderFee = bonderFeeRelative
    adjustedDestinationTxFee = destinationTxFee
  } else {
    // adjusted fee is the fee in the canonical token after adjusting for the hToken price
    adjustedBonderFee = await this.calcFromHTokenAmount(
        bonderFeeRelative,
        destinationChain
      )

    adjustedDestinationTxFee = await this.calcFromHTokenAmount(
          destinationTxFee,
          destinationChain
     )
  }

  // enforce bonderFeeAbsolute after adjustment
  const bonderFeeAbsolute = await this.getBonderFeeAbsolute()
  adjustedBonderFee = adjustedBonderFee.gt(bonderFeeAbsolute)
    ? adjustedBonderFee
    : bonderFeeAbsolute
  totalFee = adjustedBonderFee.add(adjustedDestinationTxFee)
}

```

miguelmota (approved):
Updates make sense 👍

packages/sdk/src/HopBridge.ts
shanefontaine: Good call. Resolved with addb6908e0a46f745f05afd98071c01726e72846.

Issue #329, opened 2021-12-17
The hop-node package should not read the `bonder` addresses from the `core` package


shanefontaine:
The `hop-node` package currently reads the `bonder` addresses from the `core` package. However, the bonder should only be concerned about themself and being aware of other bonders is unnecessary except for the available credit watcher. Maybe the available credit watcher can read bonder addresses from the local config.

This fix should achieve two things:
* Bonder software should not rely on bonder addresses from `core` package
* Available credit watcher should still work as expected

miguelmota:
Closed by https://github.com/hop-protocol/hop/pull/365

PR #365, opened 2022-01-05
hop-node: Config: Read route bonder addresses from config file instead of core package. #329

miguelmota:
- Removes reliance on bonder addresses from `core` package
- Adds option to use `bonders` from config file

### DIFF
--- a/packages/hop-node/src/watchers/SyncWatcher.ts
+++ b/packages/hop-node/src/watchers/SyncWatcher.ts
@@ -16,7 +16,7 @@ import { L1ERC20Bridge as L1ERC20BridgeContract } from '@hop-protocol/core/contr
 import { L2Bridge as L2BridgeContract, TransferSentEvent, TransfersCommittedEvent } from '@hop-protocol/core/contracts/L2Bridge'
 import { Transfer } from 'src/db/TransfersDb'
 import { TransferRoot } from 'src/db/TransferRootsDb'
-import { config as globalConfig, oruChains } from 'src/config'
+import { getConfigBonderForRoute, config as globalConfig, oruChains } from 'src/config'
 
 type S3JsonData = {
   [token: string]: {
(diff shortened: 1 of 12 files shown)


packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -1201,12 +1201,17 @@ class SyncWatcher extends BaseWatcher {
...
   private async updateAvailableCreditMap (destinationChainId: number) {
     const destinationChain = this.chainIdToSlug(destinationChainId)
-    const bonder = this.bridge.getConfigBonderAddress(destinationChain)
+    const bonder = await this.getBonderAddress(destinationChain)
     const { availableCredit, baseAvailableCredit } = await this.calculateAvailableCredit(destinationChainId, bonder)
     this.availableCredit[destinationChain] = availableCredit
     this.baseAvailableCredit[destinationChain] = baseAvailableCredit
   }
 
+  async getBonderAddress (destinationChain: string): Promise<string> {
```
shanefontaine, line 1210: Can this be cached as to potentially not make an on-chain call every time? Alternatively, can/should `this.bridge.getBonderAddress()` be cached (I don't see why not)?

shanefontaine (commented):
Code looks good and runs as expected, I believe.

* It probably makes sense to log the bonder(s) on bonder startup now that they can be defined in the config.
* Updated config option should be added to docs [table](https://docs.hop.exchange/hop-node/configuration#properties) and [its own information section with expected format](https://docs.hop.exchange/hop-node/configuration#fees) (I do not believe it should be added to the example config above the table because that is used throughout the docs as a reference config file to use when starting up the bonder).
* Loose opinion, there should be some validation that `bonders` in the config file is an object. I originally used a string and the program appeared to run even though I believe it would not have worked as expected.
* Curious: what happens if a bonder is bonding x -> p USDC but defines bonders on a -> o USDT (something completely different). Will the bonder just populate available credits for both or will the latter overwrite the former?

packages/hop-node/src/watchers/SyncWatcher.ts
miguelmota: I don't think `signer.getAddress()` does any on-chain calls. The method is async on the ethers signer because some wallets might require async processing to retrieve the address but there's no on-chain rpc call

shanefontaine: Understood and agreed 👍

miguelmota:
>    It probably makes sense to log the bonder(s) on bonder startup now that they can be defined in the config.

Updated with https://github.com/hop-protocol/hop/pull/365/commits/05cefdd82fb8d27068f0f33d458d88fa0bf8948b
 
>    Updated config option should be added to docs [table](https://docs.hop.exchange/hop-node/configuration#properties) and [its own information section with expected format](https://docs.hop.exchange/hop-node/configuration#fees) (I do not believe it should be added to the example config above the table because that is used throughout the docs as a reference config file to use when starting up the bonder).

Docs [updated](https://docs.hop.exchange/hop-node/configuration#bonders)

>    Loose opinion, there should be some validation that `bonders` in the config file is an object. I originally used a string and the program appeared to run even though I believe it would not have worked as expected.

Validation added with Updated with https://github.com/hop-protocol/hop/pull/365/commits/05cefdd82fb8d27068f0f33d458d88fa0bf8948b

>   Curious: what happens if a bonder is bonding x -> p USDC but defines bonders on a -> o USDT (something completely different). Will the bonder just populate available credits for both or will the latter overwrite the former?

The validation would throw an error in this case if the config for `bonders` contains a route that is not specified in `routes`

shanefontaine:
LGTM. Can be merged 👍 

@miguelmota

Issue #341, opened 2021-12-17
TransferRoot DB items are indexed by `committedAt` timestamp but it might not always be the case that `committedAt` exists in order to index the entry


shanefontaine:
_This issue was inspired by [this discussion](https://github.com/hop-protocol/hop/pull/318#discussion_r769187693)._

We currently index transfer roots [by their `committedAt` timestamp](https://github.com/hop-protocol/hop/blob/f10a3234399581b021239e4fbfb39c135f8bc50e/packages/hop-node/src/db/TransferRootsDb.ts#L297) and act on these DB entries accordingly. This works most of the time. However, there are edge cases where a malicious bonder will bond an invalid `transferRoot` that was never committed. In this case, there will be no `committedAt` timestamp to index the DB entry by.

I believe the implications of a change like this would be large, so it is worth a discussion and general consensus prior to implementation.

miguelmota:
@shanefontaine  Since we store the `bondedAt` timestamp in db, we can add another sub db to index by the bonded timestamp. In `getChallengeableTransferRoots` we can then retrieve a list of transferRootIds by querying this sub (`subDbBondedAt`) to get the transfer root db items. It's a pretty simple addition but let me know what you think

miguelmota:
Closed by https://github.com/hop-protocol/hop/pull/366

PR #366, opened 2022-01-05
hop-node: TransferRootsDb: Add sub db to track transfer root items by bondedAt timestamp. #341

miguelmota:
Adds sub db to index by `bondedAt` timestamp

### DIFF
--- a/packages/hop-node/src/db/TransferRootsDb.ts
+++ b/packages/hop-node/src/db/TransferRootsDb.ts
@@ -1,4 +1,4 @@
-import BaseDb, { KeyFilter } from './BaseDb'
+import BaseDb, { KV, KeyFilter } from './BaseDb'
 import chainIdToSlug from 'src/utils/chainIdToSlug'
 import { BigNumber } from 'ethers'
 import { Chain, ChallengePeriodMs, OneHourMs, OneWeekMs, RootSetSettleDelayMs, TxRetryDelayMs } from 'src/constants'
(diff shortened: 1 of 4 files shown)


packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -1,41 +1,41 @@
...
 import { normalizeDbItem } from './utils'
 import { oruChains } from 'src/config'
 
 interface BaseTransferRoot {
-  transferRootHash?: string
-  totalAmount?: BigNumber
-  destinationChainId?: number
-  sourceChainId?: number
-  sentCommitTxAt?: number
+  allSettled?: boolean
```
shanefontaine, line 9: nitpick: this alphabetization is awesome. can we do the same with `TransfersDb`.

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -378,6 +405,18 @@ class TransferRootsDb extends BaseDb {
     })
   }
 
+  async getBondedTransferRootsFromTwoWeeks (): Promise<TransferRoot[]> {
+    await this.tilReady()
+    const fromUnix = Math.floor((Date.now() - (OneWeekMs * 2)) / 1000)
+    const filter: KeyFilter = {
+      gte: `${fromUnix}`
```
shanefontaine, line 412: Do we need to add the prefix here along with the `fromUnix` value, like [this](https://github.com/hop-protocol/hop/blob/feature/341/packages/hop-node/src/db/TransferRootsDb.ts#L372)?

shanefontaine (commented):
@miguelmota back to you

packages/hop-node/src/db/TransferRootsDb.ts
miguelmota: The prefix is not necessary in this case because the db is already namespaced by using a sub db (each sub db already requires a prefix in the constructor). The prefix on the other sub dbs are also unnecessary but can't be removed without doing a migration

packages/hop-node/src/db/TransferRootsDb.ts
miguelmota: Will do

shanefontaine:
LGTM 👍

PR #367, opened 2022-01-05
hop-node: Config: Use AWS region from config instead of hardcoded region

miguelmota:
- Uses AWS region from env var instead of hardcoded region if available
- Add ability to set specific region for AWS parameter store


shanefontaine (commented):
Code looks good.

* Does this new structure apply for both the CloudWatch logs as well as the paramstore? Should we have separate config options for them? IIRC, one of the bonders currently has paramstore on a different region than CloudWatch. If so, I think the entity running the bonder should specify two regions, one for CloudWatch and one for paramstore.
* I think the docs should probably be updated [here](https://docs.hop.exchange/hop-node/docker-cloudwatch-logs) to reflect this change, if applicable

miguelmota:
>   Does this new structure apply for both the CloudWatch logs as well as the paramstore? 

Docker CloudWatch log region is set in `docker-compose.yml` (or by using `--log-opt awslogs-region` flag with `docker run`). This is separate from `src/aws/cloudWatch` where this for the cli command to log cloudwatch logs and this uses the environment variable because this command doesn't require the config file. The `awsRegion` value in the `keystore` config is for the `parameterStore` AWS region.

> Should we have separate config options for them? IIRC, one of the bonders currently has paramstore on a different region than CloudWatch. 

I don't think this is the case because it's always been the `us-east-1` region

> If so, I think the entity running the bonder should specify two regions, one for CloudWatch and one for paramstore. The code may already do this and I'm not realizing it, so please let me know if so.

The Docker CloudWatch region has always been configurable via `docker-compose.yml` (or as a flag to `docker run`). The parameter store region was hardcoded until this PR

>  I think the docs should probably be updated [here](https://docs.hop.exchange/hop-node/docker-cloudwatch-logs) (both `bash` and `docker-compose`) to reflect this change, if applicable

Thanks, the docs have been updated

shanefontaine:
All SGTM. I misunderstood some pieces of this originally, but your explanations cleared everything up. Looks good to merge, IMO.

PR #368, opened 2022-01-06
All: rename xdai chain to gnosis

kangarang:


packages/core/src/networks/mainnet.ts
```diff
@@ -25,8 +25,8 @@ export const networks: Networks = {
     nativeBridgeUrl: 'https://gateway.optimism.io/welcome',
     waitConfirmations: 1
   },
-  xdai: {
-    name: chains.xdai.name,
+  gnosis: {
+    name: chains.gnosis.name,
     networkId: 100,
     publicRpcUrl: 'https://rpc.xdaichain.com',
```
miguelmota, line 31: All `https://rpc.xdaichain.com` should be changed to `https://rpc.gnosischain.com`

packages/sdk/tsconfig.json
```diff
@@ -11,7 +11,7 @@
     "declarationMap": true,
     "skipLibCheck": true,
     "target": "es2015",
-    "noImplicitAny": true,
+    "noImplicitAny": false,
```
shanefontaine, line 14: Any reason why this changed?

shanefontaine (commented):
Code looks good and the hop-node appears to run as expected. Some higher level notes/questions:

* Does the SDK allow for people to pass in either `gnosis` or `xdai` when instantiating the SDK?
* Are we making a new gnosis-chain subgraph?
* The docs need to be updated to reflect this change (can be done after merge)
* The hop-node DB needs a migration before running this on production bonders (can be done after merge)

Nice job!

kangarang:
@shanefontaine thanks for the review!

> Does the SDK allow for people to pass in either gnosis or xdai when instantiating the SDK?

good question. i believe this is a breaking change, and SDK consumers will have to instantiate using 'gnosis'

ex:

```
const amm = new AMM('mainnet', Token.USDC, Chain.Gnosis)
```

> Are we making a new gnosis-chain subgraph?

i think we can leave the subgraph as `hop-xdai`. we handle the discrepancy in [queries.ts](https://github.com/hop-protocol/hop/blob/all/rename-xdai-chain-to-gnosis/packages/frontend/src/utils/queries.ts#L9), which should suffice for now.

packages/sdk/tsconfig.json
kangarang: ah, i was fooling around with some tests, and meant to change this back to `true`. good catch

miguelmota:
We'll need to use the gnosis chain logo on frontend

https://assets.hop.exchange/logos/gnosis.svg

kangarang:
TODO after merge:

- update s3 config (`https://assets.hop.exchange/mainnet/v1-core-config.json`) / uncomment line 128 in `sdk/src/Base.ts`
- update env var: REACT_APP_ENABLED_CHAINS='ethereum,arbitrum,optimism,polygon,gnosis'
- run hop-db migration
- update sdk docs

shanefontaine (approved):
LGTM!

PR #371, opened 2022-01-07
UI: fix add xdai to metamask

kangarang:
this pr sets the token symbol to DAI (when adding a token to metamask) if the asset is XDAI and the destination network is not xDai

### DIFF
--- a/packages/frontend/src/hooks/useAddTokenToMetamask.ts
+++ b/packages/frontend/src/hooks/useAddTokenToMetamask.ts
@@ -8,7 +8,7 @@ export function useAddTokenToMetamask(
   token?: Token | null,
   destNetworkName?: string | null
 ): {
-  addToken: () => void
+  addToken: (networkId: number) => void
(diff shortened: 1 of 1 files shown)


packages/frontend/src/hooks/useAddTokenToMetamask.ts
```diff
@@ -17,9 +17,14 @@ export function useAddTokenToMetamask(
   const [success, setSuccess] = useState<boolean>(false)
 
   const addToken = useCallback(
-    (networkId?: number) => {
+    (networkId: number) => {
       if (provider && token) {
-        const { symbol, image, decimals } = token
+        let { symbol, image, decimals } = token
+
+        if (symbol === 'XDAI' && networkId !== 100) {
```
miguelmota, line 24: maybe it'd be better to compare against the slug here instead of hardcoded network number

```ts
Chain.xDai !== networkIdToSlug(networkId)
```

miguelmota (approved):
👍

PR #374, opened 2022-01-12
fix errors when routing w/ unsupported url params

kangarang:
previously, when routing to the `/pool` or `/convert` routes,
the `selectedNetwork` would initialize as defaultL2Network (`l2Networks[0]`), causing
errors if the asset was set to `MATIC` (defaultL2Network = arbitrum)

Screen Shot 2022-01-11 at 8 15 39 PM

- fix `defaultL2Network`: `MATIC` not supported on arbitrum/optimism
- add checks for `unsupportedAsset`
- clean up types


kangarang:
closing. rebased into #377

PR #375, opened 2022-01-13
UI: improve ens handling and system styles

kangarang:
this pr improves ens handling and cleans up the custom design system component styles

fixes an issue where the TxPill does not switch properly when going from an account that resolves to an ens name -> one that does not

### DIFF
--- a/packages/frontend/src/contexts/AppContext/useNetworks.ts
+++ b/packages/frontend/src/contexts/AppContext/useNetworks.ts
@@ -34,7 +34,7 @@ const useNetworks = () => {
   return {
     networks: allNetworks,
     l2Networks,
-    defaultL2Network: l2Networks[0],
+    defaultL2Network: l2Networks[2],
   }
 }
 
--- a/packages/frontend/src/hooks/useEns.ts
+++ b/packages/frontend/src/hooks/useEns.ts
@@ -1,33 +1,49 @@
+import { isAddress } from 'ethers/lib/utils'
 import { useEffect, useState } from 'react'
 import logger from 'src/logger'
-import Address from 'src/models/Address'
-import { getEnsAvatar, getEnsName } from 'src/utils/ens'
+import { getEnsAddress, getEnsAvatar, getEnsName } from 'src/utils/ens'
 
-export function useEns(address?: Address) {
+export function useEns(addressOrEnsName?: string) {
   const [ensAvatar, setEnsAvatar] = useState<string>()
-  const [ensName, setEnsName] = useState<string>()
+  const [ensName, setEnsName] = useState<string | null>()
+  const [ensAddress, setEnsAddress] = useState<string>()
 
   useEffect(() => {
-    if (address?.address) {
-      try {
-        getEnsName(address.address).then(en => en && setEnsName(en))
-      } catch (err) {
-        logger.error(`error during setEnsName:`, err)
-      }
+    setEnsName('')
+    setEnsAddress('')
+    setEnsAvatar('')
+
+    if (!addressOrEnsName) {
+      return
     }
-  }, [address?.address])
+
+    try {
+      if (isAddress(addressOrEnsName)) {
+        getEnsName(addressOrEnsName).then(setEnsName)
+      } else {
+        getEnsAddress(addressOrEnsName).then(resolvedAddress => {
+          if (resolvedAddress) {
+            setEnsName(addressOrEnsName)
+            setEnsAddress(resolvedAddress)
+          }
+        })
+      }
+    } catch (err) {
+      logger.error(`error during setEnsName/setEnsAddress:`, err)
+    }
+  }, [addressOrEnsName])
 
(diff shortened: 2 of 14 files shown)


packages/frontend/src/contexts/AppContext/useNetworks.ts
```diff
@@ -34,7 +34,7 @@ const useNetworks = () => {
   return {
     networks: allNetworks,
     l2Networks,
-    defaultL2Network: l2Networks[0],
+    defaultL2Network: l2Networks[2],
```
kangarang, line 37: cherry-picked from #374

packages/frontend/src/hooks/useEns.ts
```diff
@@ -1,33 +1,49 @@
...
+    setEnsAddress('')
+    setEnsAvatar('')
+
+    if (!addressOrEnsName) {
+      return
+    }
+
+    try {
+      if (isAddress(addressOrEnsName)) {
+        getEnsName(addressOrEnsName).then(setEnsName)
```
miguelmota, line 22: should set `setEnsAddress` be set in addition to `setEnsName`, similar to how both are set in the else statement?

packages/frontend/src/hooks/useEns.ts
```diff
@@ -1,33 +1,49 @@
...
+    if (ensNameOrAddress) {
       try {
-        getEnsAvatar(addrOrEnsName).then(setEnsAvatar)
+        getEnsAvatar(ensNameOrAddress).then(setEnsAvatar)
       } catch (err) {
         logger.error(`error during setEnsAvatar:`, err)
       }
     }
-  }, [address, ensName])
+  }, [ensName])
```
miguelmota, line 46: does `addressOrEnsName` need to be included in dependency array?

packages/frontend/src/utils/ens.ts
```diff
@@ -22,3 +22,15 @@ export async function getEnsAvatar(ensNameOrAddress: string) {
     // noop
   }
 }
+
+export async function getEnsAddress(ensName: string) {
+  if (!ensName.endsWith('.eth')) {
```
miguelmota, line 27: Technically there can be other ENS other than `.eth`, for example

DeepinScreenshot_select-area_20220113175234

I think we should attempt to resolve any ENS name here

packages/frontend/src/utils/styles.ts
```diff
@@ -33,9 +33,19 @@ export type ComposedStyleProps = BackgroundProps &
...
 
-export function squareDimensions({ size }: { size?: number | string }) {
+export interface SquareDimensions {
+  size?: number | string
+  width?: number | string
+}
+
+export function squareDimensions({ size, width }: SquareDimensions) {
   if (!size) {
     size = 24
```
miguelmota, line 43: the default value can be set in function signature

```typescript
squareDimensions({ size = 24, width }: SquareDimensions) {
```

packages/frontend/src/hooks/useEns.ts
miguelmota, line 22: shouldn't there be an `await` here in order for it to be caught in try/catch? otherwise it throws an `Uncaught in promise` error

error uncaught without await

DeepinScreenshot_select-area_20220113175714

error caught with await

DeepinScreenshot_select-area_20220113175751

packages/frontend/src/hooks/useEns.ts
```diff
@@ -1,33 +1,49 @@
...
 
   useEffect(() => {
-    const addrOrEnsName = ensName || address?.address
+    const ensNameOrAddress = ensName || addressOrEnsName
 
-    if (addrOrEnsName) {
+    if (ensNameOrAddress) {
       try {
-        getEnsAvatar(addrOrEnsName).then(setEnsAvatar)
+        getEnsAvatar(ensNameOrAddress).then(setEnsAvatar)
```
miguelmota, line 41: same comment about using await here

packages/frontend/src/utils/ens.ts
```diff
@@ -22,3 +22,15 @@ export async function getEnsAvatar(ensNameOrAddress: string) {
...
+    return
+  }
+
+  try {
+    return provider.resolveName(ensName)
```
miguelmota, line 32: I don't think the try/catch here is effective because it's missing an await after the return keyword

miguelmota (changes requested):

packages/frontend/src/hooks/useEns.ts
kangarang: i figure we should only display the avatar if a valid `ensName` is present

packages/frontend/src/hooks/useEns.ts
kangarang: will use `.catch()`

miguelmota (approved):
@kangarang updates look good👌

PR #376, opened 2022-01-14
hop-node: SyncWatcher: Mark root as not found if there's no TransfersCommitted end event found

miguelmota:
This PR marks a transfer root as `isNotFound: true` if an `endEvent` for `TransfersCommitted` event cannot be found. 

Currently it just [returns](https://github.com/hop-protocol/hop/blob/288c4e2582d5fc0c582477a079dc311abe9fb3be/packages/hop-node/src/watchers/SyncWatcher.ts#L1003) from the function causing an endless loop, using RPC resources.

For example,
bonder: ETH
transferRootId: `0x916959add0d2e687fe94a67bb431c0eab2bdaeeb8d3143d03dc2b5b7ee0c07eb`
transferRootHash: `0x855eb94754f6ce53853ab81e927c7cdd8de068b896766234f490c551aef017bb`

CloudWatch logs showing endless loop trying to find transferIds for the transfer root hash
 
DeepinScreenshot_select-area_20220113172628

DeepinScreenshot_select-area_20220113174357

The transfer root hash cannot be found in subgraphs

Arbitrum subgraph
DeepinScreenshot_select-area_20220113173024

Optimism subgraph
DeepinScreenshot_select-area_20220113173004

xDai subgraph
DeepinScreenshot_select-area_20220113173045

Polygon subgraph
DeepinScreenshot_select-area_20220113173111

Therefore this transfer root should be marked as not found so it doesn't try to be processed in the sync watcher.


shanefontaine:
This looks great. Would it be reasonable to remove the invalid arrays [1](https://github.com/hop-protocol/hop/blob/5560c91438c48eec02b00693d2a269a40a60f141/packages/hop-node/src/db/TransferRootsDb.ts#L58) [2](https://github.com/hop-protocol/hop/blob/5560c91438c48eec02b00693d2a269a40a60f141/packages/hop-node/src/db/TransfersDb.ts#L55) with this update?

miguelmota:
@shanefontaine I think they can be removed. Do you foresee any possible issues in doing so?

miguelmota:
There were two more places where we need to set a transfer id and transfer root as not found.

---

Problem:

A transfer root would try to get processed by [`populateTransferRootDbItem`](https://github.com/hop-protocol/hop/blob/22839afeb6041ad54885bccfacbbf2d132483231/packages/hop-node/src/watchers/SyncWatcher.ts#L672) endlessly if there is no `sourceChainId` for that root.

This is because [`isItemIncomplete`](https://github.com/hop-protocol/hop/blob/22839afeb6041ad54885bccfacbbf2d132483231/packages/hop-node/src/db/TransferRootsDb.ts#L608) returns true if `sourceChainId` is missing (which is technically correct because the item is incomplete if it's missing these properties)

DeepinScreenshot_select-area_20220114194435

The reason that there is no `sourceChainId` (or `destinationChainId` and `commitTxBlockNumber` for that matter) is because there's no associated `TransfersCommitted` event for that transfer root hash (I checked all subgraphs and pure handlers didn't log event)

I added a check to mark it as not found in `populateTransferRootDbItem` otherwise it'll go through all the populate methods successfully but then get processed again in populateTransferRootDbItem in the next poll because it's missing `sourceChainId`

DeepinScreenshot_select-area_20220114194117

This screenshot shows the sync handlers called for the transfer root (note there's no `handling TransfersCommitted event` log) and shows it endless looping the populate call.


DeepinScreenshot_select-area_20220114125802

This was from a fresh sync.

transferRootId: `0xfe8af196071d57f00a51816bc0193c59d4b2d14111fc6f40863496e463cd678d`

---

The same issue happens with transfer ids.

I added a similar check to `populateTransferDbItem`

DeepinScreenshot_select-area_20220114194111

This screenshot shows no `handling TransferSent event` event from a fresh sync and it continue polls the populate call for this transfer id.

DeepinScreenshot_select-area_20220114131226

transferId: `0xffff7c3073f97d0174cef459aaef0739b9daf3de979fbf2275cc98c2d1a03ac0`

---

I ran the bonder from a fresh sync without the hard coded preregensis transfer roots and transfer ids.

---

Side note: It look about ~3-4hrs to do a full sync. I also tried a full sync without the db batching and that took ~1-2hrs. The reasons it was faster was because there's no need to wait (no [`batchTimeLimit`](https://github.com/hop-protocol/hop/blob/22839afeb6041ad54885bccfacbbf2d132483231/packages/hop-node/src/db/BaseDb.ts#L60)) but the CPU usage did spike way higher compared to using batching. Memory usage was similar as with batching. This is known info but thought I'd mention it again.

shanefontaine:
Awesome data and great finds! That stuff isn't easy to nail down. It all LGTM and can be merged 👍

Issue #339, opened 2021-12-17
Add DB and subDB types or visualization


shanefontaine:
_Up for debate._

With the addition of new subDBs, it is not obvious what the structure of their contents look like without prior context or a robust understanding. This becomes problematic for new or existing contributors when it comes to DB-specific work. DB migrations, in particular, are difficult to reason about without this knowledge.

If possible, providing a type for the structure of these DBs would be ideal. Less ideal would be some sort of visualization.

As a non-working pseudocode example:

```
type SubDbTimestampsKey: string = '<transferRootId>:<committedAtTimestamp>'
type SubDbTimestampsValue: string = '<transferRootId>'
type SubDbTimestamps: Record<SubDbTimestampsKey, SubDbTimestampsValue>

...

subDbTimestamps:  SubDbTimestamaps
```

or

```
type DbTransferRootsKey: string = '<transferRootId>'
type DbTransferRootsValue: TransferRoot
type DbTransferRoots: Record<DbTransferRootsKey, DbTransferRootsValue>

...

dbTransferRoots:  DbTransferRoots
```

miguelmota:
Closed by https://github.com/hop-protocol/hop/pull/378

PR #378, opened 2022-01-19
hop-node: Add subdb classes

miguelmota:
- Addresses https://github.com/hop-protocol/hop/issues/339
- Wraps each sub db in a class to contain methods only related to them
- Add comments about structure key/value


shanefontaine (approved):
LGTM. I think the new architecture is super nice -- specifically the DRYness of the function names 👍

PR #379, opened 2022-01-20
UI: cleanup

kangarang:


packages/frontend/src/hooks/useTxResult.ts
```diff
@@ -0,0 +1,46 @@
...
+export function useTxResult(
+  token?: Token,
+  srcNetwork?: Network,
+  destNetwork?: Network,
+  amount?: BigNumber,
+  fn?: (opts: any) => any,
+  opts?: any
+) {
+  const queryKey = `txResult:${srcNetwork?.slug}:${destNetwork?.slug}:${amount?.toString()}`
+  const { interval, ...rest } = opts
```
miguelmota, line 15: minor suggestion:

the default value could be set here:

```typescript
const { interval = 5 * 1000, ...rest } = opts
```

packages/frontend/src/hooks/useTxResult.ts
```diff
@@ -0,0 +1,46 @@
...
+  const { isLoading, isError, data, error } = useQuery(
+    [queryKey, srcNetwork?.slug, token?.symbol, amount?.toString()],
+    async () => {
+      if (!(token && srcNetwork?.slug && destNetwork?.slug && amount && fn)) {
+        return
+      }
+
+      const options = {
+        token: token,
+        fromNetwork: srcNetwork,
```
miguelmota, line 26: minor suggestion:

This suggestion applies to the entire application;
it would be good to choose consistent naming for the network directions; `fromNetwork` vs `sourceNetwork`

The contracts refer to them as `sourceChain` and `destinationChain` so maybe we use that convention throughout

packages/frontend/src/hooks/useAssets.ts
```diff
@@ -4,23 +4,23 @@ import { hopAppNetwork } from 'src/config'
 import logger from 'src/logger'
 import Network from 'src/models/Network'
 
-export function useAssets(selectedBridge?: HopBridge, fromNetwork?: Network, toNetwork?: Network) {
+export function useAssets(selectedBridge?: HopBridge, network?: Network, toNetwork?: Network) {
```
miguelmota, line 7: minor suggestion;

I think the previous var name `fromNetwork` was more descriptive than having `network` because there is also a `toNetwork`

packages/sdk/src/AMM.ts
```diff
@@ -117,7 +118,7 @@ class AMM extends Base {
       amounts,
       minToMint,
       deadline
-    ]
+    ] as const
```
miguelmota, line 121: nice

packages/sdk/src/Base.ts
```diff
@@ -1,18 +1,54 @@
...
 
 const s3FileCache : Record<string, any> = {}
 
+function isL2Factory (arg: any): arg is L2Factory {
+  const factories = [L2PolygonChildERC20__factory, L2XDaiToken__factory, ArbERC20__factory, L2OptimismTokenBridge__factory]
+  if (arg instanceof L2PolygonChildERC20__factory || arg instanceof L2XDaiToken__factory || arg instanceof ArbERC20__factory || arg instanceof L2OptimismTokenBridge__factory) {
+    return true
+  }
+  return false
+  // const names = factories.map(f => f.createInterface().contractName)
```
miguelmota, line 48: commented code should be removed

miguelmota (commented):
This is an awesome PR of removing dead code and cleaning up. The code changes looks good. I want to do a final pass through to check for regressions before merging

packages/frontend/src/hooks/useTxResult.ts
kangarang: yeah, i was doing some thinking about `Network` / `Chain` models as well, and i believe we could consolidate them (just use `Chain`). totally agree about keeping consistent naming. for now, changing `fromNetwork` -> `sourceChain` and `toNetwork` -> `destinationChain` causes a huge diff, so i will address this in a more focused, near-future commit/pr.

packages/frontend/src/hooks/useAssets.ts
kangarang: i changed it to `network` here because this hook could technically work with just one network (both the `fromNetwork` and the `toNetwork`), which is the case for [`/convert`](https://github.com/hop-protocol/hop/blob/09b876b84df96874613044bb82e384a1a5e0eb05/packages/frontend/src/pages/Convert/ConvertContext.tsx#L129) and `/pool`. but i also see it from your point of view. lemme know what you think ^^ @miguelmota

miguelmota:
issue:

expected: The unsupported asset warning to show when token is MATIC and chain is either Optimism or Arbitrum on send page
actual: No warning is displayed

DeepinScreenshot_select-area_20220121165540

DeepinScreenshot_select-area_20220121165545

packages/frontend/src/hooks/useAssets.ts
miguelmota: that sounds good to me,
no strong opinion

packages/frontend/src/hooks/useSelectedNetwork.ts
```diff
@@ -0,0 +1,79 @@
...
+  l2Only: boolean
+}
+
+export function useSelectedNetwork(opts: Options = { l2Only: false }) {
+  const [selectedNetwork, setSelectedNetwork] = useState<Network>(defaultL2Network)
+  const { queryParams, updateQueryParams } = useQueryParams()
+
+  useEffect(() => {
+    if (queryParams?.sourceNetwork !== selectedNetwork.slug) {
+      const matchingNetwork = findNetworkBySlug(queryParams.sourceNetwork)
```
miguelmota, line 24: I get this type error on start

DeepinScreenshot_select-area_20220125172921

Type asserting as string solves the issue

 `queryParams.sourceNetwork as string`

miguelmota:
expected: show matic not supported when switching from arbitrum to optimism
actual: no error shown, error in console

Peek 2022-01-25 17-35

PR #382, opened 2022-01-24
hop-node: L2Bridge: Use Type 0 tx type for Polygon commitTransfers txs

miguelmota:
- Uses `type: 0` for tx type for `commitTransfers` if source chain is polygon

### DIFF
--- a/packages/hop-node/src/gasboost/GasBoostTransaction.ts
+++ b/packages/hop-node/src/gasboost/GasBoostTransaction.ts
@@ -35,6 +35,7 @@ type MarshalledItem = {
   id: string
   createdAt: number
   txHash: string
+  type?: number
   from: string
   to: string
   data: string
@@ -92,6 +93,8 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
   private _is1559Supported: boolean // set to true if EIP-1559 type transactions are supported
   readonly minMultiplier: number = 1.10 // the minimum gas price multiplier that miners will accept for transaction replacements
 
+  type?: number
+
   // these properties are required by ethers TransactionResponse interface
   from: string // type 0 and 2 tx required property
   to: string // type 0 and 2 tx required property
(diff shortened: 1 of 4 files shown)


shanefontaine:
@miguelmota does the `type` need to be defined throughout the `GasBoostTransaction` class as well? Some poentital places: [1](https://github.com/hop-protocol/hop/blob/develop/packages/hop-node/src/gasboost/GasBoostTransaction.ts#L108-L163), [2](https://github.com/hop-protocol/hop/blob/develop/packages/hop-node/src/gasboost/GasBoostTransaction.ts#L238-L253) [3](https://github.com/hop-protocol/hop/blob/develop/packages/hop-node/src/gasboost/GasBoostTransaction.ts#L261-L271), [4](https://github.com/hop-protocol/hop/blob/develop/packages/hop-node/src/gasboost/GasBoostTransaction.ts#L300-L308) [5](https://github.com/hop-protocol/hop/blob/develop/packages/hop-node/src/gasboost/GasBoostTransaction.ts#L589-L595)

miguelmota:
@shanefontaine you are correct. The code has been updated and I added a test to verify type 0 txs

packages/hop-node/src/gasboost/GasBoostTransaction.ts
```diff
@@ -374,7 +385,8 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
   }
 
   async getBumpedGasFeeData (multiplier: number = this.gasPriceMultiplier): Promise<Partial<GasFeeData>> {
-    const use1559 = await this.is1559Supported() && !this.gasPrice
+    const use1559 = await this.is1559Supported() && !this.gasPrice && this.type !== 0
```
shanefontaine, line 388: Technically if `use1559` should always be false if `tyis.type !== 0`, so maybe it is `const use1559 = (await this.is1559Supported() && !this.gasPrice) || this.type !== 0`

shanefontaine (approved):
@miguelmota one small comment, but feel free to merge after addressed. The rest LGTM :+1:

packages/hop-node/src/gasboost/GasBoostTransaction.ts
shanefontaine: Updated the comment from `this.type !== 0` to `this.type ===0`

miguelmota: > Technically if use1559 should always be false if this.type === 0

That's what the `&& this.type !== 0` is doing

DeepinScreenshot_select-area_20220124130712

so I think it's correct

Issue #330, opened 2021-12-17
Bonders still bond withdrawals even with `dry` mode on


shanefontaine:
In some cases, bonders will bond withdrawals even with `dry` mode enabled. As an example, search the ETH bonder log stream `63082aefe1f5fcf9f4f77b67474fcc4c7f86a567d1d3f1ff0729ab6620bb08bd` for "sent bondWithdrawal on" and "dry mode:" to see that `dry` mode is on but transactions were processed. 

The `docker.sh` file at the time was:

```
docker run \
  ...
  -v ~/.hop-node:/root \
  $docker_image \
  --config /root/config.json \
  --dry
```

miguelmota:
The dry mode check in [`BondWithdrawalWatcher`](https://github.com/hop-protocol/hop/blob/f10a3234399581b021239e4fbfb39c135f8bc50e/packages/hop-node/src/watchers/BondWithdrawalWatcher.ts#L159) is the same as the other watchers (such as [`CommitTransfersWatcher`](https://github.com/hop-protocol/hop/blob/f10a3234399581b021239e4fbfb39c135f8bc50e/packages/hop-node/src/watchers/CommitTransfersWatcher.ts#L143), [`BondTransferRootWatcher`](https://github.com/hop-protocol/hop/blob/f10a3234399581b021239e4fbfb39c135f8bc50e/packages/hop-node/src/watchers/BondTransferRootWatcher.ts#L150), [`SettleBondedWithdrawalWatcher`](https://github.com/hop-protocol/hop/blob/f10a3234399581b021239e4fbfb39c135f8bc50e/packages/hop-node/src/watchers/SettleBondedWithdrawalWatcher.ts#L191)) and it's checked before sending the tx, so I don't think it's due to the dry mode setting. It could be related to the transfer state not properly updating

miguelmota:
Is this still an issue? I was not able to reproduce a few weeks ago

shanefontaine:
Yes it is. I was just able to reproduce it on the latest ETH logstream by adding `--dry` to the end of `docker.sh` and running `./docker.sh`.

miguelmota:
@shanefontaine the dry mode state is being overridden [here](https://github.com/hop-protocol/hop/blob/1f491b0f1b0801674032c803e03fe0234db78c00/packages/hop-node/src/watchers/classes/BaseWatcher.ts#L207) which is what's causing it to keep bonding withdrawals

[...]

PR #385, opened 2022-01-26
hop-node: BaseWatcher: Don't override dry mode status if it was started with dry mode. #330

miguelmota:
- See issue [comment](https://github.com/hop-protocol/hop/issues/330#issuecomment-1021738333)


shanefontaine:
Per our discussion, I believe this PR will encompass the removal of state update contract logic. If so, I will review at that time. If not, then feel free to merge this and open a new PR with the removal of the state update contract logic.

Issue #340, opened 2021-12-17
In-memory timestamp checks are not ideal and should be removed in order to keep watcher standardized


shanefontaine:
Some watchers have in-memory timestamp checks. These are not ideal, as they (1) are unstandard relative to the rest of the watchers, (2) might not work as expected on bonder startup (before memory is written), and (3) are difficult to debug.

These _should be able_ to be removed, however, the updated implementation is not 100% clear to me.

Two cases where this matters:
* [`settleAttemptedAt`](https://github.com/hop-protocol/hop/blob/f10a3234399581b021239e4fbfb39c135f8bc50e/packages/hop-node/src/watchers/SettleBondedWithdrawalWatcher.ts#L24)
* [`commitTxSentAt`](https://github.com/hop-protocol/hop/blob/f10a3234399581b021239e4fbfb39c135f8bc50e/packages/hop-node/src/watchers/CommitTransfersWatcher.ts#L27)

miguelmota:
I think they could be moved to be a db item property to be consistent with the rest (like `bondWithdrawalAttemptedAt`) and would make debugging easier

shanefontaine:
That might work for `commitTxSentAt` (I'm not 100% sure if there is a reason why it is not like this in the first place).

I'm not sure that works for `settleAttemptedAt`. `settleAttemptedAt` currently serves the role of `"Mark a settlement as attempted here so that multiple db reads are not attempted every poll. This comes into play when a transfer is bonded after others in the same root have been settled"`. Maybe that can be done in a more ideal way rather than an in-memory store.

I think this one requires some more thought through edge cases if/when it gets implemented.

miguelmota:
I think the reason `commitTxSentAt` is in memory is because we don't know what the transferRootId is yet (we know what it is only after commitTransfers), so for that reason we can't update it in the db and I think we should leave it as is.

The settleAttemptedAt state has been moved to the transferRootsDb ([PR](https://github.com/hop-protocol/hop/pull/386)). I think it was in memory because in the past we were indexing by transferRootHash and not by transferRootId and the db getters weren't as efficient.

miguelmota:
Closed by https://github.com/hop-protocol/hop/pull/386

PR #386, opened 2022-01-26
hop-node: Move settleAttemptedAt to TransferRootsDb. #340

miguelmota:
- See issue [comment](https://github.com/hop-protocol/hop/issues/340#issuecomment-1021762046)


packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -574,6 +575,11 @@ class TransferRootsDb extends BaseDb {
           Date.now()
       }
 
+      let settleAttemptTimestampOk = true
+      if (item.settleAttemptedAt) {
+        settleAttemptTimestampOk = item.settleAttemptedAt + OneHourMs < Date.now()
```
shanefontaine, line 580: We should use `TxRetryDelayMs` instead of `OneHourMs`. We should have probably been doing that this whole time.

packages/hop-node/src/watchers/SettleBondedWithdrawalWatcher.ts
```diff
@@ -45,16 +44,11 @@ class SettleBondedWithdrawalWatcher extends BaseWatcher {
...
-      // This comes into play when a transfer is bonded after others in the same root have been settled
-      if (!this.settleAttemptedAt[transferRootId]) {
-        this.settleAttemptedAt[transferRootId] = 0
-      }
-      const timestampOk = this.settleAttemptedAt[transferRootId] + OneHourMs < Date.now()
-      if (!timestampOk) {
-        continue
-      }
-      this.settleAttemptedAt[transferRootId] = Date.now()
+
```
shanefontaine, line 47: I think it would be useful to keep the two lines of comments in here, as they're useful context even after this update has been implemented.

shanefontaine (commented):
I think it makes sense to add code comments above the `commitTxSentAt` logic that describes what you said [here](https://github.com/hop-protocol/hop/issues/340#issuecomment-1021762046). I agree with your comment and it would be nice to have that context in the code.

packages/hop-node/src/watchers/SettleBondedWithdrawalWatcher.ts
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/386/commits/2486709139caafa02c59530489d107b6965ccba2

packages/hop-node/src/db/TransferRootsDb.ts
miguelmota: Updated

Resolved with https://github.com/hop-protocol/hop/pull/386/commits/2486709139caafa02c59530489d107b6965ccba2

PR #388, opened 2022-01-27
core/sdk/ui: L1 canonical bridge option

kangarang:
this pr adds support for sending assets via L2-native L1 canonical bridges

Screen Shot 2022-02-05 at 1 31 22 AM


shanefontaine (changes requested):
Nice! This is going to be a solid feature.

* It looks like the Arbitrum bridge that the tx is sent to is not the actual address that gets approved. See attached image. It looks like the approval address for sending DAI over Arbitrum is `0xd3b5b60020504bc3489d6949d545893982ba3011`, and I don't see that in the repo anywhere. I did not look at any other bridges, but it might be worth confirming that the approval addresses are accounted for if they differ from the actual bridge address that is interacted with

Screen Shot 2022-02-05 at 14 42 27

* I thought we landed on keeping the canonical bridge logic out of the SDK and keeping the SDK pure with Hop-only logic. I see that the code is pretty closely coupled, so I'd imagine that might be difficult. Do you remember where we ended up landing on that last time we discussed this?
  * Similar comment but WRT the canonical bridge addresses in `core/src/addresses/`

* I see a hardCoded `gasLimit` of `200e3` [here](https://github.com/hop-protocol/hop/pull/388/files#diff-8d5518358c199a7acaf429850259ed0b559b724895d70fca003708ab0479d3cdR101) (and maybe other places):
  * I wouldn't think we would ever need to hard code a `gasLimit`. If that is true, can it be removed?
  * Arbitrum has a different concept of gas than the other chains. Because of this their gasLimit values are much, much higher — higher than `200e3` for most or all txs. I’d imagine the `200e3` might not work for Arbiturm. See these equivalent txs gas costs on [Optimism](https://optimistic.etherscan.io/tx/0x95378c74e0af381364ec32508b7cfca8bbb5572dc8d4e2c4b5df3e7e42cd12d4 ) and [Arbitrum](https://arbiscan.io/tx/0x9d4a0885dd5d8c21909a934b3b2ea98949c577a42e24319c3b5fafbd455a8ac0).
  * I will caveat this comment with the knowledge that there are Alchemy bugs on both Mainnet and Polygon WRT gas limit stuff

packages/frontend/src/pages/Send/useL1CanonicalBridge.ts
```diff
@@ -0,0 +1,103 @@
...
+  const [l1CanonicalBridge, setL1CanonicalBridge] = useState<CanonicalBridge | undefined>()
+  const [usingL1CanonicalBridge, setUl1cb] = useLocalStorage('using-l1-canonical-bridge', false)
+
+  useEffect(() => {
+    if (sourceTokenAmount && estimatedReceived && l1CanonicalBridge) {
+      if (usingL1CanonicalBridge == null && sourceTokenAmount.gt(estimatedReceived)) {
+        return setUl1cb(true)
+      }
+
+      if (sourceTokenAmount.lte(estimatedReceived)) {
```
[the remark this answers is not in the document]
kangarang: yeah, the logic in this hook ignores that case and only changes to the hop bridge if the price is better. otherwise we proceed with the user's choice

packages/sdk/src/CanonicalBridge.ts
```diff
@@ -106,10 +128,27 @@ class CanonicalBridge extends Base {
...
+    }
+    return l1CanonicalToken.allowance(spender)
+  }
+
+  public getDepositApprovalAddress (chain: TChain = this.chain): string {
+    chain = this.toChainModel(chain)
     let spender = this.getL1CanonicalBridgeAddress(this.tokenSymbol, chain)
-    if (chain.equals(Chain.Polygon)) {
+    if (chain.equals(Chain.Polygon) && this.tokenSymbol !== Token.MATIC) {
       spender = this.getL1PosErc20PredicateAddress(this.tokenSymbol, chain)
```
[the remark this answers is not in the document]
kangarang: let me know if i'm mistaken, but if chain is polygon and token is not MATIC, `getL1PosErc20PredicateAddress` should return a different address than `getL1CanonicalBridgeAddress`

USDC / USDT / DAI:

```
l1CanonicalBridge: '0xA0c68C638235ee32657e8f720a23ceC1bFc77C77'
l1PosPredicate: '0x40ec5B33f54e0E8A33A975908C5BA1c14e5BbbDf'
```

ETH:

```
l1CanonicalBridge: '0xA0c68C638235ee32657e8f720a23ceC1bFc77C77'
l1PosPredicate: '0x8484Ef722627bf18ca5Ae6BcF031c23E6e922B30'
```

for MATIC we approve the canonical Plasma bridge (`0x401F6c983eA34274ec46f84D70b31C151321188b`)

packages/sdk/src/CanonicalBridge.ts
```diff
@@ -139,7 +173,19 @@ class CanonicalBridge extends Base {
         `token "${this.tokenSymbol}" on chain "${chain.slug}" is unsupported`
       )
     }
-    return l1CanonicalToken.approve(spender, amount)
+    return l1CanonicalToken.approve(spender, ethers.constants.MaxUint256)
```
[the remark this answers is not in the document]
kangarang: normally, the app triggers a tx confirm modal which defaults to `MaxUint256`. if it's not too much work to follow the same pattern here, i'll change it back to `amount`. in my opinion, allowance txs should default to the max. metamask also allows the user to change the allowance value via `Edit Permission`.

packages/sdk/src/CanonicalBridge.ts
```diff
@@ -159,87 +217,135 @@ class CanonicalBridge extends Base {
...
-        this.tokenSymbol,
-        chain
+    if (chain.equals(Chain.Gnosis)) {
+      if (this.tokenSymbol === Token.DAI) {
+        return (l1CanonicalBridge as L1XDaiPoaBridge).populateTransaction.relayTokens(
+          signerAddress,
+          amount,
+          {
+            // Gnosis requires a higher gas limit
+            gasLimit: 500e3,
```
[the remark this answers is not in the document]
kangarang: yeahh, most of this is legacy, and i'm unsure if they are necessary anymore. if you have any insight on this, i'll defer to it!

packages/sdk/src/CanonicalBridge.ts
```diff
@@ -159,87 +217,135 @@ class CanonicalBridge extends Base {
...
+      if (this.tokenSymbol === Token.ETH) {
+        bridge = await this.getContract(ArbitrumInbox__factory, l1CanonicalBridge.address, provider)
+        return (bridge as ArbitrumInbox).populateTransaction.depositEth(payload, {
+          value: amount,
+          from,
+        })
+      }
+
+      const maxGas = 200e3
+      const gasPriceBid = '100'
```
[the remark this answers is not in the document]
kangarang: i think these values were obtained from a tx that i populated via the arbitrum bridge. i'm unsure if they are variable. in any case, i'm going to change the values to use the ones found in [this example](https://github.com/OffchainLabs/arbitrum/blob/develop/packages/arb-ts/integration_test/arb-bridge.test.ts#L455-L462), as this is what is recommended by the arb-ts package

kangarang:
@shanefontaine 

> It looks like the Arbitrum bridge that the tx is sent to is not the actual address that gets approved. See attached image. It looks like the approval address for sending DAI over Arbitrum is 0xd3b5b60020504bc3489d6949d545893982ba3011, and I don't see that in the repo anywhere. I did not look at any other bridges, but it might be worth confirming that the approval addresses are accounted for if they differ from the actual bridge address that is interacted with

ahh, that makes a lot of sense! i was wondering how arbitrum handled approvals, since i had not approved the `0x72Ce9c` address. thanks for catching this.

> I thought we landed on keeping the canonical bridge logic out of the SDK and keeping the SDK pure with Hop-only logic. I see that the code is pretty closely coupled, so I'd imagine that might be difficult. Do you remember where we ended up landing on that last time we discussed this?

yeaa, i believe we all agreed that it'd be best to localize the L1 canonical bridge in the frontend. i mentioned that i wanted to get it working first, and then clean up the code after. thanks for the reminder -- i think it'll be a quick refactor

packages/sdk/src/CanonicalBridge.ts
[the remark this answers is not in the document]
shanefontaine: That sounds right, but I am not 100% certain, as I'm not sure what the exact naming convention is for their bridge/token contracts. If you tested it successfully, then I believe it is good to go 👍

packages/sdk/src/CanonicalBridge.ts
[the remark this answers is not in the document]
shanefontaine: Have you tested a transaction with these values on mainnet? I know their tests have these values, but their L2 recipient or alias address might be pre-funded for these tests, which means it would work in the test but not in prod.

From a top level perspective, the user needs to pay for the bridging of the asset when using the Arbitrum bridge. This could come in the form of having funds pre-deposited on an address in L2 OR sending ETH with the L1 tx. In this case, if the `gasPriceBid` is `0`, then I think we are telling the system that we expect the funds to already exist on L2 to be paid, which I do not believe will be the case for Hop users.

It would be best to test this on the actual Arbitrum bridge itself and trace the tx and see what happens, as some of my info above might be outdated or misinformed.

Happy to call and talk through this if that is easier.

packages/sdk/src/HopBridge.ts
```diff
@@ -428,10 +428,8 @@ class HopBridge extends Base {
       // a `from` address is required if using only provider (not signer)
       populatedTx.from = await this.getGasEstimateFromAddress(sourceChain, destinationChain)
     }
-    return sourceChain.provider.estimateGas({
-      ...populatedTx,
-      gasLimit: 500000
-    })
+
+    return sourceChain.provider.estimateGas(populatedTx)
```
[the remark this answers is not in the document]
shanefontaine: Any insights here?

shanefontaine:
@kangarang back to you. Changes look good! Left some additional comments.

It is my understanding that we will localize the addresses to the FE and effectively make no change to the `core` package before merging this PR, so I'll give one more review after that is done. Happy to talk through it if that is not what you're envisioning.

packages/frontend/src/contexts/Web3Context.tsx
```diff
@@ -270,9 +269,13 @@ const Web3ContextProvider: FC = ({ children }) => {
 
   // TODO: cleanup
   const checkConnectedNetworkId = useCallback(
-    async (networkId?: number): Promise<boolean> => {
+    async (networkId?: number | string): Promise<boolean> => {
```
miguelmota, line 272: in what context or scenario will networkId be a `string`?

packages/frontend/src/pages/Send/L1CanonicalBridgeOption.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { RadioButtonChecked, RadioButtonUnchecked } from '@material-ui/icons'
+
+interface Props {
+  amount?: BigNumber
+  destNetwork?: Network
+  destToken?: Token
+  estimatedReceivedDisplay?: string
+  l1CanonicalBridge?: CanonicalBridge
+  sendL1CanonicalBridge?: () => void
+  setUl1cb: (u: boolean) => void
```
miguelmota, line 17: this callback name `setUl1cb` isn't clear what it's for, the name can be more descriptive

packages/core/build/addresses/mainnet.json
```diff
@@ -34,34 +34,34 @@
     },
```
miguelmota, line 34: this file shouldn't be removed because [defillama](https://github.com/DefiLlama/DefiLlama-Adapters/blob/c091de607442c1430b00770c77ad2f737d6e0ed4/projects/hop/index.js#L18) relies on it (sorry it wasn't clear)

miguelmota:
A few observations

expected: Select bridge with better rate after inputting amount
actual: Native bridge was not chosen when it had a better rate (maybe a debounce here is needed?)

Peek 2022-02-16 15-52

expected: Disconnect wallet without error
actual: App crashed when disconnecting wallet on UI (looks like a hook somewhere needs to be updated on provider or address change) 

Peek 2022-02-16 15-55

miguelmota:
@kangarang let me know if you have any questions on the comments

packages/core/build/addresses/mainnet.json
kangarang: ahh, i see. are the changes here okay though?

packages/frontend/src/pages/Send/L1CanonicalBridgeOption.tsx
kangarang: i'll go with `setUsingNativeBridge`

packages/sdk/src/CanonicalBridge.ts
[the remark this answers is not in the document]
kangarang: wow, it really varies. here's two successful deposits using the arbitrum bridge ui:

Screen Shot 2022-03-07 at 9 40 25 AM

Screen Shot 2022-03-07 at 9 40 22 AM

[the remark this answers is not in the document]
kangarang: it may be best for us to integrate `arb-ts` (https://developer.offchainlabs.com/docs/frontend_integration#arbitrum--ethereum-bridge), as they suggest in their docs. it seems pretty straightforward

kangarang:
@shanefontaine 

> Similar comment but WRT the canonical bridge addresses in core/src/addresses/

for now, let's keep the addresses in `core/src/addresses`, as the models helpers and each type relies on that single source of truth

@miguelmota 

> expected: Select bridge with better rate after inputting amount
> actual: Native bridge was not chosen when it had a better rate (maybe a debounce here is needed?)

i considered this as well, but i decided on this approach because i believe my implementation gives users more agency and less confusion.

the logic gates are: if the user has not selected either hop bridge or native bridge, the app will choose for them (whichever is cheaper). if the user explicitly chooses a particular bridge, their selection is stored in local storage and is used as the default going forward. the only other time that the app will switch bridges on behalf of the user is if the user has defaulted to native bridge, but hop's bridge is cheaper. the main benefit of this approach is that it minimizes the potential for auto-switching bridges and confusing the user. this would especially be confusing if that happened after pressing `send`.

another option is to set the bridge option at the beginning of the transaction, and prevent switching if any subsequently polled data suggest a different rate. thoughts?

packages/sdk/src/HopBridge.ts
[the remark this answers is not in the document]
kangarang: we do have a util that adds some buffer gas that we can use, but it's just a hard coded number. i feel like a percentage would indeed be more realistic and reliable.

packages/frontend/src/contexts/Web3Context.tsx
kangarang: i see in many places that we convert to number just prior to calling the method. this change adds some flexibility.

Screen Shot 2022-03-07 at 9 54 17 AM

miguelmota:
@kangarang that makes sense. the problem is that it's not clear that it'll continue using the option that was previously selected. I think there needs to be an optional setting (maybe under the cog settings) to always use the cheapest option regardless of what was previously chosen (or a way to 'reset' this state)

shanefontaine:
@kangarang 

> their selection is stored in local storage and is used as the default going forward

I might be misunderstanding, but I was under the impression we would always default to the cheapest option. Is that the case? _If_ the user wanted to use the more expensive option for some reason, then they'd be able to do so on a per-tx basis (not persisted in local storage).

I might be misunderstanding or not have the correct context, so let me know!

> another option is to set the bridge option at the beginning of the transaction, and prevent switching if any subsequently polled data suggest a different rate. thoughts?

I think this makes sense in general to avoid any race conditions.

kangarang:
@shanefontaine @miguelmota totally understand your points of views. i think a good compromise will be a one-time-only auto-bridge option set by the app (based on price) immediately after getting send data (could happen multiple times, but the app will not auto-switch beyond that point, even if a subsequent poll to get send data suggests otherwise.

kangarang:
working as expected:

https://user-images.githubusercontent.com/18040654/157194487-c7b9b2be-e0aa-445a-a022-7b3c0a8b75db.mp4

kangarang:
some notes:

bridge instantiation with `arb-ts` requires both L1 and L2 signers. due to the way our sdk configures signers/providers, injecting an Arbitrum `StaticJsonRpcProvider` prevented the bridge from performing any methods. super hard to figure out what was goin on, as the error message was the generic exception thrown by `ethers.utils.getAddress`. an actual signer was needed (i.e. connected to an address). this is resolved by instantiating an arbitrum L2 signer via `provider.getUncheckedSigner(address)`.

[this was the most complete example](https://github.com/OffchainLabs/arbitrum/blob/master/packages/arb-ts/scripts/deployStandard.ts) of `arb-ts`.

not sure if we should also be [implementing retry-ability](https://github.com/OffchainLabs/arbitrum/blob/98d33a5e92de47de97aec857c7fd92eb63db543e/packages/arb-ts/scripts/deployStandard.ts#L137)

noticed [lots of helpers](https://github.com/OffchainLabs/arbitrum/blob/master/packages/arb-ts/src/lib/bridge_helpers.ts) in `arb-ts`

packages/frontend/src/hooks/useEstimateTxCost.ts
```diff
@@ -81,62 +94,67 @@ export function useEstimateTxCost(selectedNetwork?: Network) {
...
-          let gasCost = await estimateGasCost(network, estimatedGasLimit)
-          if (gasCost && network.slug === ChainSlug.Optimism) {
-            const { gasLimit, data, to } = await token.getWrapTokenEstimatedGas(network.slug)
-            const l1FeeInWei = await token.estimateOptimismL1FeeFromData(gasLimit, data, to)
-            gasCost = gasCost.add(l1FeeInWei)
-          }
-          return gasCost
+  const estimateWrap = useCallback(async (options: EstimateTxOptions) => {
+    const { token, network } = options
+    if (!network) {
```
miguelmota: do we need to check that `token` exists here as well?

packages/frontend/src/hooks/useEstimateTxCost.ts
```diff
@@ -81,62 +94,67 @@ export function useEstimateTxCost(selectedNetwork?: Network) {
...
+      return
+    }
+
+    try {
+      // Get estimated gas cost
+      const estimatedGasLimit = await token.wrapToken(BigNumber.from(10), true)
```
miguelmota: suggestion: the `true` parameter should be a variable for clarity: `const estimateOnly = true`

packages/frontend/src/models/CanonicalBridge.ts
```diff
@@ -0,0 +1,341 @@
...
+
+class CanonicalBridge extends Base {
+  public chain: Chain
+  public tokenSymbol: TokenSymbol
+  address: string
+  l2TokenAddress: string
+
+  constructor(network: string, signer: Signer, token: TToken, chain: TChain) {
+    super(network, signer)
+    if (!token) {
```
miguelmota: should all the parameters be checked here that they exists?

packages/frontend/src/models/CanonicalBridge.ts
```diff
@@ -0,0 +1,341 @@
...
+
+  public async deposit(amount: BigNumberish, customRecipient?: string) {
+    if (this.chain.equals(Chain.Arbitrum) && this.tokenSymbol !== CanonicalToken.ETH) {
+      return this.populateDepositTx(amount, customRecipient)
+    }
+    const populatedTx = await this.populateDepositTx(amount)
+    return this.signer.sendTransaction(populatedTx)
+  }
+
+  public async populateDepositTx(amount: BigNumberish, customRecipient?: string): Promise<any> {
```
miguelmota, line 143: suggestion: instead of multiple if statements checking chain and token, using polymorphism would be cleaner ([example](https://github.com/hop-protocol/hop/blob/develop/packages/sdk/src/watchers/Watcher.ts))

packages/frontend/src/pages/Convert/ConvertOption/HopConvertOption.tsx
```diff
@@ -68,11 +67,12 @@ class HopConvertOption extends ConvertOption {
...
-    const {
-      totalFee,
-      adjustedBonderFee,
-      adjustedDestinationTxFee,
-    } = await bridge.getSendData(amountIn, sourceNetwork.slug, destNetwork.slug, true)
+    const { totalFee, adjustedBonderFee, adjustedDestinationTxFee } = await bridge.getSendData(
+      amountIn,
+      sourceNetwork.slug,
+      destNetwork.slug,
+      true
```
miguelmota: suggestion: move boolean parameters to variables for clarity

packages/sdk/src/Base.ts
```diff
@@ -373,36 +388,27 @@ class Base {
...
+
     if (Signer.isSigner(signer)) {
       if (signer.provider) {
         const connectedChainId = await signer.getChainId()
-        // console.log('connectedChainId: ', connectedChainId)
-        // console.log('chain.chainId: ', chain.chainId)
         if (connectedChainId !== chain.chainId) {
-          if (!signer.provider) {
-            // console.log('connect provider')
-            return (signer as Signer).connect(chain.provider)
```
miguelmota, line 392: curious, why was this line removed?

packages/frontend/src/pages/Send/useL1CanonicalBridge.ts
```diff
@@ -0,0 +1,136 @@
...
+import { BigNumber, constants } from 'ethers'
+import Network from 'src/models/Network'
+import { useWeb3Context } from 'src/contexts/Web3Context'
+import logger from 'src/logger'
+import { findNetworkById, formatError, toTokenDisplay } from 'src/utils'
+import CanonicalBridge from 'src/models/CanonicalBridge'
+import { handleTransaction } from './useSendTransaction'
+import { useTransactionReplacement } from 'src/hooks'
+
+async function needsApproval(l1CanonicalBridge: CanonicalBridge, sourceToken, sourceTokenAmount) {
```
miguelmota: the first parameter has a type but the other two do not. I think they should all have types

packages/frontend/src/pages/Send/useL1CanonicalBridge.ts
```diff
@@ -0,0 +1,136 @@
...
+  }, [provider, sourceTokenAmount?.toString(), sourceToken?.chain.chainId, destNetwork?.slug])
+
+  async function sendL1CanonicalBridge() {
+    if (!(l1CanonicalBridge && sourceTokenAmount)) {
+      return
+    }
+
+    const sourceNetwork = findNetworkById(sourceToken?.chain.chainId.toString()!)
+
+    if (await needsApproval(l1CanonicalBridge, sourceToken, sourceTokenAmount)) {
```
miguelmota: for best practice, the `await` should be moved out of the condition into a var (ie `shouldApprove`)

packages/frontend/src/pages/Send/useL1CanonicalBridge.ts
```diff
@@ -0,0 +1,136 @@
...
+        },
+        estimatedReceived: toTokenDisplay(
+          estimatedReceived,
+          sourceToken?.decimals,
+          sourceToken?.symbol
+        ),
+      },
+      onConfirm: async () => {
+        try {
+          const isNetworkConnected = await checkConnectedNetworkId(1)
```
miguelmota: we should avoid using hard coded chain ids since it will make it harder to make it work with a testnet. suggestion: use the util to get L1 network id since that reads the network ids from config

packages/frontend/src/pages/Send/useSendData.tsx
```diff
@@ -31,7 +31,7 @@ const useSendData = (
     {
       enabled:
         !!token?.address && !!fromNetwork?.slug && !!toNetwork?.slug && !!fromAmount?.toString(),
-      refetchInterval: 5e3,
+      refetchInterval: 10e3,
```
miguelmota: it looks like the intervals vary in components; ie `5`, `7`, or `10` seconds. maybe this can be more consistent throughout

packages/frontend/src/models/CanonicalBridge.ts
```diff
@@ -0,0 +1,341 @@
...
+  L1XDaiForeignOmniBridge__factory,
+  L1XDaiForeignOmniBridge,
+  L1XDaiPoaBridge__factory,
+  L1XDaiPoaBridge,
+  L1XDaiWETHOmnibridgeRouter__factory,
+  L1XDaiWETHOmnibridgeRouter,
+} from '@hop-protocol/core/contracts'
+import { ChainSlug, CanonicalToken } from '@hop-protocol/sdk'
+import Base, { L1Factory } from '@hop-protocol/sdk/dist/src/Base'
+import { metadata } from '@hop-protocol/sdk/dist/src/config'
```
miguelmota: I don't think it should be importing from `dist` like this `@hop-protocol/sdk/dist/src/config` because that makes it fragile. I suggest importing from `@hop-protocol/sdk` or structuring the typescript build so the output dirs are more structured like `@hop-protocol/sdk/config`

packages/frontend/src/models/CanonicalBridge.ts
```diff
@@ -0,0 +1,341 @@
...
+    this.address = this.getL1CanonicalBridgeAddress(this.tokenSymbol, this.chain)
+    this.l2TokenAddress = this.getL2CanonicalTokenAddress(this.tokenSymbol, this.chain)
+  }
+
+  public connect(signer: Signer) {
+    return new CanonicalBridge(this.network, signer, this.tokenSymbol, this.chain)
+  }
+
+  public getL1CanonicalAllowance(): Promise<BigNumber> {
+    const l1CanonicalToken = this.getL1Token().connect(this.signer)
```
miguelmota: suggestion: move `.connect(this.signer)` be to `getL1Token` method to avoid having to manually connect every time `getL1Token` is called

packages/frontend/src/models/CanonicalBridge.ts
```diff
@@ -0,0 +1,341 @@
...
+        if (this.tokenSymbol === Token.DAI) {
+          const bridge = L1OptimismDaiTokenBridge__factory.connect(
+            l1CanonicalBridge.address,
+            this.signer
+          )
+          return bridge.populateTransaction.depositERC20(
+            l1CanonicalToken.address,
+            l2TokenAddress,
+            amount,
+            BigNumber.from('1920000'),
```
miguelmota: this should be moved to a variable to know what it means

packages/frontend/src/models/CanonicalBridge.ts
```diff
@@ -0,0 +1,341 @@
...
+            '0x',
```
miguelmota: this should be moved to a variable to know what it means

packages/frontend/src/models/CanonicalBridge.ts
```diff
@@ -0,0 +1,341 @@
...
+        const { maxFeePerGas, gasPrice } = await arbBridge.l1Provider.getFeeData()
+        const price = maxFeePerGas || gasPrice
+        const walletBal = await arbBridge.l1Provider.getBalance(from)
+
+        if (!price) {
+          console.log('Warning: could not get gas price estimate; will try depositing anyway')
+        } else {
+          const fee = price.mul(gasNeeded)
+          if (fee.gt(walletBal)) {
+            console.log(
```
miguelmota: does metamask throw error in this case?

packages/frontend/src/contexts/Web3Context.tsx
miguelmota: the `Number` wrap here is legacy code. the config `networkId` should always be a number ([type](https://github.com/hop-protocol/hop/blob/94b2570eebe7c71da7ad9d8f657d5daf0aa8dc8d/packages/core/src/networks/types.ts#L3)) and `provider.getNetwork()` will always return `number` so i don't think there's need to have string type

packages/sdk/src/Base.ts
kangarang: this block of code shouldn't ever run because it is within the condition set by line 382

kangarang (commented):
just a couple more edits and i think we should be good to merge. thanks for the awesome code reviews

packages/frontend/src/hooks/useEstimateTxCost.ts
kangarang: i did not realize we had a method `getWrapTokenEstimatedGas` for this already

packages/sdk/src/Base.ts
miguelmota: good catch
i think we should still connect `chain.provider` if `signer.provider` doesn't exist since it's possible to use in a signer without a provider and let the sdk handle connecting provider

packages/frontend/src/contexts/Web3Context.tsx
kangarang: i see what's happening. actually, `sourceNetwork` in this particular usage is the `Network` model ([type](https://github.com/hop-protocol/hop/blob/develop/packages/frontend/src/models/Network.ts#L25)), which converts the original `networkId.toString()` ([here](https://github.com/hop-protocol/hop/blob/develop/packages/frontend/src/config/mainnet.ts#L27)). i just tested removing `.toString()` in that file, and it doesn't seem to break the app, as `Number` accepts `any`.

packages/sdk/src/Base.ts
kangarang: yes, thank you, very thoughtful. on line [409](https://github.com/hop-protocol/hop/blob/796bdf3d106b521579f998a86a7fcd718bdc426c/packages/sdk/src/Base.ts#L409), we connect `chain.provider` as a fallback

miguelmota: This should return the signer with connected provider instead of just provider, otherwise transaction won't be able to be made with just the provider

```typescript
return signer.connect(chain.provider)
```

packages/core/build/addresses/mainnet.json
miguelmota: yes

packages/sdk/src/Base.ts
kangarang: ahhhh, i see. thanks for clarifying

packages/frontend/src/contexts/Web3Context.tsx
miguelmota: nice, removing the string types is much cleaner

miguelmota:
Code updates look good 👍 

can we deploy to a staging site (mainnet2) for live testing?

packages/frontend/src/config/nativeTokenBridges.ts
```diff
@@ -0,0 +1,36 @@
...
+  L1ArbitrumDaiGateway__factory,
+  L1OptimismDaiTokenBridge__factory,
+  L1OptimismTokenBridge__factory,
+  L1PolygonPlasmaBridgeDepositManager__factory,
+  L1PolygonPosRootChainManager__factory,
+  L1XDaiForeignOmniBridge__factory,
+  L1XDaiWETHOmnibridgeRouter__factory,
+} from '@hop-protocol/core/contracts'
+
+export const nativeTokenBridges = {
```
kangarang, line 13: this isn't being used atm, but will be useful in a subsequent refactor

packages/frontend/src/models/CanonicalBridge.ts
kangarang: that is a good question. i haven't run into this issue yet, and this is copy/pasta from an [arb-ts example](https://github.com/OffchainLabs/arbitrum/blob/338fa2da63c290faeaa27b848f1d93d60038aa09/packages/arb-ts/scripts/deployStandard.ts#L121)

kangarang:
i've tested each native bridge 3 times: (1) ETH, (2) native asset, and (3) DAI. where 2 of the 3 are the same, I sent USDC (i.e. not polygon).

the final few commits includes fixes where [such transfers failed](https://etherscan.io/address/0x9997da3de3ec197c853bcc96caecf08a81de9d69)

miguelmota:
minor bug 

expected: when wallet is not connected, send button should prompt to connect wallet or be disabled with message to connect wallet
actual: send button is active but nothing happens

DeepinScreenshot_select-area_20220325093046

pools page send button for reference

DeepinScreenshot_select-area_20220325093230

miguelmota:
bug

expected: ability to approve or send tokens, or show message why buttons are disabled
actual: all action buttons are disabled

approve and send buttons also appear to be different sizes

DeepinScreenshot_select-area_20220325093455

recording

Peek 2022-03-25 09-37

miguelmota:
bug

expected: show warning and disable button if account doesn't have enough balance to send
actual: no warning is shown and send button is enabled but transaction fails leaving ui in a stuck state

DeepinScreenshot_select-area_20220325093912

recording

Peek 2022-03-25 09-39

miguelmota:
bug

expected: ability to approve or send tokens, or show message why buttons are disabled
actual: all action buttons are disabled without a reason message

Peek 2022-04-11 11-43

kangarang:
@miguelmota, thank you for your thorough bug reports. i have resolved these issues in my latest changes locally, but i need to do some further testing.

Screen Shot 2022-04-12 at 12 43 36 PM

will push fixes shortly this afternoon

kangarang:
Known bug: other routes than `/send` are no longer working as expected. seems to have something to do with the `txConfirm` changes

shanefontaine:
Replaced with #435

PR #389, opened 2022-01-27
hop-node: Add 10k buffer to gas limit for bondWithdrawal to L1

miguelmota:
- adds 10k buffer to bondWithdrawal to L1

(not sure why it's including other prev commits from yesterday in this PR when i merged develop)


shanefontaine (approved):
@miguelmota LGTM. I think it makes sense to get this one merged today and pushed to the bonders, given how simple it is 👍

PR #391, opened 2022-01-28
sdk/ui: fix network switching and sufficient balance hook

kangarang:
previously, if a user manually switched networks prior to confirming triggering a send tx, the app would use an old `provider` instance in `checkConnectedNetworkId()`. this is now addressed by recreating the function via `useCallback` which depends on `provider`.

previously, if a user did not have enough native token to pay for tx fees, the tx buttons would not be enabled and the `useSufficientBalance` hook would not display a warning message to indicate the error. this was due to the `useSufficientBalance` hook depending on the value `estimatedGasCost`. however, because the user doesn't have enough gas, `estimatedGasCost` would be undefined because of failed tx. this is now addressed by setting a default estimated gas cost.

additionally, a new field `nativeTokenSymbol` is added to the chain metadata for ease of use

### DIFF
--- a/packages/frontend/src/hooks/useSufficientBalance.ts
+++ b/packages/frontend/src/hooks/useSufficientBalance.ts
@@ -6,7 +6,7 @@ import { toTokenDisplay } from 'src/utils'
 export function useSufficientBalance(
   token?: Token,
   amount?: BigNumber,
-  estimatedGasCost?: BigNumber,
+  estimatedGasCost: BigNumber = BigNumber.from(500_000),
   tokenBalance: BigNumber = BigNumber.from(0)
 ) {
   const [sufficientBalance, setSufficientBalance] = useState(false)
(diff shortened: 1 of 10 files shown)


packages/frontend/src/hooks/useSufficientBalance.ts
```diff
@@ -6,15 +6,15 @@ import { toTokenDisplay } from 'src/utils'
 export function useSufficientBalance(
   token?: Token,
   amount?: BigNumber,
-  estimatedGasCost?: BigNumber,
+  estimatedGasCost: BigNumber = BigNumber.from(500_000),
```
miguelmota, line 9: Is the default estimatedGasCost `500_000` wei (`0.0000000000005` ETH) correct? seems kind of low

kangarang: thanks for the catch! i must have confused this with gas limit

miguelmota (approved):
lgtm

sdk will require a bump before pushing live

PR #393, opened 2022-02-04
UI: fix network switching requests

kangarang:
previously, switching between networks would incorrectly return `true` if a user cancelled the `wallet_addEthereumChain` request (made by the app). this pr adds an additional double-check which ensures that `checkConnectedNetworkId` will *only* return true if the provider network matches the network id for the transaction.

### DIFF
--- a/packages/frontend/src/contexts/Web3Context.tsx
+++ b/packages/frontend/src/contexts/Web3Context.tsx
@@ -1,4 +1,12 @@
-import React, { FC, createContext, useContext, useMemo, useState, useEffect, useCallback } from 'react'
+import React, {
+  FC,
+  createContext,
+  useContext,
+  useMemo,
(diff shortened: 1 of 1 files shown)


packages/frontend/src/contexts/Web3Context.tsx
```diff
@@ -261,64 +269,73 @@ const Web3ContextProvider: FC = ({ children }) => {
...
         }
-
-        await onboard.walletCheck()
       }
-      return false
-    }
+      const p = await provider.getNetwork()
+      if (p.chainId === networkId) {
+        return true
+      }
```
kangarang, line 331: previously, this function would `return true` simply after sending the `wallet_addEthereumChain`, regardless of what the user selected. now we double-check to make sure the network matches.

PR #394, opened 2022-02-09
UI: fix staking data calculations

kangarang:
previously, the ui would run into issues calculating staking apr and total rewards if a user switched between networks. this was caused by the contracts being set asynchronously, intermittently causing the subsequent hooks in `StakeWidget` that listed the contracts as dependencies to calculate stake data values incorrectly. this is now addressed by synchronously setting contracts. additionally, the `useBalance` hook is simplified, some calculations are grouped together, and less requests are made to fetch `rewardTokenUsdPrice`.


miguelmota:
lgtm! works as expected

miguelmota (approved):

PR #395, opened 2022-02-16
Add support for Gnosis-Safe as a safe-app.

nicosampler:
How to test it:

1. Go to any gnosis safe. Make sure that your wallet is in the same network as your safe. Also, your wallet has to be using an account that is the owner of the safe. 
2. go to `Apps` section
3. add custom app. for testing purposes, you can use this link [hop-nicosampler.vercel.app](https://hop-5ro04hq46-nicosampler.vercel.app)
5. enjoy 🎉

### DIFF
--- a/packages/frontend/src/hooks/useIsSmartContractWallet.ts
+++ b/packages/frontend/src/hooks/useIsSmartContractWallet.ts
@@ -1,20 +1,20 @@
-import { useEffect, useState } from "react";
-import { useWeb3Context } from "src/contexts/Web3Context";
+import { useEffect, useState } from 'react'
+import { useWeb3Context } from 'src/contexts/Web3Context'
 
 export default function useIsSmartContractWallet() {
-  const { provider, address } = useWeb3Context();
-  const [isSmartContractWallet, setIsSmartContractWallet] = useState(false);
+  const { provider, address } = useWeb3Context()
+  const [isSmartContractWallet, setIsSmartContractWallet] = useState(false)
 
   useEffect(() => {
     const checkAddress = async () => {
-      if (!provider || !address) return;
+      if (!provider || !address) return
 
-      const code = await provider.getCode(address.address);
-      setIsSmartContractWallet(!code || code !== '0x');
-    };
+      const code = await provider.getCode(address.address)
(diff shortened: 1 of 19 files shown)


packages/frontend/src/hooks/useIsSmartContractWallet.ts
```diff
@@ -0,0 +1,20 @@
...
+export default function useIsSmartContractWallet() {
+  const { provider, address } = useWeb3Context();
+  const [isSmartContractWallet, setIsSmartContractWallet] = useState(false);
+
+  useEffect(() => {
+    const checkAddress = async () => {
+      if (!provider || !address) return;
+
+      const code = await provider.getCode(address.address);
+      setIsSmartContractWallet(!code || code !== '0x');
```
kangarang: is there a case that `code` will be falsy here? (i.e. an empty string). if not, i think we can simply this for readability

```suggestion
      setIsSmartContractWallet(code !== '0x');
```

packages/frontend/src/pages/Send/Send.tsx
```diff
@@ -562,8 +564,17 @@ const Send: FC = () => {
         styles={styles}
         customRecipient={customRecipient}
         handleCustomRecipientInput={handleCustomRecipientInput}
+        isOpen={isSmartContractWallet}
       />
 
+      <div className={styles.smartContractWalletWarning}>
+        <Alert severity="warning">{
+          isSmartContractWallet 
+            ? `You are connecting with a smart contract wallet, may be the case you don't own the same address on the destination network. Proceed with caution!`
```
kangarang: Let's change this copy to:

`The connected account is detected to be a smart contract wallet. Please provide a custom recipient to proceed with this transaction.`

kangarang (changes requested):
this is shaping up nicely. just 1 more copy edit and we should be good to merge. thanks again for your contribution @nicosampler

packages/frontend/src/pages/Send/Send.tsx
nicosampler: done!

kangarang (approved):
@nicosampler, LGTM.

great job on this pr, and thanks again for your contribution!

PR #396, opened 2022-02-18
frontend: Add WithdrawalProof class and page

miguelmota:
Example

I sent tx with too low of a bonder fee so bonder did not bond it (as expected)

DeepinScreenshot_select-area_20220219081525

It checks that the transfer root has been set at the destination when trying to withdraw. The proof code is visible in accordion menu

Peek 2022-02-19 08-13

Transfer can be withdrawn after root is set

Peek 2022-02-19 11-13

Transfer cannot be withdrawn again

DeepinScreenshot_select-area_20220219111416

Successful withdrawal tx

Peek 2022-02-19 11-15

Explorer shows bonded status

DeepinScreenshot_select-area_20220219111617


shanefontaine (approved):
LGTM 👍 I only skimmed the FE section, but the rest of it looks good.

One question (does not block merge): is this the first instance of using TheGraph on the FE?

miguelmota:
The frontend was already [using thegraph](https://github.com/hop-protocol/hop/blob/99ccad77fe3b1fcbcb8d35ea3278d016724aa5e5/packages/frontend/src/util/queries.ts) for tracking destination txs

shanefontaine:
Sounds good. Can be merged from my point of view

PR #398, opened 2022-02-23
 Add additional Swap CLI functionality. #348 

### LINKED ISSUE #348: Additional CLI functionality
The CLI could use the following functionality:

- [ ] (`swap`) Implement nativeToken -> wToken (`deposit`) on all chains
- [ ] (`swap`) Implement nativeToken -> hToken on all chains
- [ ] (`swap`) Implement hToken -> nativeToken on all chains
  * I believe this would be two steps: hToken -> canonicalToken then canonicalToken -> nativeToken 
- [x] (`send` and `swap`) implement `--max` when sending or swapping from a native token
  * I believe this means that we would simply take (`balance - (gasLimit * gasPrice)`)
- [ ] (`swap`) Add ability to swap on [Uniswap on Polygon](https://twitter.com/Uniswap/status/1472200289919053826)
- [ ] (`swap`) Update [Uniswap addresses](https://github.com/hop-protocol/hop/blob/314434d60e8ddd7f407c25019dd6283f757aab1b/packages/hop-node/src/uniswap/swap.ts#L134-L159) to make available all of Hop's supported assets on more chains, if liquid
- [ ] If reasonable, allow for two-way Uniswap swaps (i.e. ETH -> USDC or USDC -> ETH). I believe we only support token -> native right now.
- [ ] (`send`) Sending the native asset with [`sendTokens`](https://github.com/hop-protocol/hop/blob/314434d60e8ddd7f407c25019dd6283f757aab1b/packages/hop-node/src/cli/send.ts#L161) does not work as intended. It errors out on the [balance check](https://github.com/hop-protocol/hop/blob/314434d60e8ddd7f407c25019dd6283f757aab1b/packages/hop-node/src/cli/send.ts#L212) because the balance check takes into account the wrapped token and not the native asset. The `tokenClass` should be updated to use the native token, not the wrapped token.
  * Please verify that this is still an issue prior to implementing a fix

miguelmota:
Closes https://github.com/hop-protocol/hop/issues/348

Examples

native -> hToken 

```bash
swap --chain gnosis --from XDAI --to hDAI --amount=1
```

```bash
swap --chain polygon --from MATIC --to hMATIC --amount=1
```

```bash
swap --chain arbitrum --from ETH --to hETH --amount=1
```

hToken -> native 

```bash
swap --chain gnosis --from hDAI --to XDAI --amount=1
```

```bash
swap --chain polygon --from hMATIC --to MATIC --amount=1
```

```bash
swap --chain arbitrum --from hETH --to ETH --amount=1
```

wrapped native -> hToken

```bash
swap --chain gnosis --from WXDAI --to hDAI --amount=1
```

```bash
swap --chain polygon --from WMATIC --to hMATIC --amount=1
```

```bash
swap --chain arbitrum --from WETH --to hETH --amount=1
```

hToken -> wrapped token 

```bash
swap --chain gnosis --from hDAI --to WXDAI --amount=1
```

```bash
swap --chain polygon --from hMATIC --to WMATIC --amount=1
```

```bash
swap --chain arbitrum --from hETH --to WETH --amount=1
```

native -> wrapped native

```bash
swap --chain gnosis --from XDAI --to WXDAI --amount=1
```

```bash
swap --chain polygon --from MATIC --to WMATIC --amount=1
```

```bash
swap --chain arbitrum --from ETH --to WETH --amount=1
```

wrapped native -> native

```bash
swap --chain gnosis --from WXDAI --to XDAI --amount=1
```

```bash
swap --chain polygon --from WMATIC --to MATIC --amount=1
```

```bash
swap --chain arbitrum --from WETH --to ETH --amount=1
```

1inch examples

```
swap --chain gnosis --from XDAI --to USDC --amount=1 // from native xdai to usdc
```

```
swap --chain gnosis --from WXDAI --to USDC --amount=1 // from wrapped xdai to usdc
```

```
swap --chain gnosis --from USDC --to XDAI --amount=1 // from usdc to native xdai 
```

```
swap --chain gnosis --from USDC --to WXDAI --amount=1 // from usdc to wrapped xdai
```

```
swap --chain gnosis --from USDC --to USDT --amount=1 // swap usdc to usdt
```

```
swap --chain polygon --from MATIC --to USDC --amount=1 // from native matic to usdc
```

```
swap --chain polygon --from WMATIC --to USDC --amount=1 // from wrapped matic to usdc
```

```
swap --chain polygon --from USDC --to MATIC --amount=1 // from usdc to native matic
```

```
swap --chain polygon --from USDC --to WMATIC --amount=1 // from usdc to wrapped matic
```

```
swap --chain arbitrum --from ETH --to USDC --amount=0.0003 // from native eth to usdc
```

```
swap --chain arbitrum --from WETH --to USDC --amount=0.0003 // from wrapped eth to usdc
```

```
swap --chain arbitrum --from USDC --to ETH --amount=1 // from usdc to native eth
```

```
swap --chain arbitrum --from USDC --to WETH --amount=1 // from usdc to to wrapped eth
```

Fixes:

bug when sending native token with `send`


miguelmota:
@shanefontaine ready to be reviewed whenever

shanefontaine (changes requested):
* I found it hard to follow the logic around the deposit/withdrawals, hToken AMM swaps, and Uniswap AMM swaps. I think I found it hard to follow not because the code is bad but because there are so many permutations of what can be done. IMO that _may_ lead to unintended bugs or may just be difficult to debug. Again, besides my inline comments, the code _looks_ good. 
 
  One suggestion I have would be to limit a `swap` CLI command to only a single tx per command call. As I understand it, a CLI user could go `--from hETH --to DAI` right now by running the command once. While that seems good, I could see that getting confusing in the code or to the user (if the user thought they were going directly from hETH to DAI, for example). 

  Maybe we can explicitly tell the user that they cannot do this in order to reduce complexity of the code and increase debugability. So if a user does `--from hETH --to DAI`, we would respond with "There is no market for hETH to DAI. Please swap your hETH for ETH first.", or something along those lines.

  Super happy to talk through and be flexible on this, because I know it is a pretty big fundamental change that we may or may not agree on.

* I believe you already started doing this with the Polygon markets. I think it makes sense to start using the pools with lower fees, assuming they are sufficiently liquid for what we are doing. For example, we use 0.3% for USDC/ETH but I think we can start using 0.05%. The former has $324m TVL while the latter has $283m, which is far more than we will ever be dealing with. The volume on the cheaper pool is also 10x that of the more expensive one. I went through the addresses and the following are what I think they can be. Please review and LMK if you agree. 

```
    pools: {
      ETH: { // ETH is "toToken"
        USDC: '0x88e6A0c2dDD26FEEb64F039a2c41296FcB3f5640', // updated from .3 to .05
        USDT: '0x11b815efB8f581194ae79006d24E0d814B7697F6', // updated from .3 to .05
        DAI: '0x60594a405d53811d3bc4766596efd80fd545a270', // no change
        MATIC: '0x290A6a7460B308ee3F19023D2D00dE604bcf5B42', // no change
        WBTC: '0x4585fe77225b41b697c938b018e2ac67ac5a20c0' // no change
      }
    }
  },
  optimism: {
    swapRouter: '0xE592427A0AEce92De3Edee1F18E0157C05861564',
    pools: {
      ETH: { // ETH is "toToken"
        USDC: '0xB589969D38CE76D3d7AA319De7133bC9755fD840', // no change
        USDT: '0xc858A329Bf053BE78D6239C4A4343B8FbD21472b', // no change
        DAI: '0x03aF20bDAaFfB4cC0A521796a223f7D85e2aAc31', // no change
        WBTC: '0x73b14a78a0d396c521f954532d43fd5ffe385216' // no change.
      }
    }
  },
  arbitrum: {
    swapRouter: '0xE592427A0AEce92De3Edee1F18E0157C05861564',
    pools: {
      ETH: { // ETH is "toToken"
        USDC: '0xc31e54c7a869b9fcbecc14363cf510d1c41fa443', // updated from .3 to .05
        USDT: '0x641c00a822e8b671738d32a431a4fb6074e5c79d', // no change
        DAI: '0xa961f0473da4864c5ed28e00fcc53a3aab056c1b', // no change
        WBTC: '0x2f5e87c9312fa29aed5c179e456625d79015299c' // no change
      }
    }
  },
  polygon: {
    swapRouter: '0xE592427A0AEce92De3Edee1F18E0157C05861564',
    pools: {
      MATIC: { // MATIC is "toToken"
        USDC: '0xa374094527e1673a86de625aa59517c5de346d32', // no change
        USDT: '0x9b08288c3be4f62bbf8d1c20ac9c5e6f9467d8b7', // no change
        DAI: '0xfe530931da161232ec76a7c3bea7d36cf3811a0d', // no change
        ETH: '0x86f1d8390222a3691c28938ec7404a1661e618e0',  // no change
        WBTC: '0x6b75f2189f0e11c52e814e09e280eb1a9a8a094a' // no change
      }
    }
```

* I did not expect this, but some of the markets are super small (for example, the most liquid [DAI on Polygon](https://info.uniswap.org/#/polygon/pools/0xfe530931da161232ec76a7c3bea7d36cf3811a0d) market only has $33k). I think we should not allow `swaps` to go through where slippage is `>1%`, unless the CLI user explicitly specifies a different slippage.

packages/hop-node/src/cli/swap.ts
```diff
@@ -15,50 +15,72 @@ root
   .description('Swap tokens on Uniswap or via Hop AMM')
   .option('--chain <slug>', 'Chain', parseString)
   .option('--from <symbol>', 'From token', parseString)
+  .option('--from-native [boolean]', 'From native token', parseBool)
```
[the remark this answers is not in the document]
miguelmota: >  a simple --from DAI --to hDAI seems to be pretty logical

not sure about that because it's unclear if `DAI` here is referring to `XDAI` or `WXDAI` since gnosis AMM only works with `WXDAI`

```bash
swap --chain gnosis --from DAI --to hDAI // unclear if swap is using XDAI or WXDAI
```

while this is more clear imo

```bsh
swap --chain gnosis --from-native --to hDAI // explicit about using XDAI for swap
```

I'm ok with removing `--from-native` and `--to-native` but change it use the exact tokens as cli arguments (`WXDAI`, `WMATIC`, `WETH`) for AMM swap on their native chains and `XDAI`, `MATIC`, `ETH` to refer to native `from` tokens which will wrap tokens before swapping

miguelmota:
> As I understand it, a CLI user could go --from hETH --to DAI right now by running the command once. 

This is not the case, it's not possible to go from hToken-A to Token-B as is and the cli will throw an error stating that they must be the same asset type for hToken swaps

Some examples

swap --from Token-A --to WToken-A // wrap
swap --from WToken-A --to Token-A // unwrap
swap --from WToken-A --to WToken-A // error, token cannot be same
swap --from Token-A --to Token-A // error, token cannot be same
swap --from hToken-A --to hToken-A // error, token cannot be same
swap --from Token-A --to Token-B // uniswap swap
swap --from Token-B --to Token-A // uniswap swap
swap --from hToken-A --to Token-A // AMM swap
swap --from Token-A --to hToken-A // AMM swap
swap --from hToken-A --to hToken-B // error, both can't be hTokens
swap --from hToken-A --to Token-B // error, must be same asset type

> start using the pools with lower fees

Thanks I updated the addresses to the ones you posted. The slippage tolerance is `.1%` by default and be configured with cli flag

packages/hop-node/src/utils/getCanonicalTokenSymbol.ts
```diff
@@ -1,5 +1,5 @@
 function getCanonicalTokenSymbol (tokenSymbol: string) {
-  return tokenSymbol.replace(/^h/, '')
+  return tokenSymbol.replace(/^[hWX]/g, '')
```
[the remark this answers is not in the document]
miguelmota: the official token symbol is `XDAI` so I don't think there would be a use case of lowercase `xDai` to refer to token

packages/hop-node/src/uniswap/swap.ts
```diff
@@ -200,17 +224,27 @@ export async function swap (config: Config) {
   let routeToken1: any = pool.token1
 
   const token0Symbol = await token0.symbol()
-  if (token0Symbol === 'WETH' || token0Symbol === 'ETH') {
+  if ((chain !== 'polygon' && token0Symbol === 'WETH' || token0Symbol === 'ETH') || (chain === 'polygon' && token0Symbol === 'WMATIC' || token0Symbol === 'MATIC')) { // eslint-disable-line
```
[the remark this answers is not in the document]
miguelmota: Yes that's a bug. Resolved with https://github.com/hop-protocol/hop/commit/f91418c9405c25da0f10e830b6327adacce25320

packages/hop-node/src/uniswap/swap.ts
```diff
@@ -200,17 +224,27 @@ export async function swap (config: Config) {
...
     sourceToken = token1
     const tmp = routeToken0
     routeToken0 = routeToken1
     routeToken1 = tmp
   }
 
-  if (toToken === 'ETH') {
+  if (chain !== 'polygon' && toToken === 'ETH') {
```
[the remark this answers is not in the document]
miguelmota: Resolved with https://github.com/hop-protocol/hop/commit/f91418c9405c25da0f10e830b6327adacce25320

packages/hop-node/src/cli/swap.ts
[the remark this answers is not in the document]
miguelmota: Resolved by https://github.com/hop-protocol/hop/pull/398/commits/74d656ec6f3c838f2d866d0aa5e97ab78fcb5774 (fromNative toNative flag removed)

packages/hop-node/src/cli/swap.ts
```diff
@@ -15,50 +15,72 @@ root
...
-    const data = ethersInterface.encodeFunctionData(
-      'withdraw', [parsedAmount]
-    )
-    tx = await wallet.sendTransaction({
-      to: wrappedTokenAddress,
-      data
-    })
+  const isWrapperSwap = (isWrapperDeposit || isWrapperWithdrawal) && !isAmmSwap
+  if (isWrapperSwap) {
+    if (isWrapperDeposit) {
```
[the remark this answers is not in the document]
miguelmota: Resolved by https://github.com/hop-protocol/hop/pull/398/commits/74d656ec6f3c838f2d866d0aa5e97ab78fcb5774

miguelmota:
@shanefontaine let me know if you have any questions on the updates

packages/hop-node/src/1inch/1inch.ts
```diff
@@ -0,0 +1,347 @@
...
+    if (!slippage) {
+      throw new Error('slippage is required')
+    }
+
+    const url = this.constructUrl('/swap', {
+      fromTokenAddress,
+      toTokenAddress,
+      amount,
+      fromAddress,
+      slippage,
```
shanefontaine, line 193: I see we use `0.1` slippage by default. Just confirming, do you know if this endpoint expect the slippage in whole numbers or percentage (i.e. does it expect our `0.1` value or is it looking for `0.001`)?

packages/hop-node/src/1inch/1inch.ts
```diff
@@ -0,0 +1,347 @@
...
+        token = new Token(tokenContracts.l1CanonicalToken)
+      } else {
+        token = new Token(tokenContracts.l2CanonicalToken)
+      }
+
+      amount = (await token.getBalance()).toString()
+      formattedAmount = Number(formatUnits(amount.toString(), getTokenDecimals(fromToken)))
+    }
+  } else {
+    amount = parseUnits(formattedAmount.toString(), getTokenDecimals(fromToken)).toString()
```
shanefontaine: We handle from native vs. token explicitly in the `(max)` case. Do we need to do the same here? In other words, does `getTokenDecimals()` work for ETH? I believe the answer is [yes](https://github.com/hop-protocol/hop/blob/749dcb013c22b666b3ef38bf7af525b1d1f4a922/packages/core/src/metadata/tokens.ts#L16-L20), but want to confirm.

packages/hop-node/src/1inch/1inch.ts
```diff
@@ -0,0 +1,347 @@
...
+
+    const url = this.constructUrl('/swap', {
+      fromTokenAddress,
+      toTokenAddress,
+      amount,
+      fromAddress,
+      slippage,
+      destReceiver
+    })
+    const result = await this.getJson(url)
```
shanefontaine, line 196: What is the result when the returned amount is out of the slippage range defined? I think we should explicitly handle that here.

packages/hop-node/src/1inch/1inch.ts
```diff
@@ -0,0 +1,347 @@
...
+    logger.debug('dryMode enabled, skipping swap tx')
+  } else {
+    logger.debug('sending swap tx')
+    const tx = await wallet.sendTransaction(txData)
+    logger.debug('swap tx:', tx.hash)
+    return tx
+  }
+}
+
+const nativeChainTokens: Record<string, string> = {
```
shanefontaine: Nitpick: maybe move this to `constants`. I think I've needed this in a few other places.

packages/hop-node/src/cli/swap.ts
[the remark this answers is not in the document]
shanefontaine: Nice. I think this makes great sense 👍

packages/hop-node/src/cli/swap.ts
```diff
@@ -8,23 +8,30 @@ import { BigNumber, utils as ethersUtils } from 'ethers'
...
 
 root
   .command('swap')
   .description('Swap tokens on Uniswap or via Hop AMM')
   .option('--chain <slug>', 'Chain', parseString)
   .option('--from <symbol>', 'From token', parseString)
+  .option('--from-native [boolean]', 'From native token', parseBool)
   .option('--to <symbol>', 'To token', parseString)
-  .option('--amount <number>', 'From token amount', parseNumber)
+  .option('--to-native [boolean]', 'To native token', parseBool)
```
shanefontaine: IICU, thie `--to-native` and the above `--from-native` options are now artifacts and can be removed.

packages/hop-node/src/cli/swap.ts
```diff
@@ -121,19 +149,40 @@ async function main (source: any) {
...
+          const unwrapTx = await unwrapToken(chain, parsedAmount)
+          logger.debug('waiting for unwrap tx confirmation')
+          await unwrapTx.wait()
+          tx = unwrapTx
+          break
+        }
+      }
+    }
   } else {
     logger.debug('uniswap swap')
```
shanefontaine: `uniswap` in the log can now be updated

packages/hop-node/src/cli/swap.ts
```diff
@@ -121,19 +149,40 @@ async function main (source: any) {
...
       amount,
       max,
-      deadline,
       slippage,
-      recipient
+      recipient,
+      dryMode
     })
   }
+  if (dryMode) {
```
shanefontaine, line 205: I believe this dry mode only stops the `wait()` and does not stop the actual tx being broadcasted. I think this should go above the swap functions.

shanefontaine (approved):
@miguelmota back to you. This looks awesome. The 1inch stuff is incredibly clean  and well done 👏 

Thanks for the detailed responses. Left a few minor comments, but LGTM.

packages/hop-node/src/1inch/1inch.ts
miguelmota: The [docs](https://docs.1inch.io/docs/aggregation-protocol/api/swagger/) state that the number should range from 1 to 50, so I think we are good with 0.1  (0.1%)

DeepinScreenshot_select-area_20220301093021

packages/hop-node/src/1inch/1inch.ts
miguelmota: good idea, updated with https://github.com/hop-protocol/hop/commit/a38b614cc480ac4cfdd492433b6bb30041374d73

packages/hop-node/src/cli/swap.ts
miguelmota: good catch, removed in https://github.com/hop-protocol/hop/commit/a38b614cc480ac4cfdd492433b6bb30041374d73

packages/hop-node/src/cli/swap.ts
miguelmota: updated in https://github.com/hop-protocol/hop/commit/a38b614cc480ac4cfdd492433b6bb30041374d73

packages/hop-node/src/cli/swap.ts
miguelmota: resolved with https://github.com/hop-protocol/hop/commit/a38b614cc480ac4cfdd492433b6bb30041374d73

packages/hop-node/src/1inch/1inch.ts
miguelmota: The api returns this `cannot estimate` error if slippage param is too low
DeepinScreenshot_select-area_20220301094949

but i'll add additional check as well

packages/hop-node/src/1inch/1inch.ts
miguelmota: yes it works with eth symbol

packages/hop-node/src/1inch/1inch.ts
miguelmota: the calculated slippage check has been [added](https://github.com/hop-protocol/hop/commit/1e3f2a74c4b1b772710515bd935e045ffd718a9d) and could use another set of eyes for sanity check

PR #399, opened 2022-02-24
UI: improve stats route styles

kangarang:
this pr adds mobile-responsiveness to the stats route


miguelmota:
some minor comments and suggestions on tables

1. more visible loading indicators would be good (a table was missing loader completely)

DeepinScreenshot_select-area_20220224141138

the previous animated loading skeleton was pretty nice 

DeepinScreenshot_select-area_20220224141122

2. since the data is a lot more compact, horizontal lines and row hover background would make readability better and easier to follow with eyes

DeepinScreenshot_select-area_20220224141632

previous row lines example

DeepinScreenshot_select-area_20220224142918


3. some tables display only chain logo but some have full name. Maybe we can be consistent and only show chain logos only (as well as rename "network" to "chain" across tables for consistency) 

DeepinScreenshot_select-area_20220224142426

DeepinScreenshot_select-area_20220224141648

4. column header names should be fully spelled out for readability (some headers are abbreviated here but not others)

DeepinScreenshot_select-area_20220224142529

5. token symbol alignment could be more consistent

DeepinScreenshot_select-area_20220224142438

DeepinScreenshot_select-area_20220224142544

6. token logo is a little redundant in this table

DeepinScreenshot_select-area_20220224143441

7. the token symbol would make more sense to be on the token column rows instead of virtual debt column rows 

DeepinScreenshot_select-area_20220224143256

8. it would be nice if the collapsible table open/close state were stored in local storage since preference is cleared when refreshing the page 
 
DeepinScreenshot_select-area_20220224141623

9. mostly a nitpick; the square white background could use round edges (maybe box shadows too?) to match rest of site theme

DeepinScreenshot_select-area_20220224143909

10. on mobile screen size the text is pretty small so a suggestion would be to increase font size and make table horizontally scrollable (you may check out explorer on mobile for reference)

DeepinScreenshot_select-area_20220224144108

---

the above is mostly ui stuff but the code changes look good!

kangarang:
@miguelmota thanks for the great review! i've addressed the ones i felt were most needed.

overall, i refactored the components to use `react-table`. handling styles is much more consistent and easier to manage.

mobile screen:

Screen Shot 2022-02-28 at 10 56 34 PM

hovering:

Screen Shot 2022-02-28 at 10 56 52 PM

horizontal scrolling:

Screen Shot 2022-02-28 at 10 57 01 PM

larger screen:

Screen Shot 2022-02-28 at 10 57 11 PM

each table is sortable by column. however the sorting is incorrect (likely needs a helper function):

Screen Shot 2022-02-28 at 11 01 37 PM

Screen Shot 2022-02-28 at 11 01 49 PM

miguelmota:
@kangarang works good!

The table titles are redundant. Maybe the collapsible title can be removed?

DeepinScreenshot_select-area_20220301092328


I think the token symbol here can be removed since the header labels what it is

DeepinScreenshot_select-area_20220301092343

---

shane and chris look at this page more than I do so it'd be good to their feedback as well

miguelmota:
@kangarang after looking at it for longer, I think the high contract between white bg and dark vertical border lines makes things harder to read

here's a mock with no bg and only light horizontal lines

DeepinScreenshot_select-area_20220301113307

PR #401, opened 2022-02-28
hop-node: GasBoostSigner: Check for reorg and rebroadcast transaction

miguelmota:
- checks for receipt again after `waitConfirmations` and rebroadcasts using same nonce if reorg occurred (puts original tx through `send` flow again)
- tested logic with a test simulating reorg


shanefontaine (approved):
This is awesome. I have no questions or suggestions. Nice job 👍

PR #404, opened 2022-03-04
Feature/vault

miguelmota:
- [x] add class to deposit, withdraw, read balance from vault
- [x] use yearn as vault strategy (using [yearn-sdk](https://github.com/yearn/yearn-sdk))
- [x] deposit/withdraw with erc20 directly (no zapIn/zapOut)
- [x] deposit/withdraw ETH with zapIn/zapOut (since vault requires WETH)
- [x] add cli command to deposit or withdraw from vault
- [x] include vault balance in available liquidity on hop node side
- [x] update sdk to read available liquidity from hop node worker which includes vault balance
- [x] add method to unstake from hop and deposit to vault
- [x] add method to withdraw from vault and stake on hop
- [x] update BondWithdrawalWatcher to withdraw from vault if needed (with mutex)
- [x] update BondTransferRootWatcher to withdraw from vault if needed (with mutex)
- [x]  update frontend stats page to include vault balance

---
note: the ability to auto-deposit back to vault when liquidity is no longer needed will be added in a later PR after to take it one-step a t time, so deposit will be manual but withdrawals will be automatic in mvp


packages/hop-node/src/vault/YearnVault.ts
```diff
@@ -0,0 +1,226 @@
...
+    super(url)
+    this.signer = signer
+  }
+
+  getSigner (accountOrIndex?: string | number) {
+    return this.signer
+  }
+}
+
+type ChainId = 1 | 250 | 1337 | 42161
```
shanefontaine: Nitpick: can we scope this down to just the relevant chainIds? Basically remove `250` and `1337`?

packages/hop-node/src/vault/YearnVault.ts
```diff
@@ -0,0 +1,226 @@
...
+    const contract = this.getErc20(token)
+    const allowance = await contract.allowance(account, vault)
+    const needsApproval = allowance.lt(amount)
+    return needsApproval
+  }
+
+  async needsWithdrawalApproval (amount: BigNumber = constants.MaxUint256) {
+    const account = await this.signer.getAddress()
+    const { token, vault, zapOut } = addresses[this.token]
+    if (token !== EthAddress) {
```
shanefontaine, line 153: I might be misunderstanding yearn, but I would think that we would never be sending ETH into a contract to `withdraw` anything. Because of this, I think that this if statement should not exist. Please LMK if I am missing something.

packages/hop-node/src/vault/Vault.ts
```diff
@@ -0,0 +1,16 @@
+import { Chain } from 'src/constants'
+import { DefaultVault } from './DefaultVault'
+import { VaultInterface } from './VaultInterface'
+import { YearnVault } from './YearnVault'
+
+export class Vault extends DefaultVault implements VaultInterface {
```
cwhinfrey: `DefaultVault` implements `VaultInterface` so it feels redundant to have `implements VaultInterface` here as well

packages/hop-node/src/vault/Vault.ts
```diff
@@ -0,0 +1,16 @@
...
+  static from (chain: Chain, token: string, signer: any): VaultInterface {
+    const yearnVaultTokens = new Set(['ETH', 'USDC', 'USDT', 'DAI'])
+    const useYearnVault = chain === Chain.Ethereum && yearnVaultTokens.has(token)
+    if (useYearnVault) {
+      return new YearnVault(chain, token, signer)
+    }
+
+    return new DefaultVault(chain, token, signer)
```
cwhinfrey: Returning a non-functional vault when none exists seems counterintuitive. I'd expect it to throw or return undefined to be handled downstream.

packages/hop-node/src/vault/YearnVault.ts
```diff
@@ -0,0 +1,226 @@
...
+  }
+}
+
+type ChainId = 1 | 250 | 1337 | 42161
+
+export class YearnVault implements VaultInterface {
+  chain: Chain
+  token: string
+  signer: any
+  slippage = 0.1
```
cwhinfrey: I think it might be irrelevant for our use cases but setting this lower to something like 0.01 might be safer. (assuming 0.1 means up to 10% slippage)

packages/hop-node/src/vault/Vault.ts
cwhinfrey: If you agree with the above, I'd consider removing `DefaultVault` and `VaultInterface` entirely and then change this class to an abstract class with all the functions in `VaultInterface` (but also marked as abstract) alongside the `from` static function. That way there's just a single base class that has the `from` helper for creating new Vaults that inherit from it.

There might be a downside I'm missing and I think the current structure works well as is too if you prefer it.

cwhinfrey (approved):
This looks great! Left a few comments but all optional so marking as approved.

shanefontaine (approved):
LGTM 👍  Only minor comments

packages/hop-node/src/vault/YearnVault.ts
miguelmota: It's needed to allow zapper contract to withdraw yvETH (the condition is checking that we are only interested in approving zap contract if vault is for ETH because it requires to zapOut)

You can see the yearn UI has an approval for ETH withdrawal

DeepinScreenshot_select-area_20220307182107

packages/hop-node/src/vault/YearnVault.ts
miguelmota: Removed in https://github.com/hop-protocol/hop/pull/404/commits/420a217717cf27b3dd78c3700f08b632d213b1cd

(the yearn sdk required the chain id type to be exact numbers)

packages/hop-node/src/vault/YearnVault.ts
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/404/commits/420a217717cf27b3dd78c3700f08b632d213b1cd

packages/hop-node/src/vault/Vault.ts
miguelmota: Resolved in https://github.com/hop-protocol/hop/pull/404/commits/bcd375614190d12e4aecaa4891c4d8715acafac8

packages/hop-node/src/vault/Vault.ts
miguelmota: Good suggestion, the redundant classes/interfaces have been removed and updated to be abstract class

Resolved in https://github.com/hop-protocol/hop/pull/404/commits/bcd375614190d12e4aecaa4891c4d8715acafac8

packages/hop-node/src/vault/Vault.ts
```diff
@@ -0,0 +1,33 @@
...
+export abstract class Vault {
+  static from (chain: Chain, token: string, signer: any) {
+    const yearnVaultTokens = new Set(['ETH', 'USDC', 'USDT', 'DAI'])
+    const useYearnVault = chain === Chain.Ethereum && yearnVaultTokens.has(token)
+    if (useYearnVault) {
+      return new YearnVault(chain, token, signer)
+    }
+  }
+
+  async getBalance (account?: string): Promise<BigNumber> {
```
cwhinfrey: Super minor but if you mark these functions as abstract instead of throwing an error, any missing function definitions in child classes will be caught in compile time rather than run time

miguelmota: oh nice, works perfectly

Resolved with https://github.com/hop-protocol/hop/pull/404/commits/19508919855581f7770c150704123ebefffb6c65

miguelmota:
Thanks for the good feedback chris and shane

Will merge after https://github.com/hop-protocol/hop/pull/405 is merged and then we can do a test run on DAI bonder

packages/hop-node/src/vault/YearnVault.ts
shanefontaine: Just played with yearn and see the same, so I believe you are correct. I guess I'm just confused about the difference between the ETH vault (which needs approval) vs. the token vaults (which don't need approval). Anyway, LGTM 👍

shanefontaine (approved):
Changes LGTM 👍 

Only comment that requires no change right now: looks like yearn is pretty friendly to `permit`. For example, the WETH yVault allows you to use [`ZapOutWithPermit`](https://etherscan.io/tx/0xc3cc6eefe07133f7efa32e720ca847c7b5c478d78e47be1818ca089143887430). Just noting this here so that we keep it in mind for a future optimization.

miguelmota:
@shanefontaine great suggestion. Will keep that in mind for when we start using yearn

PR #405, opened 2022-03-07
Feature/gasboost nonce db

miguelmota:
- [x] use db to store nonce state and read nonce state (removes in-memory nonce state)
- [x] on init, read on-chain pending nonce and compare against db nonce. Set current nonce to which ever is highest
- [x] uses [`level-party`](https://github.com/Level/party) (official lib by level team) to work around db locks 

An example script to show how 2 separate node process can access the db at the same time

```js
const level = require('level-party')
const wait = require('wait')
const db = level(__dirname + '/data', { valueEncoding: 'json' })
const reset = process.argv[2]

async function main() {
  if (reset) {
    await db.put('nonce', {nonce: 1})
    for (let i = 0; i < 10; i++) {
      const key = 'nonce'
      await wait(1000)
      let { nonce } = await db.get(key)
      console.log('nonce used', { nonce} )
      let value = {nonce: ++nonce}
      await db.put(key, value)
      console.log('set', key, '=', await db.get(key))
    }
  } else {
    const key = 'nonce'
    let { nonce } = await db.get(key)
    console.log('nonce used', nonce)
    let value = {nonce: ++nonce}
    await db.put(key, value)
    console.log('set', key, '=', await db.get(key))
  }
}

main()
```

The top half is simulating the bonder that is always running.
The bottom half simulating one-off txs as separate command.

Peek 2022-03-07 18-04

```
$ node test.js
nonce used { nonce: 1 }
set nonce = { nonce: 2 }
nonce used { nonce: 2 }
set nonce = { nonce: 3 }
nonce used { nonce: 3 }
set nonce = { nonce: 4 }
nonce used { nonce: 5 }
set nonce = { nonce: 6 }
nonce used { nonce: 6 }
set nonce = { nonce: 7 }
nonce used { nonce: 8 }
set nonce = { nonce: 9 }
nonce used { nonce: 9 }
set nonce = { nonce: 10 }
nonce used { nonce: 10 }
set nonce = { nonce: 11 }
nonce used { nonce: 11 }
set nonce = { nonce: 12 }
nonce used { nonce: 12 }
set nonce = { nonce: 13 }
```

```
$ node test.js
nonce used 4
set nonce = { nonce: 5 }

$ node test.js
nonce used 7
set nonce = { nonce: 8 }
```

Notice how `nonce used` is never repeated by either process, as expected

You can also run multiple instances of bonder node on same db to try it out

I've confirmed it works with docker

### DIFF
--- a/packages/hop-node/src/gasboost/GasBoostSigner.ts
+++ b/packages/hop-node/src/gasboost/GasBoostSigner.ts
@@ -71,10 +71,8 @@ class GasBoostSigner extends Wallet {
   }
 
   private async setLatestNonce () {
-    const dbNonce = await this.getDbNonce()
     const onChainNonce = await this.getOnChainNonce()
-    const nonce = Math.max(dbNonce, onChainNonce)
-    await this.setDbNonce(nonce)
+    await this.setDbNonce(onChainNonce)
   }
 
   // this is a required ethers Signer method
--- a/packages/hop-node/src/gasboost/GasBoostSigner.ts
+++ b/packages/hop-node/src/gasboost/GasBoostSigner.ts
@@ -1,8 +1,9 @@
-import GasBoostTransaction from './GasBoostTransaction'
 import GasBoostTransactionFactory, { Options } from './GasBoostTransactionFactory'
 import Logger from 'src/logger'
+import MemoryStore from './MemoryStore'
 import Store from './Store'
 import getProviderChainSlug from 'src/utils/getProviderChainSlug'
+import wait from 'src/utils/wait'
 import { Mutex } from 'async-mutex'
 import { NonceTooLowError } from 'src/types/error'
 import { Notifier } from 'src/notifier'
@@ -14,7 +15,6 @@ class GasBoostSigner extends Wallet {
   items: string[] = []
   lastTxSentTimestamp: number = 0
   delayBetweenTxsMs: number = 7 * 1000
-  nonce: number = 0
   chainSlug: string
   gTxFactory: GasBoostTransactionFactory
   signer: Signer
@@ -22,8 +22,9 @@ class GasBoostSigner extends Wallet {
   logger: Logger
   notifier: Notifier
   mutex: Mutex
+  ready: boolean = false
 
-  constructor (privateKey: string, provider?: providers.Provider, store?: Store, options: Partial<Options> = {}) {
+  constructor (privateKey: string, provider?: providers.Provider, store: Store = new MemoryStore(), options: Partial<Options> = {}) {
     super(privateKey, provider)
     this.signer = new Wallet(privateKey, provider)
     if (store != null) {
@@ -35,7 +36,7 @@ class GasBoostSigner extends Wallet {
     }
     this.chainSlug = chainSlug
     this.mutex = new Mutex()
-    this.gTxFactory = new GasBoostTransactionFactory(this.signer, this.store)
+    this.gTxFactory = new GasBoostTransactionFactory(this.signer)
     const tag = 'GasBoostSigner'
     const prefix = `${this.chainSlug}`
     this.logger = new Logger({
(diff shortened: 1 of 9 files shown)


packages/hop-node/src/gasboost/GasBoostSigner.ts
```diff
@@ -46,72 +47,86 @@ class GasBoostSigner extends Wallet {
...
+  protected async tilReady (): Promise<boolean> {
+    if (this.ready) {
+      return true
+    }
+
+    await wait(100)
+    return await this.tilReady()
+  }
+
+  private async setLatestNonce () {
```
shanefontaine, line 73: IMO the nonce from the initialization of the bonder should only come from on-chain and should not care about the DB. I cannot think of a case where the DB would be correct but on-chain would be incorrect.

shanefontaine (commented):
* Why does there need to be a DB per chain? If there is only 1 db with _keys_ for every chain, would the gasBoostDb logic in `src/db/db.ts` be more standard and be included in `DbSet` and match all the other DB logic in that file instead of having a separate function?
  * My naive suggestion would be to make it a single DB with per-chain keys, but I may be missing something.

* I believe we used to have memory issues with the `store` in the gas boost tx class because we were not tearing them down after we created them. My understanding is that the signer now accepts a DB as a store, but defaults to an in-memory store if a DB is not provided. If the in-memory store is used for some reason, will we run into memory issues? I believe the answer is no, since we should only (a) instantiate a single signer class, and (b) use a single `nonce` key, but want to get your confirmation.

@miguelmota back to you

packages/hop-node/src/gasboost/GasBoostSigner.ts
miguelmota: it will cause issues if the stored nonces are being overridden by separate processes which defeats the purpose

a scenario,
1. bonder is running as usual
3. a bonder command (like a stake command) is ran to send tx
4. right before the bonder command sends tx, the running bonder sends tx with nonce 10 (ie a usual bond withdawal tx)
5. the bonder command sends tx with nonce 10 as well because it fetched on-chain nonce within seconds of last tx send instead of using db nonce which would of been nonce 11. The bonder command tx used a slightly higher gas price so the running bonder tx fails and causes 1hr delay 

related: https://github.com/hop-protocol/hop/pull/178#issuecomment-929880265

miguelmota:
> Why does there need to be a DB per chain? 

There is a GasBoostSigner per chain so it makes sense to have a GasBoostSignerDb per chain

> If there is only 1 db with keys for every chain, would the gasBoostDb logic in src/db/db.ts be more standard and be included in DbSet and match all the other DB logic in that file instead of having a separate function?

I don't think so because `DbSet` is for getting db per token (not per chain)

> If the in-memory store is used for some reason, will we run into memory issues? 

I don't think so because only the db store is used for wallet. The in-memory store is only used for testing

shanefontaine:
> There is a GasBoostSigner per chain so it makes sense to have a GasBoostSignerDb per chain

Understood and agreed now.

> I don't think so because only the db store is used for wallet. The in-memory store is only used for testing

Gotcha 👍

packages/hop-node/src/gasboost/GasBoostSigner.ts
shanefontaine: I see. I think there is another scenario where a tx gets re-orged out (although I believe the bug has been fixed) and the DB is out of sync. In that case, I think there is no way for the bonder to recover, since a new instantiation would use the DB state. Right now, we `sendToSelf`, which uses on-chain state and fixes the issue, but that would not be possible with this update.

If what I said is true, then I'm pretty torn. I'd lean towards still not doing `max()` and only using on-chain for init _only_ because it can get the bonder into an unrecoverable state (either the re-org mentioned above or a potential yet-unknown bug with the bonder software or RPC provider). I do not believe we should not allow the bonder to get in an unrecoverable state, if possible.

I _think_ the only downside of not using `max()` is that a user waits 1 hour if the timing is wrong.

LMK what you think, pretty open to discussion here.

miguelmota: That makes sense 👍  I'll update it accordingly to set pending nonce on start

```ts
  private async setLatestNonce () {
    const onChainNonce = await this.getOnChainNonce()
    await this.setDbNonce(onChainNonce)
}
```

PR #406, opened 2022-03-08
Feature/vault auto deposit

miguelmota:
- [x] add `vault` config to config file with threshold and deposit amounts
- [x] check if it needs to deposit after settling by checking is available liquidity is above threshold


miguelmota:
Ready for review @shanefontaine

miguelmota:
Combined with https://github.com/hop-protocol/hop/pull/446

Closing

PR #408, opened 2022-03-11
UI: Add support for disabling network-specific txs

kangarang:
this pr enables easy disabling of network-specific transactions.

hook: `useDisableTxs.ts`
config: `disabled.ts`
env vars: `REACT_APP_DISABLED_ROUTES`, `REACT_APP_WARNING_ROUTES`

the `REACT_APP_WARNING_ROUTES` env var is optional. the default behavior is to disable transactions listed in `REACT_APP_DISABLED_ROUTES`.

syntax: `[source-slug]:[destination-slug]` (for multiple pairs, string is split by `,`)
ex: `REACT_APP_DISABLED_ROUTES=polygon:arbitrum,optimism:gnosis`

to disable all chains for source or destination, the slug can be replaced with `all`
ex. `REACT_APP_DISABLED_ROUTES=polygon:all,all:polygon`


packages/frontend/src/config/disabled.ts
```diff
@@ -0,0 +1,95 @@
...
+  const disabledRoutes = serializedDisabledRoutes?.split(',')
+  const warningRoutes = serializedWarningRoutes?.split(',')
+
+  if (disabledRoutes[0] !== '') {
+    return (disabledRoutes as any[]).map((disabledRoute, i) => {
+      const indexOfColon = disabledRoute.indexOf(':')
+      const source = disabledRoute.slice(0, indexOfColon)
+      const destination = disabledRoute.slice(indexOfColon + 1)
+
+      if (!(source in Slug || source === ALL) || !(destination in Slug || destination === ALL)) {
```
miguelmota: minor suggestion: use a set here to simplify the condition

```ts
const validValues = new Set([...Object.values(Slug), ALL])
if (!(validValues.has(source) && validValues.has(destination))) {
  // error
}
```

packages/frontend/src/config/disabled.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export function parseDisabledRoutes(
+  serializedDisabledRoutes: string,
+  serializedWarningRoutes: string
+): DisabledRoute[] {
+  const disabledRoutes = serializedDisabledRoutes?.split(',')
+  const warningRoutes = serializedWarningRoutes?.split(',')
+
+  if (disabledRoutes[0] !== '') {
```
miguelmota, line 50: minor suggestion: return early here instead

```js
if (!disabledRoutes[0]) {
 return []
}
```

packages/frontend/src/config/disabled.ts
```diff
@@ -0,0 +1,95 @@
...
+    return (disabledRoutes as any[]).map((disabledRoute, i) => {
+      const indexOfColon = disabledRoute.indexOf(':')
```
miguelmota: suggestion: do a split here instead to simplify logic

```js
const [ source, destination ] = disabledRoute.split(':')
```

packages/frontend/src/pages/Convert/ConvertContent.tsx
```diff
@@ -105,9 +106,16 @@ const ConvertContent: FC = () => {
     approveTokens()
   }
 
+  useEffect(() => {
+    if (sourceNetwork?.slug === ChainSlug.Polygon || destNetwork?.slug === ChainSlug.Polygon) {
+      return setManualWarning('Warning: transfers to/from Polygon are temporarily down.')
```
miguelmota: is this still needed?

miguelmota:
minor bug: The error messages says `from All` but should say `from Gnosis` here

DeepinScreenshot_select-area_20220314181801

DeepinScreenshot_select-area_20220314181808

packages/frontend/src/config/disabled.ts
kangarang: cool idea

packages/frontend/src/pages/Convert/ConvertContent.tsx
kangarang: this is old code. after rebasing, it's up-to-date

packages/frontend/src/config/disabled.ts
```diff
@@ -0,0 +1,102 @@
...
+export function parseDisabledRoutes(
+  serializedDisabledRoutes: string,
+  serializedWarningRoutes: string
+): DisabledRoute[] {
+  const disabledRoutes = serializedDisabledRoutes?.split(',')
+  const warningRoutes = serializedWarningRoutes?.split(',')
+
+  if (!disabledRoutes[0]) return []
+
+  if (disabledRoutes[0] !== '') {
```
miguelmota, line 50: I think this if statement can be removed now

miguelmota (approved):

PR #411, opened 2022-03-16
UI: add sourceChain catch-all check to txConfirm hook

kangarang:
this pr adds a catch-all network check to the `useSendingTransaction()` hook

calling `useSendingTransaction` now requires `source`, `token`, or `token0` within its `inputProps`


packages/frontend/src/components/txConfirm/useSendingTransaction.ts
```diff
@@ -1,17 +1,49 @@
...
   async function handleSubmit(opts?: any) {
     try {
+      let src: Partial<NetworkTokenEntity> | undefined
+      if ('source' in props) {
+        src = props.source
+      } else if ('token' in props) {
+        src = props.token
+      } else if ('token0' in props) {
+        src = props.token0
+      }
```
miguelmota, line 46: maybe we should throw if `else`? to make sure there's always a chain passed in

miguelmota (approved):

PR #412, opened 2022-03-17
sdk: Set chainId for tx payload when sending

miguelmota:
- [x] set `chainId` when sending transaction based on method passed in chain parameter (ie `send` on HopBridge) or chain property on class instance (ie `Token`, `AMM`)


miguelmota:
Ready for review

packages/sdk/src/Base.ts
```diff
@@ -164,6 +164,11 @@ class Base {
     }
   }
 
+  sendTransaction (transactionRequest: any, chain: TChain) {
```
kangarang: minor suggestion: set type to ethers `TransactionRequest`

```suggestion
  sendTransaction (transactionRequest: TransactionRequest, chain: TChain) {
```

kangarang: or `transactionRequest: TransactionRequest | ContractTransaction`

kangarang (approved):
other than the 1 comment, LGTM

packages/sdk/src/Base.ts
miguelmota: thanks :+1: 

Resolved by https://github.com/hop-protocol/hop/pull/412/commits/1b993a50ec48a897213310a8e03f7a0f08c6c058

Issue #409, opened 2022-03-15
Limit the bonder’s ability to bond transfers whose `amount` is greater than the expected hToken amount on the source chain


shanefontaine:
The bonder should not bond a transfer that sends more hTokens than have been deposited into the source chain. This would protect against the case where a chain has a bug and a malicious user was able to mint ETH/Tokens/etc. out of thin air, among other things.

An example is an attack where an attacker can somehow mint hTokens unexpectedly and try to send them to a chain. For example, imagine there are 1,500 hETH on Optimism at this time. The bonder software should not bond any transfer that is ≥1,500 ETH on any chain when the source is Optimism because, if they did, that would mean that  the attacker was able to mint hTokens unexpectedly.

The implementation can likely use the chain balance and pending amount from the chain.

miguelmota:
This old [PR](https://github.com/hop-protocol/hop/pull/413) could use a review to handle this https://github.com/hop-protocol/hop/pull/413.

shanefontaine:
Handled with new redundant logic.

PR #413, opened 2022-03-18
hop-node: BondWithdrawalWatcher: Check bond amount doesn't exceed available amount on source chain. #409

miguelmota:
- [x] checks amount to bond doesn't exceed amount available on L1SourceChainBalance+PendingAmount
- [x] send notification


packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
```diff
@@ -185,6 +185,17 @@ class BondWithdrawalWatcher extends BaseWatcher {
         throw new BonderFeeTooLowError(msg)
       }
 
+      const hToken = await sourceL2Bridge.hToken()
+      const sourceChainTotalHTokens = await hToken.getTotalSupply()
+      const pendingAmount = await sourceL2Bridge.getPendingAmountForChainId(destinationChainId!)
```
kangarang: non-blocking: is `destinationChainId` *always* defined at this point? if so, we should improve the various `Transfer` types in `TransfersDb.ts`

packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
```diff
@@ -185,6 +185,17 @@ class BondWithdrawalWatcher extends BaseWatcher {
...
+      const hTokensAvailable = sourceChainTotalHTokens.add(pendingAmount)
+      if (amount!.gt(hTokensAvailable)) {
```
kangarang: same comment as above -- if `amount` is defined here, our types could be improved. again, this is non-blocking

kangarang (approved):
LGTM

packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
miguelmota: good point, the db getters can be updated to return specific types. I'll address this in another PR since the changes will affect many other places

miguelmota:
@shanefontaine going to merge this if you think it looks good

shanefontaine:
@miguelmota I think this would not work in the case where a malicious person can mint unlimited hTokens (i.e. if the chain has a bug in it).

Because of this, we would need to check the `chainBalance` from L1 along with the pending amount, and possibly other items as well.

miguelmota:
👍 checking L1 balance makes sense now

I was confused about what chainBalance meant; I see that it's a public getter on the contract

miguelmota:
Ready for review

packages/hop-node/src/watchers/BondWithdrawalWatcher.ts
```diff
@@ -186,6 +187,19 @@ class BondWithdrawalWatcher extends BaseWatcher {
         throw new BonderFeeTooLowError(msg)
       }
 
+      const l1Bridge = this.getSiblingWatcherByChainSlug(Chain.Ethereum).bridge as L1Bridge
+      const sourceChainTotalHTokens = await l1Bridge.getChainBalance(sourceChainId)
+      const pendingAmount = await sourceL2Bridge.getPendingAmountForChainId(destinationChainId)
+      const hTokensAvailable = sourceChainTotalHTokens.add(pendingAmount)
+      const sourceChain = this.chainIdToSlug(sourceChainId)
+      const isFromNonRollup = [Chain.Polygon, Chain.Gnosis].includes(sourceChain)
```
shanefontaine, line 195: This should apply to every chain, so the concept of `isFromNonRollup` is unnecessary here.

shanefontaine (changes requested):
I believe this does not track in-flight items that are neither part of `chainBalance` nor `pendingAmount`. Do you remember talking about this and if there was any conclusion here?

miguelmota:
Closing as stale

PR #414, opened 2022-03-21
Feature/low priority signer

miguelmota:
- [x] add `BONDER_LOW_PRIORITY_PRIVATE_KEY` env var to use for low priority signer
- [x] add `LOW_PRIORITY_TIME_TIL_BOOST_MS` to set when the signer should boost tx. This will be 1hr, meaning signer will put transaction through the boosting process if it has been pending in mempool for more than that time.
- [x] add etherchain gas estimation api to get estimated gas to use for low priority transactions, fallback to etherscan gas estimation api
- [x] update L1 signers in exit watcher to use low priority signer for exit transactions
- [x] if no low priority signer key is found, then it uses regular signer as usual


cwhinfrey (approved):
This all looks great to me!

Might be good to talk through some of the edge cases on standup tomorrow but all this code looks perfect.

packages/hop-node/src/gasboost/GasBoostTransaction.ts
```diff
@@ -98,7 +100,7 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
   reorgWaitConfirmations: number = 1
   originalTxParams: providers.TransactionRequest
 
-  type?: number
+  type?: number // type 0 or type 2
```
kangarang: could we make this type stricter?

```suggestion
  type?: 0 | 2
```

miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/414/commits/1cb67b11caae1c4deae52a0c15ef2734bdfe58eb

miguelmota:
Ready for review

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -459,10 +460,11 @@ class TransferRootsDb extends BaseDb {
 
       let timestampOk = true
       if (item.sentConfirmTxAt) {
-        timestampOk =
-          item.sentConfirmTxAt + TxRetryDelayMs < Date.now()
+        timestampOk = item.sentConfirmTxAt + TxRetryDelayMs < Date.now()
       }
 
+      const txStatusOk = item.confirmTxStatus !== TxStatus.Broadcasted
```
shanefontaine: I think this should technically be `item.confirmTxStatus === TxStatus.Incomplete`. This is because we don't want to re-send if it is broadcasted and we don't want to re-send if it is completed.

packages/hop-node/src/watchers/ArbitrumBridgeWatcher.ts
```diff
@@ -81,27 +82,8 @@ class ArbitrumBridgeWatcher extends BaseWatcher {
...
-    const tx = await this.relayXDomainMessage(commitTxHash)
-    if (!tx) {
-      logger.warn(`No tx exists for exit, commitTxHash ${commitTxHash}`)
-      return
-    }
-
-    const msg = `sent chain ${this.bridge.chainId} confirmTransferRoot exit tx ${tx.hash}`
-    logger.info(msg)
-    this.notifier.info(msg)
+  async isCheckpointed (l2BlockNumber: number) {
```
shanefontaine: Can this be removed? I think it may be misleading when exits are performed on Arbitrum. Arbitrum exits have been sensitive ([see here](https://github.com/hop-protocol/hop/blob/af7858918bff9b2cb4edd5d6f8fd8e8366d0c31b/packages/hop-node/src/db/TransferRootsDb.ts#L526)), so I think we should be explicit about the checkpoint state of these txs, if possible.

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -62,9 +62,24 @@ class SyncWatcher extends BaseWatcher {
...
+
     this.ready = true
   }
 
+  async updateBroadcastedStatus () {
+    const entries = await this.db.transferRoots.getTransferRootsFromTwoWeeks()
+    for (const entry of entries) {
+      if (entry.confirmTxStatus === TxStatus.Broadcasted) {
+        entry.confirmTxStatus = TxStatus.Incomplete
+        if (entry.confirmed) {
```
shanefontaine: I believe that if `entry.confirmed` is true, then the `entry.confirmTxStatus` would not matter, since it is implied that it is `Completed`. Because of this, I think this line should not exist within the `Broadcasted` check. 

This is a little tricky because we now have two states that represent the same thing (`entry.confirmTxStatus` (when `TxStatus.Brodcasted` is `true`) and `entry.confirmed`). Either way, if the entry has been marked `confirmed`, we know it should 100% be on-chain and completed. Because of this, maybe:

```
      if (!entry.confirmed) {
        if (entry.confirmTxStatus === TxStatus.Broadcasted) {
            entry.confirmTxStatus = TxStatus.Incomplete
        }
    } else {
      entry.confirmTxStatus = TxStatus.complete
    }
```

Open to suggestions here.

shanefontaine (commented):
* I was under the impression that a low-priority tx that is boosted would wait again for the low-priority time for each subsequent boost. I believe that is not the case here. Is there a reason we boost it like normal after the initial tx does not go through?
* Is the reason for 4 diff `gasPrice` providers because there are only 2 that support 1559 and 2 that support non-1559? If so, that is unfortunate because I don't like the thought of relying on so many diff services, but I do believe the fallback architecture is solid 👍 
* This should work on settlements as well. If that is a big change, I'm fine with that being in another PR in order to get this merged and running in prod. However, is this code "modular" in that sense that it can be applied to any watcher?
* Piggybacking off the prior comment, L2s should be supported as well because of the settlements. Is that also possible with these changes? I'd imagine the gas estimation services might make this a nightmare.

packages/hop-node/src/watchers/ArbitrumBridgeWatcher.ts
miguelmota: yes it needs to be removed, it's an artifact and shouldn't have been committed

packages/hop-node/src/db/TransferRootsDb.ts
miguelmota: yes you're correct

packages/hop-node/src/watchers/SyncWatcher.ts
miguelmota: I think that makes sense 

Resolved with https://github.com/hop-protocol/hop/pull/414/commits/edf2a5f22cf8200f1f08ca51703cfa4568f0802b

packages/hop-node/src/watchers/ArbitrumBridgeWatcher.ts
miguelmota: Resolved by https://github.com/hop-protocol/hop/commit/7ccc7dbf7886280c4b110aa0717ee7e3a21488c0

packages/hop-node/src/db/TransferRootsDb.ts
miguelmota: Resolved by https://github.com/hop-protocol/hop/commit/7ccc7dbf7886280c4b110aa0717ee7e3a21488c0

miguelmota:
> I was under the impression that a low-priority tx that is boosted would wait again for the low-priority time for each subsequent boost. I believe that is not the case here. Is there a reason we boost it like normal after the initial tx does not go through?

This is to make the sure tx the eventually goes through (as discussed a while ago)

> Is the reason for 4 diff gasPrice providers because there are only 2 that support 1559 and 2 that support non-1559? If so, that is unfortunate because I don't like the thought of relying on so many diff services, but I do believe the fallback architecture is solid +1

The blocknative and alchemy gas apis (1559 only) have been removed to simplify the logic. It'll use etherchain api and fallback to etherscan for getting slow gasPrice which will then be used to set the maxFeePerGas value

> This should work on settlements as well. If that is a big change, I'm fine with that being in another PR in order to get this merged and running in prod. However, is this code "modular" in that sense that it can be applied to any watcher?

I think all that's needed is to connect the low priority signer before sending the particular transaction. For the mvp it'll only be exit txs but it'll be trivial to use low priority signer for other types of transactions

```ts
const tx = await this.bridgeContract.connect(wallets.lowPriority.get(this.chain)).settleBondedWithdrawals(...)
```

> Piggybacking off the prior comment, L2s should be supported as well because of the settlements. Is that also possible with these changes? I'd imagine the gas estimation services might make this a nightmare.

only L1 txs will be supported for the mvp but should be trivial to support L2s as long as apis exists

miguelmota:
Ready for review

miguelmota:
Closing as stale

PR #415, opened 2022-03-22
Feature/hop node db getter return types

miguelmota:
- [x] add explicit db getter return types for TransfersDb and TransferRootsDb to avoid using `!`  non-null assertion operator when reading properties


miguelmota:
@shanefontaine going to merge this but let me know if you have any questions

shanefontaine:
@miguelmota SGTM. No questions and sorry for late review 👍

PR #420, opened 2022-03-29
UI: fix stats page

kangarang:
this pr improves the stats page in some ways:

- data population: previously, the pending amounts stats section would intermittently not display optimism and arbitrum -> L1 rows. this is now resolved by ensuring the supported Chain models have a `chainId: number` field
- rm extraneous code
- visual design styles
- added @types/react-table
- fixed bug in `fetchPendingAmounts`: check both source and destination chains for unsupported assets

### DIFF
--- a/packages/frontend/src/models/Network.ts
+++ b/packages/frontend/src/models/Network.ts
@@ -8,9 +8,11 @@ export type NetworkProps = {
   slug: string
   imageUrl: string
   rpcUrl: string
-  networkId: string
+  networkId: number
+  chainId?: number
   nativeTokenSymbol: string
   isLayer1?: boolean
+  isL1?: boolean
   nativeBridgeUrl?: string
   waitConfirmations?: number
   explorerUrl: string
@@ -22,9 +24,11 @@ class Network {
   readonly imageUrl: string
   readonly provider: ethers.providers.Provider
   readonly rpcUrl: string
-  readonly networkId: string
+  readonly networkId: number
+  readonly chainId: number
   readonly nativeTokenSymbol: string
   readonly isLayer1: boolean
+  readonly isL1: boolean
   readonly nativeBridgeUrl: string | undefined
   readonly waitConfirmations?: number
   readonly explorerUrl: string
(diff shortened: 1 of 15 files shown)


packages/frontend/src/models/Network.ts
```diff
@@ -22,9 +24,11 @@ class Network {
   readonly imageUrl: string
   readonly provider: ethers.providers.Provider
   readonly rpcUrl: string
-  readonly networkId: string
+  readonly networkId: number
+  readonly chainId: number
```
miguelmota, line 28: what's the difference between `networkId` and `chainId` here? I think `networkId` should be removed (assuming everywhere it's being used in our packages has been renamed to use the new chainId property)

kangarang: this is intermediary, just to avoid any issues where networkId may be used. eventually we should consolidate `Network` and `Chain` models.

packages/sdk/src/Base.ts
```diff
@@ -549,12 +549,22 @@ class Base {
 
   getSupportedAssets () {
     const supported : any = {}
+
+    const unsupportedAssets: any = {
```
miguelmota: i think ideally this mapping is constructed dynamically based on the config tokens

packages/sdk/src/models/Chain.ts
```diff
@@ -12,11 +12,11 @@ class Chain {
   isL1: boolean = false
   nativeTokenSymbol: string
 
-  static Ethereum = newChain(ChainSlug.Ethereum)
-  static Optimism = newChain(ChainSlug.Optimism)
-  static Arbitrum = newChain(ChainSlug.Arbitrum)
-  static Gnosis = newChain(ChainSlug.Gnosis)
-  static Polygon = newChain(ChainSlug.Polygon)
+  static Ethereum = newChain(ChainSlug.Ethereum, metadata.networks.ethereum.chainId)
```
miguelmota: why not read the chain id from `packages/config/networks/mainnet/`? (instead of re-adding it to metadata file)

packages/sdk/src/Base.ts
kangarang: you're right, this was an initial approach i took, but is not necessary. will revert. thanks for the catch!

miguelmota (approved):
LGTM 

nice clean up

PR #424, opened 2022-04-20
sdk: set manual polygon gasLimit in tx overrides

kangarang:


packages/sdk/src/Base.ts
```diff
@@ -495,6 +495,7 @@ class Base {
       if (txOptions.gasPrice?.lt(MinPolygonGasPrice)) {
         txOptions.gasPrice = BigNumber.from(MinPolygonGasPrice)
       }
+      txOptions.gasLimit = BigNumber.from(300e3)
```
miguelmota: I think this value should be a constant `PolygonMinGasLimit`

Also are we certain that all polygon transaction won't exceed 300k gas limit? I had originally thought there was going to be an additional buffer added to the estimated gas limit but I know that may be more complicated so if we know that 300k gas limit is enough for all polygon txs then we can merge

packages/sdk/src/Base.ts
```diff
@@ -163,8 +163,16 @@ class Base {
...
+    chain = this.toChainModel(chain)
+    if (chain.equals(Chain.Polygon)) {
+      if (
+        !transactionRequest.gasLimit ||
+        BigNumber.from(transactionRequest.gasLimit)?.lt(MinPolygonGasLimit)
+      ) {
+        transactionRequest.gasLimit = MinPolygonGasLimit
+      }
+    }
+    return this.signer.sendTransaction({ ...transactionRequest, chainId: chain.chainId } as any)
```
kangarang: @miguelmota how do you feel about setting the gasLimit here? (this change comes from the last commit)

miguelmota: I think it made more sense in txOverrides because calling the sdk populate methods uses txOverrides as well

miguelmota: I think how you had it last but increase MinPolygonGasLimit to maybe 500k just to be on the safe side

packages/frontend/src/hooks/useApprove.ts
```diff
@@ -61,12 +61,13 @@ const useApprove = token => {
...
-          }
-        }
+          },
+        },
       },
       onConfirm: async (approveAll: boolean) => {
+        const overrides = await token.txOverrides(chain)
         const approveAmount = approveAll ? constants.MaxUint256 : amount
-        return token.approve(spender, approveAmount)
+        return token.approve(spender, approveAmount, overrides)
```
miguelmota: The overrides parameter here in token approve shouldn't be necessary since the sdk token approve function calls populateApproveTx which already includes the overrides. I think the overrides parameter can be removed because of that reason

packages/sdk/src/Base.ts
```diff
@@ -162,9 +162,9 @@ class Base {
     }
   }
 
-  sendTransaction (transactionRequest: providers.TransactionRequest, chain: TChain) {
+  sendTransaction (transactionRequest: providers.TransactionRequest, chain: TChain, overrides?: any) {
```
miguelmota: it seems redundant to have an overrides parameter here because the overrides should already be includes in the transactionRequest object. I think it can be removed

miguelmota:
lgtm

PR #434, opened 2022-05-24
Feat/exit multiple tx hashes

shanefontaine:
@miguelmota Only downside of this is that it fires off a tx per exit. However, I think the simplicity of that (as opposed to trying to batch many exits in a single tx) is worth it.

### DIFF
--- a/packages/hop-node/src/cli/exitCommitTx.ts
+++ b/packages/hop-node/src/cli/exitCommitTx.ts
@@ -21,7 +21,7 @@ root
   .action(actionHandler(main))
 
 async function main (source: any) {
-  const { chain, token, txHash: commitTxHash, txHashes: commitTxHashes, dry: dryMode } = source
+  const { chain, token, txHashes: commitTxHashes, dry: dryMode } = source
   if (!chain) {
     throw new Error('chain is required')
   }
--- a/packages/hop-node/src/cli/exitCommitTx.ts
+++ b/packages/hop-node/src/cli/exitCommitTx.ts
@@ -12,7 +12,7 @@ root
   .description('Exit the commit transaction')
   .option('--chain <slug>', 'Chain', parseString)
   .option('--token <symbol>', 'Token', parseString)
-  .option('--tx-hash <hash>', 'Tx hash with CommitTransfers event log', parseString)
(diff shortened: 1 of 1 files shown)


packages/hop-node/src/cli/exitCommitTx.ts
```diff
@@ -12,7 +12,7 @@ root
   .description('Exit the commit transaction')
   .option('--chain <slug>', 'Chain', parseString)
   .option('--token <symbol>', 'Token', parseString)
-  .option('--tx-hash <hash>', 'Tx hash with CommitTransfers event log', parseString)
+  .option('--tx-hashes <hash>', 'Comma-separated tx hashes with CommitTransfers event log', parseString)
```
miguelmota, line 15: current

```
<hash>
```

suggestion

```
<hash, ...>
```

---

current

```ts
parseString
```

suggestion

```ts
parseStringArray
```

and add this to `shared.ts`

```ts
export function parseStringArray (value: string) {
  return (value ?? '').split(',').map((v: string) => v.trim())
}
```

miguelmota: @shanefontaine

packages/hop-node/src/cli/exitCommitTx.ts
```diff
@@ -38,6 +38,9 @@ async function main (source: any) {
   }
 
   const chainSpecificWatcher: ExitWatcher = watcher.watchers[chain]
-  await chainSpecificWatcher.relayXDomainMessage(commitTxHash)
+  const txHashes = commitTxHashes.split(',')
```
miguelmota, line 41: with above suggestion, `commitTxHashes` would be an array

packages/hop-node/src/cli/exitCommitTx.ts
shanefontaine: Resolved in 62d3ebb08a4d703cac759d6ab80f7d9dfdabfeac

packages/hop-node/src/cli/exitCommitTx.ts
shanefontaine: Resolved in 9ee3ae58c72894d53f400a51afb8db5cc4f721f8

PR #436, opened 2022-05-30
hop-node: Rebroadcast latest inflight item if it can't be boosted

miguelmota:


packages/hop-node/src/gasboost/GasBoostTransaction.ts
```diff
@@ -569,20 +569,28 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
...
+    const isLatestItem = item === this.getLatestInflightItem()
+    const shouldRebroadcast = isLatestItem && this.maxGasPriceReached
+    if (shouldRebroadcast) {
+      try {
+        await this.rebroadcastLatestTx()
+      } catch (err) {
+        const { isAlreadyKnown } = this.parseErrorString(err.message)
+        if (!isAlreadyKnown) {
+          this.logger.error('rebroadcastLatestTx error:', err)
+        }
```
shanefontaine: It might make sense to add a log here for debugging purposes. If the mempool is not behaving as we are assuming (as we've seen through various errors on various chains), it would be nice to see what error is being thrown at the bonder.

shanefontaine (commented):
@miguelmota I think this will re-broadcast a maxed tx every 10s. Is that the intention? If this is true, then I _think_ this could get pretty RPC-call heavy if, for example, gas spikes to 1k GWei for 1+ hour and we are re-calling `sendTransaction()` every 10s per Tx. I think ideally it rebroadcasts at the same cadence as the boosting (3 min, I think). This is a loose opinion.

IMO logic looks good 👍

packages/hop-node/src/gasboost/GasBoostTransaction.ts
miguelmota: should log all errors [now](https://github.com/hop-protocol/hop/pull/436/commits/7e3c6438101db68fe041558578fc23b73acb8c31)

miguelmota:
@shanefontaine good catch, it's been [updated](https://github.com/hop-protocol/hop/pull/436/commits/7e3c6438101db68fe041558578fc23b73acb8c31#diff-be5791e55b6353175456441175c8d86f188ce21871bbf9571f7e0d8de89a3351R591) to rebroadcast based on `timeTilBoostMs`

shanefontaine (approved):
@miguelmota LGTM 👍

PR #437, opened 2022-06-14
hop-node: GasBoost: Make sure maxFeePerGas is always higher than baseFeePerGas

miguelmota:
- Makes sure that baseFeePerGas is not greater than maxFeePerGas
  - if baseFeePerGas is greater than maxFeePerGas, then maxFeePerGas is baseFeePerGas*2
    - if new maxFeePerGas is greater than maxGasPrice, then maxFeePerGas is maxGasPrice

### DIFF
--- a/packages/hop-node/src/gasboost/GasBoostTransaction.ts
+++ b/packages/hop-node/src/gasboost/GasBoostTransaction.ts
@@ -418,8 +418,7 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
       const currentBaseFeePerGas = await this.getCurrentBaseFeePerGas()
       const maxGasPrice = this.getMaxGasPrice()
       if (maxFeePerGas.lte(currentBaseFeePerGas)) {
-        const newMaxFeePerGas = currentBaseFeePerGas.mul(2)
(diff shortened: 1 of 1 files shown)


shanefontaine:
@miguelmota is the case where `baseFeePerGas > maxGasPrice` handled? I'm thinking of the case where we have a 500 gwei max and `baseFee` is spiking to 500+.

miguelmota:
@shanefontaine good point, PR has been updated to always clamp maxFeePerGas to maxGasPrice if it's over

shanefontaine:
@miguelmota I think it looks good. One more clarification: does `getCurrentBaseFeePerGas()` return the `baseFee` if the chain supports 1559 and `gasPrice` if it does not?

miguelmota:
@shanefontaine if the chain supports 1559 then `getCurrentBaseFeePerGas` will always return the `baseFeePerGas` and `getCurrentBaseFeePerGas` is only called if the transaction type is 1559 so there shouldn't be an instance where getCurrentBaseFeePerGas is called and the tx type is not 1559

shanefontaine (approved):
👍 LGTM

PR #443, opened 2022-07-06
sdk: Check if transfer will cause bond withdrawal to fail

miguelmota:
checks before sending and throws error if bond withdrawal will fail at the destination


shanefontaine (changes requested):
@miguelmota logic looks good. I think this will only handle L2 -> any. I think `willTransferFail()` needs to generate `populatedTx` differently when the source is L1 and `willTransferFail()` needs to  be added the the L1 -> L2 sends as well.

shanefontaine (approved):
👍 LGTM

PR #444, opened 2022-07-08
hop-node: Commit transfers if pending count threshold is reached on polygon

miguelmota:


shanefontaine (changes requested):
* I think `pendingCountOk` should default to `false` or else the poller will commit on every poll on any chain other than Polygon
* nitpick: add a log within the `if (!canCommit)` scope that just logs the `pendingCountOk` just so we can explicitly search it in the logs

PR #446, opened 2022-07-21
Feature/aave vault

miguelmota:
Adds Aave vault strategy for deposit/withdrawal

supports USDC, USDT, DAI, ETH on arbitrum and ethereum, and MATIC on polygon


packages/hop-node/src/vault/Vault.ts
```diff
@@ -9,6 +10,13 @@ export abstract class Vault {
     if (useYearnVault) {
       return new YearnVault(chain, token, signer)
     }
+
+    // TODO: determine what tokens and chain to use aave vault for
+    const aaveVaultTokens = new Set(['USDC'])
+    const useAaveVault = chain === Chain.Arbitrum && aaveVaultTokens.has(token)
+    if (useAaveVault) {
+      return new AaveVault(chain, token, chain)
```
shanefontaine: The last param looks like a duplicate and should be `signer`. It will be the GasBoost signer, correct?

packages/hop-node/src/vault/AaveVault.ts
```diff
@@ -0,0 +1,293 @@
...
+    if (!aaveAddresses[chain]) {
+      throw new Error('chain is not supported')
+    }
+    this.chain = chain
+    this.token = token
+    this.tokenAddress = tokenAddresses[token][chain].token
+    this.aTokenAddress = tokenAddresses[token][chain].aToken
+    this.provider = getRpcProvider(chain)!
+    this.signer = signer.connect(this.provider)
+    this.decimals = getTokenDecimals(token)
```
shanefontaine, line 101: It looks like it does not matter, but do any of the `this.decimal` use-cases (which look like the parsing/formatting), use the `aToken` instead of the canonical token? If so, should we store `this.aTokenDecimals`?

packages/hop-node/src/vault/AaveVault.ts
```diff
@@ -0,0 +1,293 @@
...
+      const tx = await this.approveWithdrawal()
+      console.log('approval tx:', tx.hash)
+      await tx.wait()
+    }
+
+    // aave v2 and v3
+    const txs = await this.pool.withdraw({
+      user: account,
+      reserve: this.tokenAddress,
+      amount: this.formatUnits(amount).toString(),
```
shanefontaine, line 206: Is this `amount` in terms of canonical token or a token? If it is a token, `formatUnits` will need to reflect the a token `decimals` (mentioned in another comment but flagging it here as well).

shanefontaine (changes requested):
Looks good. One comment is that the approved address (`LENDING_POOL`) is an upgradeable contract, so unlimited approvals seem not-ideal. With that said, I believe the entire chain has huge issues if that address is upgraded maliciously.

packages/hop-node/src/vault/Vault.ts
miguelmota: nice catch, fixed

packages/hop-node/src/vault/AaveVault.ts
miguelmota: the aave tokens should always have the same decimals as their canonical token (similar to how hTokens and canonical tokens have same decimals in hop)

packages/hop-node/src/vault/AaveVault.ts
miguelmota: The decimals will always be the same for both canonical token and their aave token (ie USDC and aUSDC both have 6 decimals)

PR #448, opened 2022-08-06
hop-node: optimize health check rpc calls

shanefontaine:
Just a small RPC optimization

### DIFF
--- a/packages/hop-node/src/constants/constants.ts
+++ b/packages/hop-node/src/constants/constants.ts
@@ -48,7 +48,7 @@ export const SettlementGasLimitPerTx: Record<string, number> = {
   arbitrum: 59105
 }
 
-const SecondsInDay = 86400
+export const SecondsInDay = 86400
 const SecondsInWeek = SecondsInDay * 7
 export const OneDayMs = SecondsInDay * 1000
 export const TotalBlocks = {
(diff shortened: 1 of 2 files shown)


packages/hop-node/src/watchers/HealthCheckWatcher.ts
```diff
@@ -675,12 +675,18 @@ export class HealthCheckWatcher {
...
+    const challengeTimeWithBuffer = SecondsInDay * 1.25
+    const blocksInADay = challengeTimeWithBuffer / mainnetBlockTimeSec
+
     for (const token of this.tokens) {
       this.logger.debug(`done ${token} bridge for challenged roots`)
       const l1BridgeContract = contracts.get(token, Chain.Ethereum).l1Bridge
-      const provider = getRpcProvider(Chain.Ethereum)!
-      const startBlockNumber = 0
       const endBlockNumber = Number((await provider.getBlockNumber()).toString())
+      const startBlockNumber = endBlockNumber - blocksInADay
```
miguelmota: Why not use `getBlockNumberFromDate` util to get block number from 24 hours ago?

packages/hop-node/src/constants/constants.ts
```diff
@@ -48,7 +48,7 @@ export const SettlementGasLimitPerTx: Record<string, number> = {
   arbitrum: 59105
 }
 
-const SecondsInDay = 86400
+export const SecondsInDay = 86400
```
miguelmota: small suggestion on naming: `SecondsInDay` -> `OneDaySeconds` (since we use `OneHourMs`, `OneHourSeconds`, etc)

packages/hop-node/src/watchers/HealthCheckWatcher.ts
shanefontaine: Updated in 3b3f8a3c0464023e8ac6bd8ec3a023029a169228

packages/hop-node/src/constants/constants.ts
shanefontaine: Updated in 3b3f8a3c0464023e8ac6bd8ec3a023029a169228

PR #450, opened 2022-08-09
Feature/hop

miguelmota:
- Add HOP metadata
- TODO: Add HOP bridge addresses


shanefontaine:
@miguelmota This PR is ready for review. Besides your updates, the biggest change is the lack of AMM wrapper. The entire app appears to work well with these changes (SDK/FE & bonder).

One known issue is this persistent error that I cannot remove for some reason. It is meant to only show when "via AMM" is selected, but it appears as an error when "via Hop Bridge" is selected as well. Can you see if you understand why that might be?

https://user-images.githubusercontent.com/9441295/191355233-10e89160-e89e-4bf4-8fed-dae53ec7d4ab.mov

packages/sdk/src/HopBridge.ts
```diff
@@ -1898,9 +1914,9 @@ class HopBridge extends Base {
     }
     recipient = checksumAddress(recipient)
 
-    let ammWrapper = await this.getAmmWrapper(sourceChain, sourceChain.provider)
-    let l2Bridge = await this.getL2Bridge(sourceChain, sourceChain.provider)
-    const attemptSwapAtSource = this.shouldAttemptSwap(amountOutMin, deadline)
+    const ammWrapper = await this.getAmmWrapper(sourceChain, sourceChain.provider)
+    const l2Bridge = await this.getL2Bridge(sourceChain, sourceChain.provider)
+    const attemptSwapAtSource = this.doesUseAmm && this.shouldAttemptSwap(amountOutMin, deadline)
```
miguelmota: I think the `shouldAttemptSwap` method should return false if it uses amm so changing this back to

```suggestion
    const attemptSwapAtSource = this.shouldAttemptSwap(amountOutMin, deadline)
```

 and adding the check here

```ts
  shouldAttemptSwap (amountOutMin: BigNumber, deadline: BigNumberish): boolean {
    if (!this.doesUseAmm) {
       return false
    }
    deadline = BigNumber.from(deadline?.toString() || 0)
    return amountOutMin?.gt(0) || deadline?.gt(0)
  }
```

packages/sdk/src/HopBridge.ts
```diff
@@ -1994,15 +2008,17 @@ class HopBridge extends Base {
     }
     recipient = checksumAddress(recipient)
 
-    let ammWrapper = await this.getAmmWrapper(sourceChain, sourceChain.provider)
+    const ammWrapper = await this.getAmmWrapper(sourceChain, sourceChain.provider)
+    const l2Bridge = await this.getL2Bridge(sourceChain, sourceChain.provider)
+    const attemptSwapAtSource = this.doesUseAmm && this.shouldAttemptSwap(amountOutMin, deadline)
```
miguelmota: same comment as above

```suggestion
    const attemptSwapAtSource = this.shouldAttemptSwap(amountOutMin, deadline)
```

miguelmota (commented):
The changes look really great 

Left one minor suggestion

miguelmota:
> It is meant to only show when "via AMM" is selected, but it appears as an error when "via Hop Bridge" is selected as well

I think the error just needs to be reset when switching convertOption in ConvertContext (ie `setError('')`)

packages/sdk/src/HopBridge.ts
shanefontaine: Resolved with 56d91bae233fc3fdecafbecb07dac35a5db04144

packages/sdk/src/HopBridge.ts
shanefontaine: Resolved with 56d91bae233fc3fdecafbecb07dac35a5db04144

PR #451, opened 2022-08-12
Feat/redemption watcher

shanefontaine:


packages/hop-node/src/cli/dbDump.ts
```diff
@@ -137,8 +137,9 @@ async function main (source: any) {
         throw new Error('chain flag is required')
       }
       items = [
```
miguelmota: suggestion: a Promise.all here would make sense

```ts
items = await Promise.all([
  db.gasCost.getNearest(chain, tokenSymbol, GasCostTransactionType.BondWithdrawal, nearest),
  db.gasCost.getNearest(chain, tokenSymbol, GasCostTransactionType.BondWithdrawalAndAttemptSwap, nearest),
  db.gasCost.getNearest(chain, tokenSymbol, GasCostTransactionType.Relay, nearest)
])
```

packages/hop-node/src/config/config.ts
```diff
@@ -57,6 +58,11 @@ export const defaultKeystoreFilePath = `${defaultConfigDir}/keystore.json`
 export const minEthBonderFeeBn = parseEther('0.00001')
 export const pendingCountCommitThreshold = 256
 
+// TODO: Remove this post-nitro
+export const isNitroLive = process.env.IS_NITRO_LIVE ?? false
+// TODO: Remove this post-nitro
+export const nitroStartTimestamp = process.env.NITRO_START_TIMESTAMP ?? 0
```
miguelmota: Use the `normalizeEnvVarNumber` util here

```ts
export const nitroStartTimestamp = normalizeEnvVarNumber(process.env.NITRO_START_TIMESTAMP ?? 0)
```

packages/hop-node/src/db/GasCostDb.ts
```diff
@@ -43,12 +44,30 @@ class GasCostDb extends BaseDb {
...
+      delete entry.value.attemptSwap
+      await this.addGasCost(entry.value)
+    }
+  }
+
   async update (key: string, data: GasCost) {
     return this._update(key, data)
   }
 
   async addGasCost (data: GasCost) {
```
miguelmota: suggestion:  rename `addGasCost` to `updateGasCost` or just `update` to match the other db classes

packages/hop-node/src/db/GasCostDb.ts
```diff
@@ -43,12 +44,30 @@ class GasCostDb extends BaseDb {
     }
   }
 
+  async migration () {
+    this.logger.debug('GasCostDb migration started')
+    const entries = await this.getKeyValues()
+    this.logger.debug(`GasCostDb migration: ${entries.length} entries`)
+    for (const entry of entries) {
+      let transactionType: GasCostTransactionType
```
miguelmota, line 56: I think there should be another check here so it doesn't re-set values on re-runs

```ts
if (typeof entry.value.transactionType !== 'undefined') {
 continue
}
```

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -545,6 +563,74 @@ class TransferRootsDb extends BaseDb {
...
+
+      const isSeenOnL1 = item?.bonded ?? item?.confirmed
+
+      let sentTxTimestampOk = true
+      if (item.sentRelayTxAt) {
+        sentTxTimestampOk = item.sentRelayTxAt + TxRetryDelayMs < Date.now()
+      }
+
+      // bondedAt should be checked first because a root can have both but it should be bonded prior to being confirmed
+      const seenOnL1Timestamp = item?.bondedAt ?? item?.confirmedAt
```
miguelmota, line 619: I think a check for this values should be done to return early if they don't exists

```ts
if (!(item?.bondedAt || item?.confirmedAt)) {
  return false
}
```

otherwise I think it's possible `seenOnL1Timestamp` can be `undefined`

packages/hop-node/src/db/TransfersDb.ts
```diff
@@ -438,6 +462,82 @@ class TransfersDb extends BaseDb {
...
+          if (delay > OneWeekMs) {
+            return false
+          }
+          timestampOk = item.relayAttemptedAt + delay < Date.now()
+        } else {
+          timestampOk = item.relayAttemptedAt + TxRetryDelayMs < Date.now()
+        }
+      }
+
+      const seenOnL1TimestampMs: number = item.transferSentTimestamp! * 1000
```
miguelmota: I think it should return early if `transferSentTimestamp` is undefined (incomplete item)

```ts
if (!item.transferSentTimestamp) {
  return false
}
```

packages/hop-node/src/theGraph/getTransferFromL1Completed.ts
```diff
@@ -0,0 +1,58 @@
...
+import { MaxInt32 } from 'src/constants'
+import { constants } from 'ethers'
+import { normalizeEntity } from './shared'
+
+export default async function getTransferFromL1Completed (
+  chain: string,
+  token: string,
+  startDate: number = 0,
+  endDate: number = MaxInt32,
+  lastId: string = constants.AddressZero
```
miguelmota: suggestion: I know we use `constants.AddressZero` in multiple places as default `lastId` but they can all be simplified by just using `"0"` (eg in [getBonds](https://github.com/hop-protocol/hop/blob/5c4613e4e7b8dd9e93ecd8ccc82ba64df720e139/packages/hop-node/src/theGraph/getInvalidBondWithdrawals.ts#L4))

```ts
lastId: string = '0'
```

since technically the `id` is not an address and it just needs to be a low string value

packages/hop-node/src/theGraph/getTransferFromL1Completed.ts
```diff
@@ -0,0 +1,58 @@
...
+) {
+  const query = `
+    query TransferFromL1Completed(${token ? '$token: String, ' : ''}$startDate: Int, $endDate: Int, $lastId: ID) {
```
miguelmota, line 13: suggestion: since the `token` parameter for the function is required then `$token` here should always be set (no ternary)


```gql
    query TransferFromL1Completed($token: String, startDate: Int, $endDate: Int, $lastId: ID) {
```

edit: i see this was expected behavior now as i was going down the list of files

packages/hop-node/src/theGraph/getTransferFromL1Completed.ts
```diff
@@ -0,0 +1,58 @@
...
+      transferFromL1Completeds(
+        where: {
+          ${token ? 'token: $token,' : ''}
```
miguelmota, line 16: same comment as above applies here

packages/hop-node/src/theGraph/getTransferRootBonded.ts
```diff
@@ -1,17 +1,22 @@
...
 export default async function getTransferRootBonded (
   chain: string,
   token: string,
+  startDate: number = 0,
+  endDate: number = MaxInt32,
   lastId: string = constants.AddressZero
 ) {
   const query = `
-    query TransferRootBonded($token: String, $lastId: ID) {
+    query TransferRootBonded(${token ? '$token: String, ' : ''}$startDate: Int, $endDate: Int, $lastId: ID) {
```
miguelmota, line 13: suggestion: since the `token` parameter for the function is required then `$token` here should always be set (no ternary)

edit: i see this was expected behavior now as i was going down the list of files

packages/hop-node/src/theGraph/getTransferRootBonded.ts
```diff
@@ -1,17 +1,22 @@
...
       transferRootBondeds(
         where: {
-          token: $token
+          ${token ? 'token: $token,' : ''}
```
miguelmota, line 16: same comment as above applies here

miguelmota (commented):
Some initial suggestions but still going through the PR and larger files

packages/hop-node/src/theGraph/getUnsetTransferRoots.ts
```diff
@@ -0,0 +1,40 @@
+import getTransferRootBonded from 'src/theGraph/getTransferRootBonded'
+import getTransferRootConfirmed from 'src/theGraph/getTransferRootConfirmed'
+import getTransferRootSet from 'src/theGraph/getTransferRootSet'
+import { Chain } from 'src/constants'
+import { getAllChains } from 'src/config'
+
+export default async function getUnsetTransferRoots (startDate: number, endDate: number) {
+  // An empty token address represents all tokens.
+  const token = ''
```
miguelmota: If the `token` parameter is optional then it should be set as optional and as the last parameter since it's not clear in the function signature that an empty string should be used

order of parameters can be `startDate`, `endDate`, `chain`, `token?` (optional)

packages/sdk/src/Base.ts
```diff
@@ -536,6 +536,19 @@ class Base {
     return bonder
   }
 
+  protected async _getMessengerWrapperAddress (token: TToken, destinationChain: TChain): Promise<string> {
+    await this.fetchConfigFromS3()
+    token = this.toTokenModel(token)
+    destinationChain = this.toChainModel(destinationChain)
+
+    const messengerWrapper = this.addresses?.[token.canonicalSymbol]?.[destinationChain.slug].l1MessengerWrapper
```
miguelmota: The optional operator `?.` before `.l1MessengerWrapper` should be used here

```ts
    [destinationChain.slug]?.l1MessengerWrapper
```

packages/hop-node/src/watchers/ArbitrumBridgeWatcher.ts
```diff
@@ -92,16 +90,34 @@ class ArbitrumBridgeWatcher extends BaseWatcher {
...
+    const txReceipt = await this.l1Wallet.provider.getTransactionReceipt(l1TxHash)
+    const l1TxnReceipt = new L1TransactionReceipt(txReceipt)
+    return l1TxnReceipt.getL1ToL2Messages(this.l2Wallet)
+  }
+
+  async isTransactionRedeemed (l1TxHash: string, messageIndex: number = 0): Promise<boolean> {
+    const l1ToL2Message = await this.getL1ToL2Message(l1TxHash, messageIndex)
+    const res = await l1ToL2Message.waitForStatus()
+    const status = res.status
+    return status === 4
```
miguelmota: suggestion: if arbitrum sdk provides an enum or map with status codes, that can be used instead for better readability (instead of using harcoded status code number)

packages/hop-node/src/watchers/HealthCheckWatcher.ts
```diff
@@ -129,6 +133,22 @@ type InvalidBondWithdrawal = {
   timestamp: number
 }
 
+type UnrelayedTransfers = {
```
miguelmota: This entity type should be named as singular and not plural: `UnrelayedTransfers` => `UnrelayedTransfer`

packages/hop-node/src/watchers/HealthCheckWatcher.ts
```diff
@@ -129,6 +133,22 @@ type InvalidBondWithdrawal = {
...
+  transactionHash: string
+  token: string
+  recipient: string
+  destinationChainId: number
+  amount: BigNumber
+  relayer: string
+  relayerFee: BigNumber
+}
+
+type UnsetTransferRoots = {
```
miguelmota: This entity type should be named as singular and not plural: `UnsetTransferRoots` => `UnsetTransferRoot`

packages/hop-node/src/watchers/HealthCheckWatcher.ts
```diff
@@ -129,6 +133,22 @@ type InvalidBondWithdrawal = {
   timestamp: number
 }
 
+type UnrelayedTransfers = {
+  transactionHash: string
+  token: string
+  recipient: string
+  destinationChainId: number
+  amount: BigNumber
```
miguelmota: If these entities are part of the health checker json output then the amount types should be `string` instead of `BigNumber` (as seen in the other entities)

packages/hop-node/src/watchers/HealthCheckWatcher.ts
```diff
@@ -129,6 +133,22 @@ type InvalidBondWithdrawal = {
...
+  relayer: string
+  relayerFee: BigNumber
+}
+
+type UnsetTransferRoots = {
+  transferRootHash: string
+  totalAmount: BigNumber
```
miguelmota: If these entities are part of the health checker json output then the amount types should be `string` instead of `BigNumber` (as seen in the other entities)

packages/hop-node/src/watchers/HealthCheckWatcher.ts
```diff
@@ -797,4 +841,75 @@ export class HealthCheckWatcher {
...
+    const now = DateTime.now().toUTC()
+    const endDate = now.minus({ hours: 1 })
+    const startDate = endDate.minus({ days: this.days })
+    const tokens = ''
+    const transfersSent = await getTransferSentToL2(Chain.Ethereum, tokens, Math.floor(startDate.toSeconds()), Math.floor(endDate.toSeconds()))
+
+    // There is no relayerFeeTooLow check here but there may need to be. If too many relayer fees are too low, then we can add logic to check for that.
+
+    const missingTransfers: any[] = []
+    const chains = [Chain.Arbitrum]
```
miguelmota: I think this variable can be replaced with the constant `relayableChains`

packages/hop-node/src/watchers/HealthCheckWatcher.ts
```diff
@@ -797,4 +841,75 @@ export class HealthCheckWatcher {
       }
     })
   }
+
+  async getUnrelayedTransfers (): Promise<UnrelayedTransfers[]> {
+    const now = DateTime.now().toUTC()
+    const endDate = now.minus({ hours: 1 })
+    const startDate = endDate.minus({ days: this.days })
+    const tokens = ''
```
miguelmota, line 849: same comment as above about making the token parameter be the last optional parameter

packages/hop-node/src/watchers/RelayWatcher.ts
```diff
@@ -0,0 +1,383 @@
...
+      dryMode: config.dryMode
+    })
+  }
+
+  async pollHandler () {
+    if (this.isL1) {
+      return
+    }
+
+    const promises: Array<Promise<any>> = []
```
miguelmota: suggestion: this can be simplified a bit to

```ts
    await Promise.all([
      this.checkTransferSentToL2FromDb(),
      this.checkRelayableTransferRootsFromDb()
    ])
```

packages/hop-node/src/watchers/RelayWatcher.ts
```diff
@@ -0,0 +1,383 @@
...
+        await this.checkTransferSentToL2(transferId)
+      } catch (err: any) {
+        logger.error('checkTransferSentToL2 error:', err)
+      }
+
+      logger.debug(`processing item ${i + 1}/${batchedDbTransfers.length} complete`)
+      logger.debug('db poll completed')
+    }, { concurrency: relayTransactionBatchSize, timeoutMs: 10 * 60 * 1000 })
+
+    this.logger.debug('checkTransferSentFromDb completed')
```
miguelmota: The log should say `checkTransferSentToL2FromDb` instead of `checkTransferSentFromDb`

packages/hop-node/src/watchers/RelayWatcher.ts
```diff
@@ -0,0 +1,383 @@
...
+        `checking ${dbTransferRoots.length} unrelayed transfer roots db items`
+    )
+
+    const promises: Array<Promise<any>> = []
+    for (const dbTransferRoot of dbTransferRoots) {
+      const { transferRootId } = dbTransferRoot
+      promises.push(this.checkRelayableTransferRoots(transferRootId))
+    }
+
+    await Promise.all(promises)
```
miguelmota, line 115: suggestion: a 'done' log here would be nice

packages/hop-node/src/watchers/RelayWatcher.ts
```diff
@@ -0,0 +1,383 @@
...
+    if (this.isL1) {
+      return
+    }
+
+    const promises: Array<Promise<any>> = []
+    promises.push(
+      this.checkTransferSentToL2FromDb(),
+      this.checkRelayableTransferRootsFromDb()
+    )
+    await Promise.all(promises)
```
miguelmota: suggestion: a 'done' log here would be nice

packages/hop-node/src/watchers/RelayWatcher.ts
```diff
@@ -0,0 +1,383 @@
...
+    await this.db.transfers.update(transferId, {
+      relayAttemptedAt: Date.now()
+    })
+
+    try {
+      logger.debug('checkTransferSentToL2 getIsRelayerFeeOk')
+      const isRelayerFeeOk = await this.getIsFeeOk(transferId, GasCostTransactionType.Relay)
+      if (!isRelayerFeeOk) {
+        const msg = 'Relayer fee is too low. Cannot relay.'
+        logger.debug(msg)
```
miguelmota: suggestion: change to `warn` log and maybe send notification

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -103,6 +112,15 @@ export type ExitableTransferRoot = {
   committedAt: number
 }
 
+export type RelayableTransferRoots = {
```
miguelmota: Should be singular name `RelayableTransferRoot`

packages/hop-node/src/watchers/RelayWatcher.ts
```diff
@@ -0,0 +1,383 @@
...
+      destinationChainId,
+      bondTxHash,
+      confirmTxHash
+    } = dbTransferRoot
+
+    const logger = this.logger.create({ root: transferRootId })
+
+    // bondTxHash should be checked first because a root can have both but it should be bonded prior to being confirmed
+    const l1TxHash = bondTxHash ?? confirmTxHash
+
```
miguelmota, line 288: suggestion: add a check here to make sure it exists

```ts
if (!l1TxHash) {
 throw new Error('expected bondTxHash or confirmTxHash')
}
```

packages/hop-node/src/watchers/RelayWatcher.ts
```diff
@@ -0,0 +1,383 @@
...
+    const dateFilter = {
+      fromUnix: transferSentTimestamp,
+      toUnix: transferSentTimestamp
+    }
+    const transfers: Transfer[] = await this.db.transfers.getTransfers(dateFilter)
+
+    // Get all transfers within the same L1 tx and store their log index
+    const logIndicesPerTransferId: Record<string, number> = {}
+    for (const transfer of transfers) {
+      if (transfer.transferSentTxHash !== transferSentTxHash) continue
```
miguelmota, line 373: suggestion: `continue` if `transferSentLogIndex` not set

```ts
if (typeof transfer.transferSentLogIndex === 'undefined') continue
```

packages/hop-node/src/watchers/RelayWatcher.ts
```diff
@@ -0,0 +1,383 @@
...
+      logIndicesPerTransferId[transfer.transferId] = transfer.transferSentLogIndex!
+    }
+
+    // Sort the transfers by their log index
+    const entries = Object.entries(logIndicesPerTransferId)
+    const sortedTransferIdsAndIndices = entries.sort((a, b) => a[1] - b[1])
+
+    let index = 0
+    for (const sortedTransferIdAndIndex of sortedTransferIdsAndIndices) {
```
miguelmota: suggestion: All these lines can be shorted to

```ts
return sortedTransferIdsAndIndices.map(([t]) => t).indexOf(transferId) 
```

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -327,6 +352,78 @@ class SyncWatcher extends BaseWatcher {
...
+      logIndex
+    )
+    const logger = this.logger.create({ id: transferId })
+    logger.debug('handling TransferSentToL2 event')
+
+    try {
+      const blockNumber: number = event.blockNumber
+      const l1Bridge = this.bridge as L1Bridge
+      const sourceChainId = await l1Bridge.getChainId()
+      const destinationChainId = Number(destinationChainIdBn.toString())
```
miguelmota: since these two do the same casting

```ts
      Number(destinationChainIdBn.toString()),
```

```ts
      const destinationChainId = Number(destinationChainIdBn.toString())
```

the const destinationChainId line can be moved to the top to make it more dry

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -660,6 +758,36 @@ class SyncWatcher extends BaseWatcher {
     await this.populateTransferRootTransferIds(transferRootId)
   }
 
+  async populateTransferSentToL2Timestamp (transferId: string) {
```
miguelmota: I think this whole function can be removed? looks to be the same as `populateTransferSentTimestamp`

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -1262,6 +1390,16 @@ class SyncWatcher extends BaseWatcher {
     return true
   }
 
+  getIsRelayable = (
+    relayerFee: BigNumber
+  ): boolean => {
+    if (relayerFee.eq(0)) {
```
miguelmota: since the condition is already a boolean it can be simplified with `return relayerFee.gt(0)`

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -1333,28 +1471,34 @@ class SyncWatcher extends BaseWatcher {
           ] as const
           const gasLimit = await bridgeContract.estimateGas.bondWithdrawalAndDistribute(...payload)
           const tx = await bridgeContract.populateTransaction.bondWithdrawalAndDistribute(...payload)
-          estimates.push({ gasLimit, ...tx, attemptSwap: true })
+          estimates.push({ gasLimit, ...tx, transactionType: GasCostTransactionType.BondWithdrawal })
+        }
+
+        if (this.chainSlug === Chain.Arbitrum) {
```
miguelmota: suggestion: it can check if chain is a relayable chain to future proof it

```ts
if (relayableChains.includes(this.chainSlug)) {
 ...
}
```

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -1374,6 +1518,132 @@ class SyncWatcher extends BaseWatcher {
...
+
+  private async _getRelayGasCost (): Promise<BigNumber> {
+    if (this.chainSlug === Chain.Arbitrum) {
+      return this._getArbitrumRelayGasCost()
+    }
+
+    throw new Error('Unsupported relay chain')
+  }
+
+  private async _getArbitrumRelayGasCost (): Promise<BigNumber> {
```
miguelmota: suggestion: I think this function (and associated functions) should be moved to a separate util since this is very arbitrum specific and it can be unit tested separately

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -1374,6 +1518,132 @@ class SyncWatcher extends BaseWatcher {
...
+    )
+
+    return encodedData
+  }
+
+  private async _getEncodedDistributeData (): Promise<string> {
+    const recipient = constants.AddressZero
+    const amount = BigNumber.from(10)
+    const amountOutMin = BigNumber.from(0)
+    const deadline = BigNumber.from('9999999999')
```
miguelmota: suggestion: move this value to a constant 

```ts
export const MaxDeadline :number = 9999999999
```

packages/hop-node/src/watchers/classes/BaseWatcher.ts
```diff
@@ -300,6 +324,98 @@ class BaseWatcher extends EventEmitter implements IBaseWatcher {
...
+    } else {
+      throw new Error('expected nearestItemToTransferSent or nearestItemToNow')
+    }
+
+    logger.debug('gasCostInToken:', gasCostInToken?.toString())
+    logger.debug('transactionType:', transactionType)
+
+    const minTxFee = gasCostInToken.div(2)
+    if (transactionType === GasCostTransactionType.Relay) {
+      const isRelayFeeOk = relayerFee!.gte(minTxFee)
```
miguelmota: suggestion: add check for relayerFee value

```ts
if (!relayerFee) {
  throw new Error('expected relayerFee')
}
```

packages/hop-node/src/watchers/classes/L1Bridge.ts
```diff
@@ -202,8 +221,14 @@ export default class L1Bridge extends Bridge {
       throw new Error(`chain ID "${destinationChainId}" is not supported`)
     }
 
-    const relayer = DefaultRelayerAddress
-    const relayerFee = '0'
+    const destinationChain = chainIdToSlug(destinationChainId)
+    const transactionType = GasCostTransactionType.BondWithdrawal
+    const now = Math.floor(Date.now() / 1000)
+    const nearestItemToTransferSent = await this.db.gasCost.getNearest(destinationChain, this.tokenSymbol, transactionType, now)
```
miguelmota, line 227: suggestion: Add check for `nearestItemToTransferSent` existence

```ts
if (!nearestItemToTransferSent) {
  throw new Error('nearestItemToTransferSent not found')
}
```

packages/hop-node/src/watchers/classes/L1Bridge.ts
```diff
@@ -237,10 +262,16 @@ export default class L1Bridge extends Bridge {
       throw new Error(`chain ID "${destinationChainId}" is not supported`)
     }
 
+    const destinationChain = chainIdToSlug(destinationChainId)
+    const transactionType = GasCostTransactionType.BondWithdrawal
+    const now = Math.floor(Date.now() / 1000)
+    const nearestItemToTransferSent = await this.db.gasCost.getNearest(destinationChain, this.tokenSymbol, transactionType, now)
```
miguelmota, line 271: suggestion: Add check for `nearestItemToTransferSent` existence

```ts
if (!nearestItemToTransferSent) {
  throw new Error('nearestItemToTransferSent not found')
}
```

packages/sdk/src/HopBridge.ts
```diff
@@ -2240,6 +2255,122 @@ class HopBridge extends Base {
...
+  private async getRelayerFee (destinationChain: TChain): Promise<BigNumber> {
+    // TODO: Introduce this post-nitro
+    // if (destinationChain === Chain.Arbitrum) {
+    //   return this.getArbitrumRelayGasCost(destinationChain)
+    // }
+
+    return BigNumber.from(0)
+  }
+
+  private async getArbitrumRelayGasCost (destinationChain: TChain): Promise<BigNumber> {
```
miguelmota, line 2269: suggestion: same comment as in hop-node as to move this logic to a util so it can be independently unit tested and to make it easier to copy/paste from hop-node<>sdk since the logic looks to be the same

miguelmota:
I think the new watcher and order of operations is good and makes sense. My feedback is mostly on additional error handling and some simplifications. The functions that deal with arbitrum gas (ie `getArbitrumRelayGasCost`) are out of my range of knowledge (i need to read arb docs and get more familiar with the calculations) so my suggestion there is to move that logic to it's own class or util and have some unit tests if possible

packages/hop-node/src/watchers/RelayWatcher.ts
shanefontaine: Instead of `continue` I threw here. It should be impossible for this to happen [because  `txHash` and `logIndex` are added atomically](https://github.com/hop-protocol/hop/blob/7f7cd207bba4d7dc67d247b8fc26548c42a7313b/packages/hop-node/src/watchers/SyncWatcher.ts#L417)).

If it does happen somehow, then there is a possibility that `getMessageIndex()` returns a different result for the same input, and handling that would likely require investigation. Throwing here will just make it more obvious that this needs to be looked into, IMO.

packages/hop-node/src/watchers/SyncWatcher.ts
shanefontaine: Resolved in 6f6cc07f0a419b7c2d18accbd3969aa1e951e472

packages/sdk/src/HopBridge.ts
shanefontaine: Resovled in 6f6cc07f0a419b7c2d18accbd3969aa1e951e472

packages/hop-node/src/watchers/RelayWatcher.ts
miguelmota: > Instead of continue I threw here.

that makes sense however I don't see the `throw` here?

if referring to the `!` operator then that wouldn't throw during runtime

shanefontaine: It is [here](https://github.com/hop-protocol/hop/pull/451/files#diff-5951b7a31d1735a4930f2d972f3a02a0f4fedcfe8413d0213803edb57637b5bcR374-R376). I think there might just be some weird UI thing, but let me know if it is not showing up for you.

miguelmota: I see it now 👍

PR #453, opened 2022-08-22
fe: add faucet

shanefontaine:


miguelmota:
@shanefontaine added the erc20 mintable contract instance

miguelmota:
I think `checkConnectedNetworkId` needs to be added to prompt the network switch but other than it seems good

PR #457, opened 2022-09-03
stats-worker: add mw stats

shanefontaine:
Adds the Arbitrum messenger wrapper (L1) balance.

Mostly curious about the DB:

* Is adding the additional `?` to the `INSERT OR REPLACE INTO bonder_balances VALUES` correct?
* Should anything be added to the migrations to reflect this update? For example:

```
if (migrations.includes(x)) {
  this.db.run(
    'ALTER TABLE bonder_balances ADD COLUMN arbitrum_messneger_wrapper_ammountt TEXT;'
  )
}
```


miguelmota (approved):

miguelmota:
Updates look good

1. yes that's correct
2. yes the migration statement should be added and is needed to add column on existing db

PR #460, opened 2022-09-26
feature/snx

miguelmota:
- Add SNX metadata to packages
- TODO: Add SNX bridge contract addresses


shanefontaine (approved):
LGTM. Can this be merged into develop even though the `core/addresses` file is incomplete?

miguelmota:
@shanefontaine yes as long as the SNX token isn't enabled in env var for frontend

PR #462, opened 2022-10-03
Feat/updated confirmation system

shanefontaine:
Updated root confirmation system for [updated contract logic](https://github.com/hop-protocol/contracts/blob/feat/updated-root-confirmation/contracts/wrappers/MessengerWrapper.sol#L53-L98).

* Add ability to confirm a root via the messengerWrapper if certain conditions are met
* Rename `exitCommitTx` CLI command to a now more accurate `confirmRoot`

As a note, I think it makes sense to rename `xDomainMessageRelayWatcher` to something more accurate since it is now a confirmation watcher. I will wait to do this until after the PR review so the diff doesn't become unreadable.


packages/hop-node/src/watchers/xDomainMessageRelayWatcher.ts
```diff
@@ -19,12 +23,20 @@ type Config = {
   dryMode?: boolean
 }
 
+type ConfirmRootData = {
+  rootHash: string
+  destinationChainId: number
+  totalAmount: string | BigNumber
```
miguelmota: I think this type should only be `BigNumber` for simplicity

packages/hop-node/src/cli/confirmRoot.ts
```diff
@@ -0,0 +1,87 @@
+import ArbitrumBridgeWatcher from 'src/watchers/ArbitrumBridgeWatcher'
+import GnosisBridgeWatcher from 'src/watchers/GnosisBridgeWatcher'
+import OptimismBridgeWatcher from 'src/watchers/OptimismBridgeWatcher'
+import PolygonBridgeWatcher from 'src/watchers/PolygonBridgeWatcher'
+import { actionHandler, parseBool, parseInputFileList, parseString, parseStringArray, root } from './shared'
+import { getXDomainMessageRelayWatcher } from 'src/watchers/watchers'
+import { BigNumber } from 'ethers'
+
+type ExitWatcher = GnosisBridgeWatcher | PolygonBridgeWatcher | OptimismBridgeWatcher | ArbitrumBridgeWatcher
+type ConfirmRootData = {
```
miguelmota: This type can be imported from `xDomainMessageRelayWatcher`

```suggestion
import { ConfirmRootData } from 'src/watchers/xDomainMessageRelayWatcher'
```

packages/hop-node/src/watchers/xDomainMessageRelayWatcher.ts
```diff
@@ -19,12 +23,20 @@ type Config = {
   dryMode?: boolean
 }
 
+type ConfirmRootData = {
```
miguelmota: This type can be exported so it's used in cli file

```suggestion
export type ConfirmRootData = {
```

packages/hop-node/src/watchers/xDomainMessageRelayWatcher.ts
```diff
@@ -19,12 +23,20 @@ type Config = {
...
+  rootCommittedAt: number
+}
+
 type Watcher = GnosisBridgeWatcher | PolygonBridgeWatcher | OptimismBridgeWatcher | ArbitrumBridgeWatcher
 
 class xDomainMessageRelayWatcher extends BaseWatcher {
   l1Bridge: L1Bridge
   lastSeen: {[key: string]: number} = {}
   watchers: {[chain: string]: Watcher} = {}
+  messengerWrapper: MessengerWrapper
```
miguelmota: general suggestion; would prefer `messengerWrapper` to be labeled as `l1MessengerWrapper` where ever it's referenced in hop node to follow existing naming convention we have in core package

packages/hop-node/src/watchers/xDomainMessageRelayWatcher.ts
```diff
@@ -71,29 +83,52 @@ class xDomainMessageRelayWatcher extends BaseWatcher {
...
+    }
+    this.messengerWrapper = new MessengerWrapper(messengerWrapperContract)
+
     // xDomain relayer is less time sensitive than others
     this.pollIntervalMs = 10 * 60 * 1000
   }
 
   async pollHandler () {
-    await this.checkTransfersCommittedFromDb()
+    await Promise.all([
```
miguelmota: any consequences of having both promises canceled/failed if one them throws here in the Promise.all?

packages/hop-node/src/watchers/xDomainMessageRelayWatcher.ts
```diff
@@ -115,15 +149,74 @@ class xDomainMessageRelayWatcher extends BaseWatcher {
...
+
+  async confirmRootsViaWrapper (rootData: ConfirmRootData[]): Promise<void> {
+    const rootHashes: string[] = []
+    const destinationChainIds: number[] = []
+    const totalAmounts: BigNumber[] = []
+    const rootCommittedAt: number[] = []
+    for (const data of rootData) {
+      rootHashes.push(data.rootHash),
+      destinationChainIds.push(data.destinationChainId),
+      totalAmounts.push(BigNumber.from(data.totalAmount)),
```
miguelmota: I think `totalAmount` should be a big number already

```suggestion
      totalAmounts.push(data.totalAmount),
```

packages/hop-node/src/watchers/xDomainMessageRelayWatcher.ts
shanefontaine: Interesting. I think the bonder could get into a bad spot if a tx throws on one of those on each poll. Do you think it makes sense to `try...catch` and/or serialize these calls?

packages/hop-node/src/watchers/xDomainMessageRelayWatcher.ts
shanefontaine: I went back and forth on this a few times. My conclusion (which I'm still 50/50 on) was that we use l1/l2 on contract names that apply to either L1 or L2 (bridge, token, etc.) and not necessarily on names that only have representation on EITHER l1 OR l2. For example, we [don't use `l2` on the `amm` stuff](https://github.com/hop-protocol/hop/blob/develop/packages/hop-node/src/contracts/contracts.ts#L97-L98).

With that said, I'm fine with (and like) using l1 for the messenger wrapper and the l1/l2 preface generally. If you give a thumbs up, I'll go ahead and update, but just wanted to provide some extra context.

packages/hop-node/src/watchers/xDomainMessageRelayWatcher.ts
miguelmota: I think try/catch would be good since they don't need to be serialized

packages/hop-node/src/watchers/xDomainMessageRelayWatcher.ts
miguelmota: I see now what you mean.
For consistency, and clarity, I do still prefer the `l1`/`l2` prefix and the other 2 instances you pointed out should have had the prefix as well because that's how it is in core package and contracts (at least for ammWrapper)

We currently use `L2Amm` and `L2AmmWrapper` for the classes which I like and the label `saddleSwap` can probably be generalized to `l2Amm` across the packages too but probably not worth the trouble changing it anyway

miguelmota (approved):

shanefontaine:
All comments resolved with ebacf82c3b5de1a9b600eaca76ea647b32d6a0ab

PR #466, opened 2022-10-07
hop-node: update rebroadcast logic

shanefontaine:
Resolves two issues:
* Previously, a tx that was re-broadcasted was retried every poll because `item!.sentAt = Date.now()` was being set _after_ the tx that was failing (because of an `already known` error). This was causing a ton of RPC usage and overhead. This PR moves the state setting logic to occur prior to the tx so that the result of the tx does not matter
* Introduce the concept of max reboosts. If a tx doesn't get picked up after max gas AND another ~30 mins, there is something wrong and should be handled accordingly

### DIFF
--- a/packages/hop-node/src/gasboost/GasBoostTransaction.ts
+++ b/packages/hop-node/src/gasboost/GasBoostTransaction.ts
@@ -81,11 +81,14 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
   initialTxGasPriceMultiplier: number = MaxGasPriceMultiplier // multiplier for gasPrice for first tx
   maxGasPriceGwei: number = 500 // the max we'll keep bumping gasPrice in type 0 txs
   maxGasPriceReached: boolean = false // this is set to true when gasPrice is greater than maxGasPrice
+  maxRebroadcastIndex: number = 10
+  maxRebroadcastIndexReached: boolean = false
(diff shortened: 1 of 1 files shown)


packages/hop-node/src/gasboost/GasBoostTransaction.ts
```diff
@@ -564,10 +571,19 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
...
+    this.logger.debug(`confirmed tx: ${tx.hash}, boostIndex: ${this.boostIndex}, rebroadcastIndex: ${this.rebroadcastIndex}, nonce: ${this.nonce.toString()}, ${this.getGasFeeDataAsString()}`)
     this.watchForReorg()
   }
 
+  private async handleMaxRebroadcastIndexReached () {
+    this.maxRebroadcastIndexReached = true
+    this.clearInflightTxs()
+    this.emit(State.Error)
+    const warnMsg = `max rebroadcast index reached. cannot rebroadcast.`
+    this.notifier.warn(warnMsg, { channel: gasBoostWarnSlackChannel })
```
miguelmota: suggestion: make this error logs instead of warns

```suggestion
   const errMsg = `max rebroadcast index reached. cannot rebroadcast.`
    this.notifier.error(errMsg, { channel: gasBoostErrorSlackChannel })
    this.logger.error(errMsg)
```

miguelmota (approved):

miguelmota:
@shanefontaine the maxRebroadcasts logic is a good idea and the PR looks good 👍

PR #470, opened 2022-10-20
Add lazy loading routes

alter-eggo:
Current bundle of hop app is quite big. First step to decrease bundle size and make first render faster may be implementing lazy loading routes.

Screenshot 2022-10-20 at 05 48 41


alter-eggo:
hey @miguelmota @shanefontaine! 
can you take a look here please?

miguelmota:
@alter-eggo thanks for the PR! looks good, we should be able to get it merged this week

PR #471, opened 2022-10-20
Add bundle analyse script

alter-eggo:
The final bundle should be analysed and reduced in order to increase first render. This pr adds bundle analyse script and `webpack-bundle-analyzer` package.


miguelmota:
hey @alter-eggo, appreciate the PR however due to supply chain risk attack vectors we're limiting the type of PRs we're accepting to only bug fixes at the moment

alter-eggo:
@miguelmota ok, I see

PR #479, opened 2023-02-09
Feature/update op sdk

miguelmota:


packages/hop-node/package.json
```diff
@@ -83,6 +82,7 @@
     "p-queue": "6.6.2",
     "pidusage": "2.0.21",
     "pify": "5.0.0",
+    "prom-client": "^14.1.1",
```
shanefontaine, line 85: Is this an artifact that should be removed?

shanefontaine (approved):
Looks good. One small comment.

packages/hop-node/package.json
miguelmota: it's always been imported in metrics server [here](https://github.com/hop-protocol/hop/blob/2612ffc32d438b2ca01fd61ab127a64de5ce2d96/packages/hop-node/src/metrics/metrics.ts#L1) but for some reason it wasn't in package.json so i was getting build errors after installing optimism packages

PR #488, opened 2023-04-21
Feat/kms signer review

shanefontaine:


packages/hop-node/test/KmsSigner.test.ts
```diff
@@ -1,48 +1,42 @@
...
-import { providers } from 'ethers'
+import { Chain } from 'src/constants'
+import { getRpcProvider } from 'src/utils/getRpcProvider'
 
-describe('KmsSigner', () => {
-  const keyId = '32a977a2-b532-40b3-af2e-f064d3980f75'
-  const region = 'us-west-1'
+describe.skip('KmsSigner', () => {
+  const keyId = process.env.TEST_KMS_KEY_ID!
+  const region = 'us-east-1'
```
miguelmota: I think the region could also be an env var here

packages/hop-node/test/KmsSigner.test.ts
```diff
@@ -1,48 +1,42 @@
...
   const signer = new KmsSigner({ keyId, region })
   it('getAddress', async () => {
     const address = await signer.getAddress()
     console.log('address:', address)
-    expect(address).toBe('0x38621a41820032e3e1b2787de196b20c19b5df74')
+    expect(address).toBe('0x7A4f3c58B52f3e8cD5100f01b9C8bF01F20C5A70')
```
miguelmota: maybe the address should also be an env var since the input key is an env var

packages/hop-node/src/cli/verifyKmsSigner.ts
```diff
@@ -0,0 +1,52 @@
...
+  // Send tx to self
+  if (sendTestTx) {
+    const address = await signer.getAddress()
+    const transaction = {
+      to: address
+    }
+    const provider = getRpcProvider(Chain.Ethereum)
+    const tx = await signer.connect(provider!).sendTransaction(transaction)
+    console.log(`Transaction sent: ${tx.hash}`)
+  }
```
miguelmota, line 57: another check to add here could be to check the `from` address of `receipt` to match the expected `address`

packages/hop-node/src/aws/KmsSigner.ts
```diff
@@ -123,9 +140,9 @@ export class KmsSigner extends Signer {
     return joinedSignature
   }
 
-  private async _getSigV (msgHash: Buffer, { r, s }: {r: string, s: string}) {
+  private async _getSigV (msgHash: Buffer, { r, s }: { r: string, s: string }) {
     const address = await this.getAddress()
-    let v = 17
+    let v = 27
```
miguelmota, line 145: good catch

packages/hop-node/package.json
```diff
@@ -57,6 +57,7 @@
     "@maticnetwork/maticjs-ethers": "1.0.2",
     "@slack/web-api": "6.7.2",
     "@yfi/sdk": "2.1.0",
+    "asn1.js": "^5.4.1",
```
miguelmota: maybe this version should also be exact (no `^`)

miguelmota (approved):
some minor feedback but looks good 👍

shanefontaine:
All points addressed in aaabb855d11d6e27c7cee70574121fb117655bda

PR #489, opened 2023-04-22
Feat/lambda signer

shanefontaine:
Adds a base AwsSigner class and separates KMS/Lambda specific logic into their own classes.


packages/hop-node/src/aws/AwsSigner.ts
```diff
@@ -0,0 +1,123 @@
...
+export type AwsSignerConfig = {
+  keyId: string
+  region?: string
+}
+
+// details:
+// https://ethereum.stackexchange.com/a/73371/5093
+// https://luhenning.medium.com/the-dark-side-of-the-elliptic-curve-signing-ethereum-transactions-with-aws-kms-in-javascript-83610d9a6f81
+// https://github.com/lucashenning/aws-kms-ethereum-signing/blob/master/aws-kms-sign.ts
+export abstract class AwsSigner extends Signer {
```
miguelmota, line 42: suggestion: rename class to `KmsUtils` since it's KMS specific and it can't be used as a standalone signer

packages/hop-node/src/aws/LambdaSigner.ts
```diff
@@ -0,0 +1,121 @@
...
+    const serializedTx = serializeTransaction(unsignedTx)
+    const hash = keccak256(serializedTx)
+    const txSig = await this._signDigest(hash, transaction)
+    return serializeTransaction(unsignedTx, txSig)
+  }
+
+  private async _signDigest (digest: Buffer | string, transaction?: providers.TransactionRequest): Promise<string> {
+    const msg = Buffer.from(arrayify(digest))
+    const signature = await this._getSig(msg, transaction)
+    const { r, s } = this.getSigRs(signature)
```
miguelmota: suggestion: since this logic is duplicated in both classes, it can be moved to the base class

Example

```ts
private async getJoinedSignature(msg: Buffer, signature: Buffer): Promise<string> {
    const { r, s } = this.getSigRs(signature)
    const { v } = await this.getSigV(msg, { r, s })
    const joinedSignature = joinSignature({ r, s, v })
    return joinedSignature
}
```

Usage

```ts
  private async _signDigest (digest: Buffer | string, transaction?: providers.TransactionRequest): Promise<string> {
    const msg = Buffer.from(arrayify(digest))
    const signature = await this._getSig(msg, transaction)
    return this.getJoinedSignature(msg, signature)
}
```

packages/hop-node/src/aws/LambdaSigner.ts
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  private async _lambdaPayloadToUint8Array (payload: any): Promise<Uint8Array> {
+    const decoder = new TextDecoder()
+    const decodedSignature: string = decoder.decode(payload)
+    const jsonSignature: any = JSON.parse(decodedSignature)
+
+    let signatureNumberArray: number[] = []
+    for (const index in jsonSignature) {
+      signatureNumberArray.push(jsonSignature[index])
```
miguelmota: suggestion: simplify this function by removing the for loop and returning a `Buffer` instead since `Uint8Array` isn't used directly in class.

It could be something like this:

```ts
  private async _lambdaPayloadToBuffer (payload: any): Promise<Buffer> {
    const decoder = new TextDecoder()
    const decodedSignature: string = decoder.decode(payload)
    const jsonSignature: any = JSON.parse(decodedSignature)

    return Buffer.from(Object.values(jsonSignature))
}
```

packages/hop-node/src/aws/LambdaSigner.ts
```diff
@@ -0,0 +1,121 @@
...
+      transaction
+    }
+    const params = {
+      FunctionName: this.lambdaFunctionName,
+      InvocationType: "RequestResponse",
+      Payload: Buffer.from(JSON.stringify(transactionRequest))
+    }
+    const command = new InvokeCommand(params)
+    const res: any = await this.client.send(command)
+    const signature: Uint8Array = await this._lambdaPayloadToUint8Array(res.Payload)
```
miguelmota: suggestion: create `_lambdaPayloadToBuffer`  instead since we don't really deal with `Uint8Array` directly

packages/hop-node/src/cli/verifyAwsSigner.ts
```diff
@@ -1,43 +1,55 @@
...
 import {
   config as globalConfig
 } from 'src/config'
 
 root
-  .command('verify-kms-signer')
-  .description('Verify a KMS signer')
+  .command('verify-aws-signer')
+  .description('Verify an AWS signer')
+  .option('--lambda-signer [boolean]', 'Use the Lambda signer', parseBool)
```
miguelmota: i could be misunderstanding something here, but i think if the option to use lambda signer is set in config file, then this flag is not necessary since it's already implied that lambda signer is being used.

The boolean var can be this:

`const lambdaSigner = type === 'lambda'`

packages/hop-node/src/cli/verifyAwsSigner.ts
```diff
@@ -1,43 +1,55 @@
...
   .action(actionHandler(main))
 
 async function main (source: any) {
-  const { sendTestTx } = source
-  const { type, keyId, awsRegion } = globalConfig.signerConfig
-  if (type !== 'kms') {
+  const { sendTestTx, lambdaSigner } = source
+  const { type, keyId, awsRegion, lambdaFunctionName } = globalConfig.signerConfig
+  if (type !== 'kms' && type !== 'lambda') {
     throw new Error('signer type must be kms')
```
miguelmota: the throw here should mention `"kms" or "lambda"`

packages/hop-node/test/LambdaSigner.test.ts
```diff
@@ -0,0 +1,43 @@
+import { LambdaSigner } from 'src/aws/LambdaSigner'
+import { Chain } from 'src/constants'
+import { getRpcProvider } from 'src/utils/getRpcProvider'
+
+describe.only('KmsSigner', () => {
```
miguelmota: test name should be `LambdaSigner`

packages/hop-node/src/aws/AwsSigner.ts
shanefontaine: I just thought about this one a bit. My current thinking is to leave it as is for two reasons:

* Technically, it extends a `Signer` class, so by definition, it is a `Signer` (if I'm thinking about that correctly)
* It is marked `abstract`, so can't be used standalone

Given that, I think it makes sense to leave it. If it was a `util` class, then I think it would be odd for the parent signing classes to extend `xxxUtil` instead of a signer. But very loose opinion here, so LMK what you think.

packages/hop-node/src/aws/LambdaSigner.ts
shanefontaine: Awesome. I found the types to be difficult to work with. This suggestion appears to simplify all of my previous issues here 🙏

shanefontaine:
@miguelmota all but one comment resolved. LMK what you think about the final comment I left.

packages/hop-node/src/aws/AwsSigner.ts
miguelmota: @shanefontaine that sounds good, I missed seeing the `abstract` keyword so make sense 👍

PR #490, opened 2023-05-05
sdk: PriceFeed: Fetch prices from S3 json file, fallback to apis

miguelmota:
Stats worker is writing to json file

```sh
curl https://assets.hop.exchange/token-prices.json | jq
```

SDK fetches data from this json file, and falls back to quering apis if price is either missing or stale.


shanefontaine (approved):
@miguelmota LGTM 👍

PR #494, opened 2023-05-16
Feat/gasboost tx updates

shanefontaine:
* Fixes bug that infinitely loops and attempts transactions if the initial tx has gas data greater than the max cap. In the current logic, if a tx is created when the network's gas is greater than the max the bonder is willing to spend, the gas data for the tx will equal the max (i.e. 500 gwei). If the tx is not included in a block, the tx will be boosted. In `boost()`, `isMaxReached` checks that the gas data is `gt` the max gas data. However, in this case, the gas data is equal to the max gas data, so `isMaxReached` is never set to `true` and the tx gets retried 10 times and eventually throws. That throw is not handled in the poller, so the tx is attempted again 10 more times, and the rest repeats.
  * Fix 1: When calling `boost()`, check for `gte` instead of `gt`
  * Fix 2: Explicitly handle the err case when max index is reached. Break out of the poller in this case.
  * Example error log from old code: `tx index 1 error: replacement fee too low [ See: https://links.ethers.org/v5-errors-REPLACEMENT_UNDERPRICED ]..."body": "{\"jsonrpc\":\"2.0\",\"id\":187499,\"error\":{\"code\":-32000,\"message\":\"replacement transaction underpriced\"}}",`
* Adds a unique timestamp to `console.time()`. Sometimes we would call `console.time()`, but error out before calling `console.timeEnd()` for that ID. If we loop again, there is no uniqueness, so the logs have an error: `(node:1) Warning: Label 'GasBoostTransaction signer.sendTransaction elapsed ethereum id: 5ebb1dd5-b62a-4fcd-b0e0-5bec648de1ca transferId: 0x1718f96dd5cec4c673274f0021cc564b11fcaab27593d432881e509491b26d95 5 ' already exists for console.time()`. This just avoids that error
* Adds redundant logs for use in future debugging

### DIFF
--- a/packages/hop-node/src/gasboost/GasBoostTransaction.ts
+++ b/packages/hop-node/src/gasboost/GasBoostTransaction.ts
@@ -116,9 +116,10 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
   createdAt: number
   txHash?: string
   receipt?: providers.TransactionReceipt
-  private _is1559Supported: boolean // set to true if EIP-1559 type transactions are supported
   readonly minMultiplier: number = 1.10 // the minimum gas price multiplier that miners will accept for transaction replacements
   logId: string
+  is1559Supported: boolean // set to true if EIP-1559 type transactions are supported
+  shouldUseBlocknative: boolean
 
   reorgWaitConfirmations: number = 1
   originalTxParams: providers.TransactionRequest
@@ -171,6 +172,15 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
     this.notifier = new Notifier(
       `GasBoost, label: ${prefix}, host: ${hostname}`
     )
+    this.init()
+      .catch((err: Error) => this.logger.error('init error:', err))
+  }
+
+  private async init () {
+    const { maxFeePerGas, maxPriorityFeePerGas } = await this.getGasFeeData()
+    const isSupported = !!((maxFeePerGas != null) && (maxPriorityFeePerGas != null))
+    this.is1559Supported = isSupported
+    this.shouldUseBlocknative = this.is1559Supported && this.chainSlug === Chain.Ethereum
   }
 
   private generateId (): string {
@@ -396,8 +406,7 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
   }
 
   async getMarketMaxPriorityFeePerGas (): Promise<BigNumber> {
-    const isMainnet = typeof this._is1559Supported === 'boolean' && this._is1559Supported && this.chainSlug === Chain.Ethereum
-    if (isMainnet) {
+    if (this.shouldUseBlocknative) {
       try {
         const baseUrl = 'https://api.blocknative.com/gasprices/blockprices?confidenceLevels='
         const url = baseUrl + this.maxPriorityFeeConfidenceLevel.toString()
@@ -462,7 +471,7 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
   }
 
   async getBumpedGasFeeData (multiplier: number = this.gasPriceMultiplier): Promise<Partial<GasFeeData>> {
-    const use1559 = await this.is1559Supported() && !this.gasPrice && this.type !== 0
+    const use1559 = this.is1559Supported && !this.gasPrice && this.type !== 0
 
     if (use1559) {
       let [maxFeePerGas, maxPriorityFeePerGas, currentBaseFeePerGas] = await Promise.all([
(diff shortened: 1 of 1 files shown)


packages/hop-node/src/gasboost/GasBoostTransaction.ts
```diff
@@ -693,12 +718,16 @@ class GasBoostTransaction extends EventEmitter implements providers.TransactionR
     const maxGasPrice = this.getMaxGasPrice()
     const priorityFeePerGasCap = this.getPriorityFeePerGasCap()
 
-    // don't boost if suggested gas is over max
-    const isGasPriceMaxReached = gasFeeData.gasPrice?.gt(maxGasPrice)
-    const isMaxFeePerGasReached = gasFeeData.maxFeePerGas?.gt(maxGasPrice)
-    const isMaxPriorityFeePerGasReached = gasFeeData.maxPriorityFeePerGas?.gt(priorityFeePerGasCap)
+    // don't boost if suggested gas is equal to or over max
+    const isGasPriceMaxReached = gasFeeData.gasPrice?.gte(maxGasPrice)
+    const isMaxFeePerGasReached = gasFeeData.maxFeePerGas?.gte(maxGasPrice)
```
miguelmota, line 726: the updates look good; my only question/concern is here with the `gte` change.

Since this boost function calls `getBumpedGasFeeData()` and in that function [here](https://github.com/hop-protocol/hop/blob/eb7fe2c7205fe8139e7dd0159d95be812281b049/packages/hop-node/src/gasboost/GasBoostTransaction.ts#L477) it clamps it to maxGasPrice if it exceeds it, does that mean when an initial tx is sent with gas price slightly below max, then after a while it attempts to boost it;  it can't boost since it checks for `gte`  which would be true because of the clamping and it'd be marked as `isMaxReached` though the max wasn't broadcasted? 

sugggestion:
The `send` function [here](https://github.com/hop-protocol/hop/blob/eb7fe2c7205fe8139e7dd0159d95be812281b049/packages/hop-node/src/gasboost/GasBoostTransaction.ts#L373) calls `this.setGasProperties(tx)` which keeps a reference to the last used `gasPrice` and `maxFeePerGas` so maybe the boost function just needs to first check `if (this.gasPrice.gte(maxGasPrice) || this.maxFeePerGas.gte(maxGasPrice)) isMaxReached = true`? or maybe it can check if the max has been reached after it calls `_sendTransaction()`?

I could be missing something here so let me know if this doesn't make sense

shanefontaine: @miguelmota That is a good point, I didn't think through that. I believe what you said is true but (now that I'm re-looking at it) is expected. In that case you described, the node would reject the boosted tx for being <10% more gas than the prior tx at that nonce.

As an example with the current code (pre-PR), max gas is 500 Gwei and the first tx is 495 gwei. Gas then spikes to 1k, so the tx doesn't get included on chain. It tries to boost, but is clamped at 500 gwei. In that case, the tx will not be marked as `maxGasPriceReached` (since the clamped value `===500`), the tx will be sent again at 500 Gwei, but the node will immediately reject the tx with `transaction underpriced` and the tx will retry a few more times at 500 Gwei, eventually failing and falling out of the loop. I haven't tested that, but I think that is how it would work, if I am thinking about this correctly.

In that same example with the PR code implemented, the first tx would be attempted at 495 Gwei and then the boosted one would be marked as `maxGasPriceReached` and not attempted. The end result would be the same (no new tx making it onchain), but the process would be more efficient (wouldn't retry known-bad txs). 

Given that, I think it makes sense to keep it as-is, but happy to continue the discussion. I might have missed something here as well.

miguelmota: @shanefontaine do you think there could be a scenario where it performs the initial tx (say <100 gwei) and then market gas spikes above the max (> 500gwei), so then it marks it `maxGasPriceReached` during attempted boost (and no boosted tx broadcasted) since [here](https://github.com/hop-protocol/hop/blob/9d8e7aec2fd2f3344063b490088c81917cd8df99/packages/hop-node/src/gasboost/GasBoostTransaction.ts#LL469C14-L469C35) in `getBumpedGasFeeData()` it uses the market gas price?

shanefontaine: @miguelmota I do believe that can happen and believe that it should be re-broadcasted at 500, which I now see it would not. 

To consolidate everything, here are the current things being tracked:

* A transfer sent at 100 Gwei should retry at 500 Gwei if the gas market spikes between the time of the initial tx and the boost
* If the initial tx is already at the max gas, that needs to be handled explicitly and not thrown into the boost logic
* `getBumpedGasFeeData()` should be consistent with clamping (or not clamping) of values
* `checkHasEnoughFunds()` should have its own error that is handled explicitly in the parent's `try...catch`. If it fails, this means the bonder is running out of funds and should be handled accordingly.

Will push a commit(s) that handles these and ping you when the updated PR is ready.

shanefontaine:
Closing since this is now deprecated. Can be reopened if needed.

PR #496, opened 2023-05-27
hop-node: normalize default path across platforms and various tweaks

shanefontaine:
* Normalizes default data path across Linux/Mac OSes to `~/.hop`. Allows bonders to not define path's in config/CLI option if they are commonly used (config, keystore, db). 
* Runs Docker instances with `node` user with restricted capabilities. This `node` user's home dir is `/home/node`.
* Removes [`.git` from Dockerfiles](https://www.digitalocean.com/community/tutorials/how-to-build-a-node-js-application-with-docker).

In practice, a new bonder should now be able to pull the image, run `hop keystore generate`, add a pretty general config with no paths (minus their token-specific info), and run docker with `-v ~/.hop:/home/node/.hop` (that should be the only place a new bonder defines a custom path).

Local testing still uses `~/.hop` by default, unless otherwise specified.

### DIFF
--- a/packages/hop-node/.dockerignore
+++ b/packages/hop-node/.dockerignore
@@ -1,3 +1,4 @@
 node_modules
 Dockerfile
 *.env
+.git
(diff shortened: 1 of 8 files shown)


shanefontaine:
Closing since this is now deprecated. Can be reopened if needed.

PR #498, opened 2023-06-03
Feat/bedrock updates

shanefontaine:


packages/frontend/src/config/config.ts
```diff
@@ -9,34 +9,40 @@ if (process.env.REACT_APP_NETWORK === 'goerli') {
 
 export { showRewards }
 
+export const orusThatRelyOnL1ConfirmationsForFinality: string[] = [
```
miguelmota, line 12: should `base` also be included here since it's based on optimism?

packages/frontend/src/utils/getIsTxFinalized.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export async function getIsTxFinalized (
+  txBlockNumber: number | undefined,
+  chainSlug: string
+) {
+  if (!txBlockNumber) return false
+
+  const provider = getProviderByNetworkName(chainSlug)
+  if (orusThatRelyOnL1ConfirmationsForFinality.includes(chainSlug)) {
+    const finalizedBlock = await provider.getBlock('safe')
```
miguelmota, line 13: when I tried `await provider.getBlock('safe')` using our optimism rpc or arbitrum rpc I got the error `Unable to complete request at this time.` (screenshot)

DeepinScreenshot_select-area_20230603214136

Do you know if I'm doing something wrong here, or is this expected?

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -484,16 +484,26 @@ class TransferRootsDb extends BaseDb {
...
       }
 
-      let timestampOk = true
+      // Since bonding of transferRoots is not time sensitive, wait an arbitrary amount of time for
+      // finality before attempting to bond. This prevents repetitive RPC calls, since that is the
+      // only true way to know finality for ORUs. The arbitrary time should represent roughly how long
+      // the longest chain should wait for finality.
+      let finalityTimestampOk = false
+      if (item?.committedAt) {
+        const longestTimeToFinalityMs = 2 * TenMinutesMs
```
miguelmota, line 493: suggestion: instead of the hardcoded 10 minutes, maybe finality times can be set in config to make this value be computed (feel free to ignore this suggestion if it doesn't make sense to do so)

```js
const finalityMinutes = { ethereum: 1, optimism: 2, arbitrum: 3 }
const longestTimeToFinalityMs = (Math.max(...Object.values(finalityMinutes)) * 2) * 60 * 1000
````

miguelmota (approved):

packages/frontend/src/config/config.ts
shanefontaine: Resolved with c7b246ffe3edead52b7e2b87fb28de38cfb990ce

packages/frontend/src/utils/getIsTxFinalized.ts
shanefontaine: On Optimism, this is intentional since it won't be exposed until Bedrock. It does work on the Goerli Optimism endpoint.

You are correct about Arbitrum. It appears as though Infura exposes this but Alchemy does not 🤔 . Will figure something out here, good call 👍

packages/hop-node/src/db/TransferRootsDb.ts
shanefontaine: For these chains there is no consistency in finalized time since it depends on when in an epoch a tx is included. All these chains should (roughly) be the same since finalization of all of these relies on L1 finality. Based on that, and the fact that this is already arbitrary and just used in a non-sensitive way (to reduce RPC calls), I think it makes sense to just leave this arbitrary-ish value as a single constant here.

PR #509, opened 2023-08-28
Airdrop reopen

max-andrew:


.gitignore
```diff
@@ -17,4 +17,7 @@ yarn-error.log*
 
 hop.sh
 1
-yarn.lock
\ No newline at end of file
+yarn.lock
+
+airdrop-reopen.sublime-workspace
```
miguelmota: suggestion: ignore all sublime workspaces/projects in gitignore

```suggestion
*.sublime-workspace
*.sublime-project
```

packages/frontend/src/components/header/HeaderRoutes.tsx
```diff
@@ -19,16 +19,13 @@ const useStyles = makeStyles((theme: any) => ({
...
   const styles = useStyles()
   const theme = useTheme()
 
   const handleChange = (event: ChangeEvent<{}>, newValue: string) => {
     event.preventDefault()
-    history.push({
-      pathname: newValue,
-      search,
-    })
+    navigate(newValue + search)
```
miguelmota: suggestion: use string interpolation instead

```suggestion
    navigate(`${newValue}${search}`)
```

packages/frontend/src/hooks/useQueryParams.tsx
```diff
@@ -14,9 +14,7 @@ const useQueryParams = () => {
   const updateQueryParams = (updates: { [index: string]: string | undefined }) => {
     Object.assign(queryParams, updates)
 
-    history.push({
-      search: stringify(queryParams),
-    })
+    navigate(location.pathname + '?' + stringify(queryParams))
```
miguelmota: suggestion: use string interpolation instead

```suggestion
    navigate(`${location.pathname}?${stringify(queryParams)}`)
```

packages/frontend/src/utils/format.ts
```diff
@@ -2,24 +2,35 @@ import { BigNumber, FixedNumber, utils } from 'ethers'
...
 
-  let errMsg = 'Something went wrong. Please try again.'
+export function formatError(error: unknown, network?: Network): string {
   if (typeof error === 'string') {
-    errMsg = error
-  } else if (error?.data?.message) {
-    errMsg = error.data.message
-  } else if (error?.message) {
-    errMsg = error.message
+    return prettifyErrorMessage(error)
```
miguelmota: I think if the error parameter is a string then it should still go through the if statement checks below that format the error message, instead of returning here.

packages/frontend/src/utils/format.ts
```diff
@@ -2,24 +2,35 @@ import { BigNumber, FixedNumber, utils } from 'ethers'
...
   }
 
   if (Array.isArray(error) && error.length === 1) {
     return formatError(error[0], network)
   }
 
+  if (typeof error !== 'object' || error === null) {
```
miguelmota: suggestion: use double equality to also check for `undefined`

```suggestion
  if (error == null) {
    return ''
  }
```

packages/frontend/src/utils/format.ts
```diff
@@ -2,24 +2,35 @@ import { BigNumber, FixedNumber, utils } from 'ethers'
...
+    return ''
+  }
+
+  const errObj = error as PossibleError
+
+  let errMsg = 'Something went wrong. Please try again.'
```
miguelmota: suggestion: if error is not a string or Object, then return generic error

```suggestion
  if (!(err instanceof Object) {
    return errMsg
  }
```

miguelmota (changes requested):
The updates look pretty good @max-andrew! 

I left various suggestions but mainly had trouble using the UI due to errors related to the dependency upgrades. Let me know if you have any thoughts on how to proceed. 

Screenshots of my steps and error messages:

DeepinScreenshot_select-area_20230828190600

DeepinScreenshot_select-area_20230828185542

DeepinScreenshot_select-area_20230828191104

max-andrew:
> The updates look pretty good @max-andrew!
> 
> I left various suggestions but mainly had trouble using the UI due to errors related to the dependency upgrades. Let me know if you have any thoughts on how to proceed.
> 
> Screenshots of my steps and error messages:
> 
> DeepinScreenshot_select-area_20230828190600
> 
> DeepinScreenshot_select-area_20230828185542
> 
> DeepinScreenshot_select-area_20230828191104

Thanks @miguelmota! This is a compatibility issue with TS 5.2+ and MUI 4. I believe I missed this due to a combination of the ^ notation on package.json automatically updating the minor version (TS 5.2 released Aug 24) and the yarn resolver being a bit more forgiving than npm's. I definitely would love to migrate MUI to 5 soon but the process was fairly involved.

For now a temporary solution is to remove that ^ to keep us at TS 5.1. We can then successfully install the dependencies using `npm i --legacy-peer-deps`. Not a permanent solution but it solves our immediate problem.

miguelmota:
@max-andrew it works now 🙏 thanks for quick fix!

max-andrew:
@miguelmota testing the error message handling updates now but I believe I integrated all your suggestions

miguelmota (approved):
@max-andrew good to merge 👍

PR #512, opened 2023-09-20
Feat/optimized sync watcher

shanefontaine:
Optimizes the sync watcher with a few changes and new features.

* Default sync cycle time updated to 30 sec (from 60 sec)
* `TransferSent` (and `TransferSentToL2` on relayable chains) events are retrieved every cycle while all others are retrieved on a different cadence (currently 60x the default cycle time, or 30 mins).
* Introduce a multiplier `SyncIterationMultiplier` for less-used chains. This multiplier updates the default cycle time (30 sec) to something like 60 sec.
* Minor optimization: the bonder was making a `getLogs` call every cycle, even if the latest known block was already at the head. Now, if the chain head is already known by the bonder, they will not make a `getLogs` call anymore.

### DIFF
--- a/packages/hop-node/src/watchers/SyncWatcher.ts
+++ b/packages/hop-node/src/watchers/SyncWatcher.ts
@@ -236,11 +236,11 @@ class SyncWatcher extends BaseWatcher {
   async syncHandler (): Promise<any> {
     // Events that are related to user transfers can be polled every cycle while
     // all other, less time-sensitive events can be polled every N cycles
-    const slowSyncModulo = this.resyncIntervalSec * SyncIterationMultiplier[this.chainSlug]
+    const fullSyncModulo = this.resyncIntervalSec * SyncIterationMultiplier[this.chainSlug]
     let promisesPerPoll: EventPromise = []
     if (
       !this.isInitialSyncCompleted() ||
-      this.syncIndex % slowSyncModulo === 0
+      this.syncIndex % fullSyncModulo === 0
     ) {
       promisesPerPoll = this.getAllPromises()
     } else {
--- a/packages/hop-node/src/watchers/SyncWatcher.ts
+++ b/packages/hop-node/src/watchers/SyncWatcher.ts
@@ -291,10 +291,9 @@ class SyncWatcher extends BaseWatcher {
     }
   }
 
-  getTransferSentToL2EventPromise (): Promise<any> {
-    if (!this.isL1) {
-      return Promise.resolve()
-    }
+  async getTransferSentToL2EventPromise (): Promise<any> {
+    if (!this.isL1) return []
+
     const l1Bridge = this.bridge as L1Bridge
     return l1Bridge.mapTransferSentToL2Events(
       event => this.handleTransferSentToL2Event(event),
@@ -302,10 +301,9 @@ class SyncWatcher extends BaseWatcher {
     )
   }
 
-  getTransferRootConfirmedEventPromise (): Promise<any> {
-    if (!this.isL1) {
-      return Promise.resolve()
-    }
+  async getTransferRootConfirmedEventPromise (): Promise<any> {
+    if (!this.isL1) return []
+
     const l1Bridge = this.bridge as L1Bridge
     return l1Bridge.mapTransferRootConfirmedEvents(
       event => this.handleTransferRootConfirmedEvent(event),
@@ -313,10 +311,9 @@ class SyncWatcher extends BaseWatcher {
     )
   }
 
-  getTransferBondChallengedEventPromise (): Promise<any> {
-    if (!this.isL1) {
-      return Promise.resolve()
-    }
+  async getTransferBondChallengedEventPromise (): Promise<any> {
+    if (!this.isL1) return []
+
     const l1Bridge = this.bridge as L1Bridge
     return l1Bridge.mapTransferBondChallengedEvents(
       event => this.handleTransferBondChallengedEvent(event),
@@ -324,10 +321,9 @@ class SyncWatcher extends BaseWatcher {
     )
   }
 
-  getTransferSentEventPromise (): Promise<any> {
-    if (this.isL1) {
-      return Promise.resolve()
-    }
+  async getTransferSentEventPromise (): Promise<any> {
+    if (this.isL1) return []
+
     const l2Bridge = this.bridge as L2Bridge
     return l2Bridge.mapTransferSentEvents(
       event => this.handleTransferSentEvent(event),
@@ -335,17 +331,16 @@ class SyncWatcher extends BaseWatcher {
     )
   }
 
-  getTransferRootSetEventPromise (): Promise<any> {
+  async getTransferRootSetEventPromise (): Promise<any> {
     return this.bridge.mapTransferRootSetEvents(
       event => this.handleTransferRootSetEvent(event),
       this.getSyncOptions(this.bridge.TransferRootSet)
     )
   }
 
-  getTransferRootBondedEventPromise (): Promise<any>{
-    if (!this.isL1) {
-      return Promise.resolve()
-    }
+  async getTransferRootBondedEventPromise (): Promise<any>{
+    if (!this.isL1) return []
+
     const l1Bridge = this.bridge as L1Bridge
     return l1Bridge.mapTransferRootBondedEvents(
       event => this.handleTransferRootBondedEvent(event),
@@ -353,10 +348,9 @@ class SyncWatcher extends BaseWatcher {
     )
   }
 
-  getTransfersCommittedEventPromise (): Promise<any> {
-    if (this.isL1) {
-      return Promise.resolve()
-    }
+  async getTransfersCommittedEventPromise (): Promise<any> {
+    if (this.isL1) return []
+
     const l2Bridge = this.bridge as L2Bridge
     return l2Bridge.mapTransfersCommittedEvents(
       event => this.handleTransfersCommittedEvent(event),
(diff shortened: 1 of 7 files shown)


packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -32,26 +39,37 @@ import { RelayerFee } from '@hop-protocol/sdk'
...
-  resyncIntervalMs?: number
   gasCostPollEnabled?: boolean
 }
 
+type EventPromise = Array<Promise<any>>
+
 class SyncWatcher extends BaseWatcher {
   initialSyncCompleted: boolean = false
-  resyncIntervalMs: number = 60 * 1000
+  syncIntervalSec: number
```
miguelmota: suggestion: to keep things simple, the `syncIntervalSec` var here can be removed and only have `syncIntervalMs`

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -66,9 +84,24 @@ class SyncWatcher extends BaseWatcher {
...
+
+    if (this.syncIntervalMs > TenMinutesMs) {
+      this.logger.error('syncIntervalMs must be less than 10 minutes. Please use a lower multiplier')
+      this.quit()
+    }
+
+    const enabledNetworks = getEnabledNetworks()
+    for (const enabledNetwork of enabledNetworks) {
+      if (RelayableChains.includes(enabledNetwork)) {
+        this.isRelayableChainEnabled = true
```
miguelmota, line 102: a `break` can be added here

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -223,138 +256,199 @@ class SyncWatcher extends BaseWatcher {
...
-        l1Bridge.mapTransferSentToL2Events(
-          async (event: TransferSentToL2Event) => {
-            return await this.handleTransferSentToL2Event(event)
-          },
-          getOptions(l1Bridge.TransferSentToL2)
-        )
-      )
+  getTransferRootConfirmedEventPromise (): Promise<any> {
+    if (!this.isL1) {
+      return Promise.resolve()
```
miguelmota: suggestion: add `async` to this function and return empty array instead of `Promise.resolve()`


```suggestion
      return []
```

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -223,138 +256,199 @@ class SyncWatcher extends BaseWatcher {
...
-        l1Bridge.mapTransferRootBondedEvents(
-          async (event: TransferRootBondedEvent) => {
-            return await this.handleTransferRootBondedEvent(event)
-          },
-          getOptions(l1Bridge.TransferRootBonded)
-        )
-      )
+  getTransferSentToL2EventPromise (): Promise<any> {
+    if (!this.isL1) {
+      return Promise.resolve()
```
miguelmota: suggestion: add `async` to this function and return empty array instead of `Promise.resolve()` for consistent return values


```suggestion
      return []
```

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -223,138 +256,199 @@ class SyncWatcher extends BaseWatcher {
...
-        l1Bridge.mapTransferRootConfirmedEvents(
-          async (event: TransferRootConfirmedEvent) => {
-            return await this.handleTransferRootConfirmedEvent(event)
-          },
-          getOptions(l1Bridge.TransferRootConfirmed)
-        )
-      )
+  getTransferBondChallengedEventPromise (): Promise<any> {
+    if (!this.isL1) {
+      return Promise.resolve()
```
miguelmota: suggestion: add `async` to this function and return empty array instead of `Promise.resolve()`


```suggestion
      return []
```

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -223,138 +256,199 @@ class SyncWatcher extends BaseWatcher {
...
-        l1Bridge.mapTransferBondChallengedEvents(
-          async (event: TransferBondChallengedEvent) => {
-            return await this.handleTransferBondChallengedEvent(event)
-          },
-          getOptions(l1Bridge.TransferBondChallenged)
-        )
-      )
+  getTransferSentEventPromise (): Promise<any> {
+    if (this.isL1) {
+      return Promise.resolve()
```
miguelmota: suggestion: add `async` to this function and return empty array instead of `Promise.resolve()`


```suggestion
      return []
```

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -223,138 +256,199 @@ class SyncWatcher extends BaseWatcher {
...
-            return await Promise.all([
-              this.handleTransfersCommittedEvent(event)
-            ])
-          },
-          getOptions(l2Bridge.TransfersCommitted)
-        )
-      )
+  getTransfersCommittedEventPromise (): Promise<any> {
+    if (this.isL1) {
+      return Promise.resolve()
```
miguelmota: suggestion: add `async` to this function and return empty array instead of `Promise.resolve()`


```suggestion
      return []
```

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -223,138 +256,199 @@ class SyncWatcher extends BaseWatcher {
...
-      .then(async () => await this.availableLiquidityWatcher.syncBonderCredit())
+  getAllPromises (): EventPromise {
+    const asyncPromises: EventPromise = [
+      this.getTransferSentToL2EventPromise(),
+      this.getTransferRootConfirmedEventPromise(),
+      this.getTransferBondChallengedEventPromise(),
+      this.getTransferSentEventPromise(),
+      this.getTransferRootSetEventPromise(),
+    ]
+    const syncPromises: EventPromise = [
```
miguelmota, line 397: suggestion (nitpick): move `syncPromises` variable to be before `asyncPromises` variable

miguelmota (approved):

packages/hop-node/src/watchers/SyncWatcher.ts
shanefontaine: For all of these, the intention was to mimic the current array-of-non-awaited-promises (like [this](https://github.com/hop-protocol/hop/blob/develop/packages/hop-node/src/watchers/SyncWatcher.ts#L253C7-L269)). I'm not 100% sure it is correct though. Since all of these get executed `Promise.all()` do you know what would happen if async was added?

packages/hop-node/src/watchers/SyncWatcher.ts
shanefontaine: The intention for leaving it here was that `syncPromises` is used in the next line to construct `orderedPromises`, so the thought was to keep them logically together. But no super good reason other than that. What was the logic behind the suggestion? I realize a more clear name for the variable might be ideal here.

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -223,138 +256,199 @@ class SyncWatcher extends BaseWatcher {
...
   async syncHandler (): Promise<any> {
-    const promises: Array<Promise<any>> = []
+    // Events that are related to user transfers can be polled every cycle while
+    // all other, less time-sensitive events can be polled every N cycles.
+    // If SyncCyclesPerFullSync is 60, then the default values will be 30 seconds for the transferSent
+    // events and 30 minutes for the rest of the events.
+    let promisesPerPoll: EventPromise = []
+    if (
+      !this.isInitialSyncCompleted() ||
+      this.syncIndex % SyncCyclesPerFullSync === 0
```
miguelmota: suggestion: for readability maybe this check can be a separate method.

For me it wasn't clear how SyncCyclesPerFullSync worked until i read through the code

packages/hop-node/src/watchers/SyncWatcher.ts
shanefontaine: Resolved with f225ecad1e24f18ee4189d447ae5dcbf770f1827

packages/hop-node/src/watchers/SyncWatcher.ts
shanefontaine: Resolved with f225ecad1e24f18ee4189d447ae5dcbf770f1827

packages/hop-node/src/watchers/SyncWatcher.ts
shanefontaine: I think thats a good point. Let me know if you think 5edee2ffe9495b6e85e358fa04d495cd93c85b8f would resolve some of the confusion for future readers.

packages/hop-node/src/watchers/SyncWatcher.ts
miguelmota: Functionally there is no difference; the event map function (ie `getTransfersCommittedEventPromise`) are async functions so returning those in a function makes the function a promise, and adding `async` also makes the function a promise and anything returned by the function is always promisifed. 

Additionally the `await` in the `async event =>` function from old code in some places isn't functionally doing anything so it can be removed as well, e.g.

```ts
async getTransferSentToL2EventPromise(): Promise<any> {
  if (!this.isL1) return [];

  const l1Bridge = this.bridge as L1Bridge;
  return l1Bridge.mapTransferSentToL2Events(
    event => this.handleTransferSentToL2Event(event),
    this.getSyncOptions(l1Bridge.TransferSentToL2)
  );
}
```

packages/hop-node/src/watchers/SyncWatcher.ts
miguelmota: The reasoning was to order the variables in the order that they are used but your reasoning makes sense and can be left as is

packages/hop-node/src/watchers/SyncWatcher.ts
miguelmota: that looks good 👍

a step further would be something like this

```ts
if (this.shouldSyncAllEvents()) {
    promisesPerPoll = this.getAllPromises()
} else {
    promisesPerPoll = this.getTransferSentPromises()
}

// ...

shouldSyncAllEvents(): boolean {
    return !this.isInitialSyncCompleted() || this.isAtNewFullSyncCyclePoint()
}
  
isAtNewFullSyncCyclePoint(): boolean {
    return this.syncIndex % SyncCyclesPerFullSync === 0
}
```

shanefontaine: I like this. My mental model for this kind of stuff has been evolving lately and has been trending more towards this kind of architecture for readability. Good to get your perspective. Resolved with b6f11e9f9c4f906baa78361f9fd97cfba9820293

packages/hop-node/src/watchers/SyncWatcher.ts
shanefontaine: Thanks for the explanation, makes a ton of sense and I now understand better what it is doing.

~~IIUC, the empty return still needs to be a `Promise.resovle()` since the return statement expects a promise. For that reason, I left that piece of the code the same. LMK if you have any other thoughts around that.~~

Resolved with b6f11e9f9c4f906baa78361f9fd97cfba9820293

shanefontaine: My comment about the return statement was wrong. I now agree with your whole suggestion. Updated with 39a826ed09925d2327b63cfb629a3a523cb8bd37

PR #513, opened 2023-09-26
Feat/pre finality data

shanefontaine:
Adds syncing at the head of the chain. The SyncWatcher now observes `TransferSent` events at the head (if enabled) and writes them to the transfers DB with a new attribute: `isFinalized: false`. 

In the happy-path, that transfer will be bonded shortly after it is seen at the head. The transfer will then be observed later by the normal (finalized) `TransferSent` worker. This will then update `isFinalized: true`.

In the non-happy path, this transfer gets reorged out (or reorged to a later block). The `BondWithdrawalWatcher` will handle any reorg case and mark the transfer as `notFound: true`. If this happens, the transfer is effectively done. If the finalized worker eventually sees the same transfer, it will mark it as `notFound: false`, at which point it will be able to be bonded now with the full finalized context. If the finalized worker never sees it, that means it was reorged out and it is just stale data at this point.

One edge case note: a valid transfer may be reorged back one block and still have all the same Hop transfer properties (i.e. `transferId`) but only have different block properties (i.e. `blockNumber`. Even though this case could be handled and the bonder can bond it early, the tx is thrown away and handled when it becomes finalized. The timing of DB reads/writes makes this case inconsistent at any point of the tx lifecycle, including the mempool where it is out of our control. In practice, I don't expect this to happen (especially with ORUs, which is where this feature will be used), and the complexity to handle an impossible race condition is too hard to handle. Instead, if this does occur, the bond just takes until finality instead of happening quickly.

Features:

* Add `headSync` option to `chain` object config. If `true`, will sync the `TransferSent` event at head instead of finalized.
* Add experimental WS support that just logs data but does not write to the DB or anything critical.

_Note: You'll see some `TODO`s in there for me to handle after this is merged with the other branch. By default, this works the same as the existing bonder unless set in the config, so none of those `TODO`s matter while the bonder does not have this feature enabled._


packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -1686,6 +1742,53 @@ class SyncWatcher extends BaseWatcher {
       return true
     }
   }
+
+  // Experimental: Websocket support methods
+
+  initWebsocket = (contract: Contract, filter: EventFilter, cb: Function) => {
+    contract.on(filter, async (...event: any) => cb(event[event.length - 1]))
```
miguelmota, line 1749: I see here it's getting the last parameter as the event; for my understanding, what are the other parameters returned in the listener?

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -549,8 +580,34 @@ class SyncWatcher extends BaseWatcher {
...
+        dbData.withdrawalBondBackoffIndex = undefined
+        logger.debug('finalized transfer seen, resetting unfinalized error states')
+      }
+
+      await this.db.transfers.update(transferId, dbData)
+
+      // Experimental: compare data against WS cache and clear the memory
+      if (this.wsCache[transferId]) {
+        this.compareWsCache(event)
+        this.wsCache[event.args.transferId] = undefined
```
miguelmota: suggestion: use `delete` to remove reference from the object for potentially faster garbage collection

```suggestion
delete this.wsCache[event.args.transferId] 
```

packages/hop-node/src/watchers/SyncWatcher.ts
```diff
@@ -1686,6 +1742,53 @@ class SyncWatcher extends BaseWatcher {
       return true
     }
   }
+
+  // Experimental: Websocket support methods
+
+  initWebsocket = (contract: Contract, filter: EventFilter, cb: Function) => {
```
miguelmota: I don't think this needs to be an arrow function but if it does need to be one, then maybe adding a comment about it be good here

packages/hop-node/src/watchers/classes/Bridge.ts
```diff
@@ -580,6 +585,25 @@ export default class Bridge extends ContractBase {
         options.syncCacheKey
       )
       state = await this.db.syncState.getByKey(syncCacheKey)
+
+      const isHeadSync = syncCacheKey.includes(HeadSyncKeySuffix)
```
miguelmota: suggestion: use `startsWith` instead of `includes`

```suggestion
const isHeadSync = syncCacheKey.startsWith(HeadSyncKeySuffix)
```

packages/hop-node/src/watchers/classes/Bridge.ts
```diff
@@ -648,22 +672,29 @@ export default class Bridge extends ContractBase {
...
 
     let end: number
     let start: number
     let totalBlocksInBatch: number
     const { totalBlocks, batchBlocks } = globalConfig.sync[this.chainSlug]
-    let blockNumberWithAcceptableFinality: number
 
-    if (getHasFinalizationBlockTag(this.chainSlug)) {
+    // TODO: Better state handling
+    const isHeadSync = syncCacheKey?.includes(HeadSyncKeySuffix)
```
miguelmota: suggestion: use `startsWith` instead of `includes`

```suggestion
const isHeadSync = syncCacheKey?.startsWith(HeadSyncKeySuffix)
```

optional: a step further would be to move to separate function

```ts
const isHeadSync = this.isHeadSyncCacheKey(syncCacheKey)
```

miguelmota (approved):
The changes make a lot of sense and look good 👌

left very minor comments

packages/hop-node/src/watchers/SyncWatcher.ts
shanefontaine: `event` is the Event type generated by TypeChain. The interface and a practical example are below. There is likely a better way to handle it but I've never nailed down the best way to work with TypeChain so I did this. Happy to do something else if there is a more natural solution.

```

export interface TypedEventFilter<_TEvent extends TypedEvent>
  extends EventFilter {}


export interface TransferSentEventObject {
  transferId: string;
  chainId: BigNumber;
  recipient: string;
  amount: BigNumber;
  transferNonce: string;
  bonderFee: BigNumber;
  index: BigNumber;
  amountOutMin: BigNumber;
  deadline: BigNumber;
}
export type TransferSentEvent = TypedEvent<
  [
    string,
    BigNumber,
    string,
    BigNumber,
    string,
    BigNumber,
    BigNumber,
    BigNumber,
    BigNumber
  ],
  TransferSentEventObject
>;

export type TransferSentEventFilter = TypedEventFilter<TransferSentEvent>;
```

```
[
  '0xce6b702d315c03950030e5c2b008a2d8dab1d0af36c870cac76270a9e354a6ce',
  BigNumber { _hex: '0x05', _isBigNumber: true },
  '0xfEfeC7D3EB14a004029D278393e6AB8B46fb4FCa',
  BigNumber { _hex: '0x02c5ddeb108cfe52', _isBigNumber: true },
  '0x08dbc1e50017cd291b53a1e6ecea166ae6a2912e1e177b5020bf3c2e38bfd501',
  BigNumber { _hex: '0x237e4bc0f3a5e3', _isBigNumber: true },
  BigNumber { _hex: '0x01', _isBigNumber: true },
  BigNumber { _hex: '0x00', _isBigNumber: true },
  BigNumber { _hex: '0x00', _isBigNumber: true },
  {
    blockNumber: 15250984,
    blockHash: '0xfcea3a54861866acc699a5de406975b755f1f1a1d2ae17a4293f439756c2c34a',
    transactionIndex: 1,
    removed: false,
    address: '0x2708E5C7087d4C6D295c8B58b2D452c360D505C7',
    data: '0x00000000000000000000000000000000000000000000000002c5ddeb108cfe5208dbc1e50017cd291b53a1e6ecea166ae6a2912e1e177b5020bf3c2e38bfd50100000000000000000000000000000000000000000000000000237e4bc0f3a5e3000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000',
    topics: [
      '0xe35dddd4ea75d7e9b3fe93af4f4e40e778c3da4074c9d93e7c6536f1e803c1eb',
      '0xce6b702d315c03950030e5c2b008a2d8dab1d0af36c870cac76270a9e354a6ce',
      '0x0000000000000000000000000000000000000000000000000000000000000005',
      '0x000000000000000000000000fefec7d3eb14a004029d278393e6ab8b46fb4fca'
    ],
    transactionHash: '0x27aa876d282d5018363839988f2b753b87d8725f9987f025728a2f156c08438d',
    logIndex: 6,
    removeListener: [Function (anonymous)],
    getBlock: [Function (anonymous)],
    getTransaction: [Function (anonymous)],
    getTransactionReceipt: [Function (anonymous)],
    event: 'TransferSent',
    eventSignature: 'TransferSent(bytes32,uint256,address,uint256,bytes32,uint256,uint256,uint256,uint256)',
    decode: [Function (anonymous)],
    args: [
      '0xce6b702d315c03950030e5c2b008a2d8dab1d0af36c870cac76270a9e354a6ce',
      [BigNumber],
      '0xfEfeC7D3EB14a004029D278393e6AB8B46fb4FCa',
      [BigNumber],
      '0x08dbc1e50017cd291b53a1e6ecea166ae6a2912e1e177b5020bf3c2e38bfd501',
      [BigNumber],
      [BigNumber],
      [BigNumber],
      [BigNumber],
      transferId: '0xce6b702d315c03950030e5c2b008a2d8dab1d0af36c870cac76270a9e354a6ce',
      chainId: [BigNumber],
      recipient: '0xfEfeC7D3EB14a004029D278393e6AB8B46fb4FCa',
      amount: [BigNumber],
      transferNonce: '0x08dbc1e50017cd291b53a1e6ecea166ae6a2912e1e177b5020bf3c2e38bfd501',
      bonderFee: [BigNumber],
      index: [BigNumber],
      amountOutMin: [BigNumber],
      deadline: [BigNumber]
    ]
  }
]
```

packages/hop-node/src/watchers/classes/Bridge.ts
shanefontaine: Woah TIL about `startsWith`. In this case, the string is at the end so I'm going to change it to `endsWith`, but I like this a lot.

packages/hop-node/src/watchers/classes/Bridge.ts
shanefontaine: Going to keep for now with the thought that the next update to this logic should be a bigger update that will have more data about how the head sync is used in practice and will likely include ripping that out into its own function. (I don't love that it is just a suffix now, but it definitely works).

shanefontaine: Resolved with 2c22d36cd66155375efc707a3572fe8fda6fc1fc

packages/hop-node/src/watchers/classes/Bridge.ts
shanefontaine: Resolved with 2c22d36cd66155375efc707a3572fe8fda6fc1fc

packages/hop-node/src/watchers/SyncWatcher.ts
shanefontaine: Good call. Resolved with d6fbb6dc38f2876435995b243b1d2ac2218c2554

packages/hop-node/src/watchers/SyncWatcher.ts
miguelmota: I see, seems fine the way it is 👍

PR #524, opened 2023-10-23
UI updates

max-andrew:
- Remove the extra arrow on Safari
- Display Total Claimed on OP reward page
- Change buttons to connect wallet button on the FE when not connected
- Allow users to send transactions more quickly by enabling the `send` button immediately after they set the values for a new route


packages/frontend/src/pages/Send/Send.tsx
```diff
@@ -603,6 +605,17 @@ const Send: FC = () => {
     }
   }, [tx])
 
+  const [currentPathSending, setCurrentPathSending] = useState(sending)
```
shanefontaine: @max-andrew Is it possible to also handle the case where the path is the same but the user inputs a new value? As it stands, I believe a user would still only be able to send two transfers on the same route "slowly".

This was the original goal when I brought it up but I did not communicate that clearly in the spec.

packages/frontend/src/pages/Send/Send.tsx
```diff
@@ -603,6 +605,17 @@ const Send: FC = () => {
...
+
+  useEffect(() => {
+    const currentPath = `${fromNetwork?.slug}-${toNetwork?.slug}-${sourceToken?.symbol}`
```
shanefontaine: nitpick: possibly turn this into a `getCurrentPath()` util (1) so that it can handle the case where any of those values are `undefined`, and (2) to DRY up the code and avoid possible human error since this is used in `useSendTransaction.ts` as well.

WRT (1), I'm not sure how common that would be, if at all, but the case I am imagining is when a value is `undefined` for some reason, then even if the network is switched it will still have the same path string. For example, if `fromNetwork` is undefined, then one path would be `undefined-Optimism-ETH`, and if the user switched the source chain only, it would still be `undefined-Optimism-ETH`.

packages/frontend/src/pages/Send/useSendTransaction.ts
```diff
@@ -158,6 +159,8 @@ export function useSendTransaction (props: any) {
         })
       }
 
+      setLastPathSent(`${fromNetwork.slug}-${toNetwork.slug}-${sourceToken.symbol}`)
```
shanefontaine: nitpick: see comment above about util

shanefontaine (changes requested):
@max-andrew looks good! Left some nitpick comments and a little bit of scope creep.

Two high level things:
* Would it be possible to apply the send page updates to the Convert page as well, since they perform identical functionality? Specifically, the faster routes and the "connect wallet" button.
* Is it possible to simplify the transfer speed logic to just a `setCurrentTxSending(false)` (or some better name) in the `useEffect` where `setFromTokenAmount('')` is unset? One unknown in my head is when `tx` would be undefined in the `if (tx)` conditional 🤔  . This is more apparent with the context of faster transactions comment below when the user only changes the value instead (as opposed to just the route).

max-andrew:
On it thanks!

PR #532, opened 2023-11-13
Simplify chain additions for frontend/sdk

miguelmota:
This PR moves some hardcoded configs from sdk and frontend to the core package to centralize new chain configuration into one place. It should cut down significantly the amount of time it takes to add a new chain to frontend/sdk.

I moved around a few component files in frontend hence the large number of files changed, but there shouldn't be that many actual files to look at in terms of functionality changes. You can skip tsx files that are completely green since those were just renamed.


shanefontaine:
@miguelmota LGTM!

PR #534, opened 2023-11-30
Feat/level config optimization

shanefontaine:
DB refactor for optimized writes and more native usage of Level. Some notable changes:

* Remove our `mutex` from the DB since Level supposedly [handles this already in an optimized way](https://github.com/google/leveldb/blob/main/include/leveldb/options.h#L75-L83). Increased the Level-mutex size from default 4MB to 8MB.
* Use an [iterator instead](https://github.com/Level/abstract-level?tab=readme-ov-file#iterator--dbiteratoroptions) of read stream for times when entire DB is being acted upon. Introduce ability to pass in callback so that each item is acted on in-line instead of pushed to memory and acted on after.
* Migrations can now be run on any DB with just a config file. Migrations are persistent and only run once per bonder without the need to remove the migration logic after the bonder has updated.
* Removed the concept of `tilReady` from the DB. From the POV of the DB, it should not care whether or not it is ready. For example, the migrations should be able to read/write freely without readiness checks. Expose an `isReady()` function that the consumer can call. I believe this is the expected way to handle this to avoid adding `tilReady()` to the beginning of every db method.
  * Introduced `isDbSetReady()` that is called by the watchers & is used to wait to begin polling.
* (Mostly) Consistently use the terminology:
  * `item`: a single key and value pair
  * `key`: key of an item
  * `value`: value of an item
  * `entry`: unused, since it is the same as `item` in this case (I believe)


packages/hop-node/src/db/SyncStateDb.ts
```diff
@@ -1,40 +1,24 @@
...
+  async update (key: string, value: State): Promise<void> {
+    await this.put(key, value)
   }
 
-  normalizeValue (key: string, value: State) {
-    if (value) {
-      value.key = key
+  async getByKey (key: string): Promise<State | null> {
+    const value: State | null = await this.get(key)
+    if (!value) {
```
miguelmota: I don't think this check and null return is necessary if `this.get(key)` already returns null for an item that is not found, but i could be wrong

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -133,84 +127,67 @@ export type ChallengeableTransferRoot = {
...
-  }
-
-  async insertItem (transferRoot: TransferRoot) {
-    const { transferRootId } = transferRoot
-    const logger = this.logger.create({ id: transferRootId })
+  async update (transferRootId: string, transferRoot: TransferRoot): Promise<void> {
     const key = this.getTimestampedKey(transferRoot)
     if (!key) {
+      // Can occur if an event has been missed
+      this.logger.debug(`key not found for transferRootId: ${transferRootId}`)
```
miguelmota: minor suggestion: mention comment in log

packages/hop-node/src/cli/dbDump.ts
```diff
@@ -128,8 +129,6 @@ async function main (source: any) {
         toUnix: toDate
       })
     }
-  } else if (dbName === 'sync-state') {
-    items = await db.syncState.getItems()
```
miguelmota, line 132: curious, what would be the recommended way to dump syncState and gasCost items if needed for debugging?

packages/hop-node/src/db/BaseDb.ts
```diff
@@ -92,351 +111,297 @@ class BaseDb extends EventEmitter {
...
+      this.logger.info('db ready')
+    }
+  }
+
+  async #init (migrations: Migration[]): Promise<void> {
+    const databaseMigrator = new DatabaseMigrator<T>(this)
+    const metadata = await this.#getMetadata()
+    const migrationIndex = metadata?._migrationIndex ?? 0
+    this.isMigrating = true
+    databaseMigrator.migrate(migrations, migrationIndex)
```
miguelmota: no strong opinion; I think it might be cleaner to instantiate the DatabaseMigrator with the list of migrations

```ts
new DatabaseMigrator<T>({ db: this, migrations })
```

and then only call migrate with index, ie `databaseMigrator.migrate(migrationIndex)`

packages/hop-node/src/watchers/classes/Bridge.ts
```diff
@@ -611,24 +611,27 @@ export default class Bridge extends ContractBase {
...
-        //   for some time, and then turn it back on again.
-        const finalizedStateKey = syncCacheKey.replace(customSyncKeySuffix, '')
-        const finalizedState = await this.db.syncState.getByKey(finalizedStateKey)
-
-        const doesCustomSyncDbExist = !!state?.latestBlockSynced
-        const isCustomSyncDataStale = state?.latestBlockSynced < finalizedState.latestBlockSynced
-        if (!doesCustomSyncDbExist || isCustomSyncDataStale) {
-          state = finalizedState
-          state.key = syncCacheKey
+      state = (await this.db.syncState.getByKey(syncCacheKey)) ?? undefined
```
miguelmota: curious, why is the `?? undefined` needed here?

packages/hop-node/src/db/BaseDb.ts
```diff
@@ -92,351 +111,297 @@ class BaseDb extends EventEmitter {
...
-    if (this.listeners(Event.Error).length > 0) {
-      this.emit(Event.Error, err)
+  #logDbOperation (operation: string, logOptions: LogOptions): void {
+    const { key, value, logMsg, batchOperations } = logOptions
+
+    let log = `DB Operation: ${operation}`
+
+    // Warn the consumer that the DB is not yet ready
+    if (!this.ready && !this.isMigrating) {
+      log += ', DB is not ready and this operation may cause unexpected results'
```
miguelmota: suggestion: instead of concatenating with `+=`,  a cleaner way would be to push to an array and then join the array, ie `this.logger.debug(log.join(','))`

packages/hop-node/src/db/db.ts
```diff
@@ -28,33 +27,36 @@ export function getDbSet (tokenSymbol: string): DbSet {
...
+    transfers: dbSets.transfersDb[tokenSymbol],
+    transferRoots: dbSets.transferRootsDb[tokenSymbol],
+    gasCost: dbSets.gasCostDb[tokenSymbol]
+  }
+}
+
+export function isDbSetReady (tokenSymbol: string): boolean {
+  for (const dbSet of Object.values(dbSets)) {
+    const db = dbSet[tokenSymbol]
+    if (!db.isReady()) {
```
miguelmota, line 55: suggestion: do `return db.isReady()` here since it's a boolean

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -292,94 +266,76 @@ class TransferRootsDb extends BaseDb {
...
-
-  private readonly filterTimestampedKeyValues = (x: any) => {
-    return x?.value?.transferRootId
-  }
-
-  private readonly sortItems = (a: any, b: any) => {
-    return a?.committedAt - b?.committedAt
+  async getByTransferRootId (transferRootId: string): Promise<TransferRoot | null> {
+    const item = await this.get(transferRootId)
+    if (!item) {
```
miguelmota: same comment here about removing the conditional here if `.get()` already returns null when not found

packages/hop-node/src/watchers/RelayWatcher.ts
```diff
@@ -199,6 +198,8 @@ class RelayWatcher extends BaseWatcher {
       logger.info(msg)
       this.notifier.info(msg)
     } catch (err: any) {
+      logger.debug('sendTransferRelayErr err:', err.message)
```
miguelmota: suggestion: log error object to see full error

```ts
logger.debug('sendTransferRelayErr err:', err)
```

packages/hop-node/src/db/TransfersDb.ts
```diff
@@ -612,6 +482,90 @@ class TransfersDb extends BaseDb {
...
+          // onchain transfer sent index always starts at 1
+          if (transfer?.transferSentIndex === 1) {
+            return transferIds
+          }
+        }
+      }
+    }
+  }
+
+  protected normalizeTransferItem (item: Transfer): Transfer {
```
miguelmota: suggestion: rename to `normalizeTransferValue` to be consistent with `normalizeDbValue` (or vice versa) since they both normalize values returned by `this.get(...)`, but i could be misunderstanding the naming convention

packages/hop-node/src/db/GasCostDb.ts
```diff
@@ -25,92 +24,115 @@ type GasCost = BaseItem & {
...
-    const items: GasCost[] = (await this.getItems(filter)).filter((item: GasCost) => {
-      return (
-        item.chain === chain &&
-        item.token === token &&
-        item.transactionType === transactionType &&
-        item.timestamp
+
+    const isRelevantItem = (key: string, value: GasCost): GasCost | null => {
+      const isRelevant = (
+        !!(value.chain === chain) &&
```
miguelmota: seems redundant to coerce to a boolean when the equality check is already a boolean

packages/hop-node/src/db/TransferRootsDb.ts
```diff
@@ -220,49 +197,47 @@ class SubDbIncompletes extends BaseDb {
...
+    // this method to conform with the interface
+    await this.insertIfNotExists(transferRootId, { transferRootId: transferRoot.transferRootId })
   }
 
-  async getByTransferRootHash (transferRootHash: string) {
-    const item = await this.getById(transferRootHash)
-    return item?.transferRootId
+  async getTransferRootId (transferRootHash: string): Promise<string | null> {
+    const item = await this.get(transferRootHash)
+    if (!item?.transferRootId) {
```
miguelmota, line 232: same comment here about removing the conditional here if .get() already returns null when not found

packages/hop-node/src/db/BaseDb.ts
```diff
@@ -92,351 +111,297 @@ class BaseDb extends EventEmitter {
...
-      this.batchQueue = []
-      this.lastBatchUpdatedAt = Date.now()
-      this.logger.debug(`attempting batch write, items: ${ops?.length} `)
-      await this.putBatch(ops)
+  /**
+   * Custom DB Operations - All items
+   */
+
+  protected async upsertAll (filters?: DbItemsFilter<T>): Promise<void> {
+    if (filters?.cbFilterGet) {
```
miguelmota: would having a specific `DbUpsertAllFilter` type be better here? so that `cbFilterGet` and `cbFilterPut` aren't shown as an option, and can be passed as `this.#processItems(filters as DbItemsFilter)` if it makes sense

packages/hop-node/src/db/BaseDb.ts
```diff
@@ -92,351 +111,297 @@ class BaseDb extends EventEmitter {
...
-  protected async _getUpdateData (key: string, data: any) {
-    const entry = await this.getById(key, {
-      _createdAt: Date.now()
-    })
-    const value = Object.assign({}, entry, data)
-    return { key, value }
-  }
+  protected async getKeys (filters?: DbItemsFilter<T>): Promise<string[]> {
+    if (filters?.cbFilterPut) {
+      throw new Error('cbFilterPut cannot be used with getKeys')
```
miguelmota: same comment as above about using specific type for filter param

packages/hop-node/src/db/BaseDb.ts
```diff
@@ -92,351 +111,297 @@ class BaseDb extends EventEmitter {
...
   }
 
-  async _updateSingle (key: string, data: any) {
-    return this.mutex.runExclusive(async () => {
-      const { value } = await this._getUpdateData(key, data)
-      return this.db.put(key, value)
-    })
+  protected async getValues (filters?: DbItemsFilter<T>): Promise<T[]> {
+    if (filters?.cbFilterPut) {
+      throw new Error('cbFilterPut cannot be used with getValues')
```
miguelmota: same comment as above about using specific type for filter param

packages/hop-node/src/db/BaseDb.ts
```diff
@@ -92,351 +111,297 @@ class BaseDb extends EventEmitter {
...
-      const groups = groupBy(ops, 'key')
-      const keys = Object.keys(groups)
-      const groupedOps = Object.values(groups).map((items: any[], i) => {
-        const value = spread(merge)(items.map(this.getValue))
-        return {
-          type: 'put',
-          key: keys[i],
-          value
+        // Handle put filter, if exists
+        if (filters?.cbFilterPut) {
```
miguelmota, line 278: minor suggestion: do function type check `if (typeof filters?.cbFilterPut === 'function') {`

miguelmota (approved):

packages/hop-node/src/db/TransfersDb.ts
shanefontaine: That is correct understanding. Updated.

packages/hop-node/src/db/TransferRootsDb.ts
shanefontaine: leaving this one since the conditional is checking for the existence of `transferRootId` on the item.

packages/hop-node/src/db/BaseDb.ts
shanefontaine: Isn't this enforced by the type?

`export type CbFilterPut<T> = (key: string, value: T) => Promise<void>`

packages/hop-node/src/db/db.ts
shanefontaine: that would return the first db status instead of iterating over all dbs in the set and returning `true` only if all are ready

packages/hop-node/src/cli/dbDump.ts
shanefontaine: I had removed it when I was under the impression that a dump required all data in memory, which is not true. I will add it back in now that iterator is implemented and better understood.

In general, my new mental model of key-value DBs is that reads/writes of the entire DB should be very rare and very explicit since, in a perfect world, everything is "indexed" in separate DBs/SubDBs (overkill for now). Migrations, dbDump CLI, and a few others make sense, but arbitrary get alls are very resource intensive if not considered carefully. IMO we are good to go and nothing else is needed, but just wanted to call it out as a reason why `getAll...` is relatively tightly scoped.

packages/hop-node/src/db/BaseDb.ts
shanefontaine: I like it & the object input.

packages/hop-node/src/watchers/classes/Bridge.ts
shanefontaine: Too lazy to change 2 lines of code :)

Updated.

packages/hop-node/src/db/BaseDb.ts
shanefontaine: Yea I think that is a good call. Added.

packages/hop-node/src/db/db.ts
miguelmota: you're right, i missed seeing the for loop

packages/hop-node/src/db/BaseDb.ts
miguelmota: it's enforced by type checker but not in runtime. Good to leave as is

packages/hop-node/src/db/TransferRootsDb.ts
miguelmota: sorry missed that, seems good

packages/hop-node/src/cli/dbDump.ts
miguelmota: understood :+1:

PR #535, opened 2023-11-30
hop node config abstractions

miguelmota:
- moves some hop node config to core
- reads as much as possible from core
- removes `as any` 

changes should be pretty straightforward, but let me know if you see any logic errors


shanefontaine (approved):
@miguelmota looks good. Only item that didn't appear super clear is the naming of the `providers` metadata file in `core`. How it is used now makes sense, but I think I'd hesitate to use it in a broader context because of the confusion with ethers providers. One suggestion would be to update the file & naming convention to be `rpcProviders` instead of `providers` as a way to differentiate between this and ethers. I don't think that is a _great_ solution, but does clarify in my head.

Other than that, everything LGMT.

miguelmota:
@shanefontaine good suggestion and agree, will rename 👍

PR #537, opened 2023-12-06
Feat/standardized message status

miguelmota:
Changes:
- uses a simple factory to create chain bridge classes for more concise code, and removes need to pass chainSlug around.
- moves cache functions out from `AbstractChainBridge` to it's own `CacheService` (can be named something else). `AbstractService` extends `CacheService`. The reason is that some services need it's own cache.
- uses "parameter properties" typescript shortand where possible for more concise code.
- changes default exports to be named exports. Makes it easier to not think about import naming and ordering.
- adds check for `CALL_EXCEPTION` in the rateLimitRetry and it won't retry if it encounters that error. This should remove the need to use separate non-retriable providers, since this error is thrown by ethers when on-chain reads fail because the contract threw error.

Notes:
- there were various merge conflicts when merging develop. Please review the business logic to make sure things weren't merged incorrectly, and since you know the implementation code best.
- hasn't been tested on running bonder.
- open for discussion on any of the changes since i could be misunderstanding things


miguelmota:
@shanefontaine ready for review

PR #539, opened 2023-12-15
Feat/chains test updates

shanefontaine:


packages/hop-node/src/chains/Chains/polygonzk/PolygonZkMessageService.ts
```diff
@@ -26,46 +26,41 @@ const polygonSdkVersion: Record<string, string> = {
...
 
-    this.#init(l1Network)
+    this.#init(networkSlug)
       .then(() => {
         this.ready = true
         this.logger.debug('zkEVM client initialized')
       })
       .catch((err: any) => {
         this.logger.error('zkEvmClient initialize error:', err)
+        throw err
```
miguelmota: I think the `throw` here will result in an uncaught promise error in nodejs, so maybe exiting with `process.exit(1)` would be better

Some recommendations from chatgpt

- Event Emitter: Use an event emitter to emit an error event from the constructor and handle it outside the class.
- Initialization Method: Instead of initializing in the constructor, provide an explicit initialization method that returns a Promise. This allows the calling code to handle initialization asynchronously and catch any errors.
- Factory Function: Create a static factory method that handles the asynchronous initialization and returns a Promise of the instance.

miguelmota (approved):
Looks great 👍 nice work on making it more DRY

packages/hop-node/src/chains/Chains/polygonzk/PolygonZkMessageService.ts
shanefontaine: Good call, updated to `process.exit(1)`

PR #540, opened 2023-12-22
Feat/template method pattern

shanefontaine:
@miguelmota one more purely architectural change for easier addition of chains.

* Correctly implements template design pattern. Now avoids child message service (i.e. ArbitrumMessageService.ts) from having to know that it has to call the abstract class `validateMessageAndSendTransaction()` in order to send a tx. Abstract class now handles everything with the child class just needing to tell it how to send a tx.
* More explicit params throughout (but open to using an `opts` object later when more chains introduce more needs)


miguelmota (approved):
changes lgtm 👍

PR #541, opened 2023-12-28
Feat/lint updates

shanefontaine:
See doc for PR info. Changes are only concerned with `hop-node` just to get on the same page but can probably be applied to all packages if/when appropriate.

## Meta

* Use `es2021` throughout since it is supported by all [browsers/servers in 2023](https://www.learningtypescript.com/articles/why-increase-your-tsconfig-target)
* Update linting, formatting and compilation rules to fit our style
  * From a high level, the rules were set by following best practices and by observing how we code. For the most part, no meaningful changes have been made that would affect code style.

## Versioning

* Update `typescript` to latest version
  * Unlocks latest features like decorators
  * [Meta](https://kinsta.com/blog/typescript-5-0/)
    * 60% smaller build
    * 80-90% faster build/startup of typescript
* Update `eslint` to latest version 
  * Latest features, best practices, rulesets
  * Compatibility - our version of `node`  is the last version supported by eslint `<8` and will probably be deprecated in `9`
* Update `node` to LTS (`20.10.0`)
  * Latest ecmascript features like decorators
  * No more need for `--legacy-peer-deps`
  * No more need for `node-fetch` since fetch is native in node since `v17.5` and is stable as of `v21` (and basically stable in `v20`. `V20` specifically optimized fetch performance)
  * `heapdump` now comes with native v8 package
  * Security updates for [16 stopped in Oct 2023, 20 stops >Jul 2026](https://nodejs.org/en/about/previous-releases)
  * Meta
    * Update `.nvmrc` to latest LTS version (`20.10.0`)
    * Remove `.nodeversion` since that only deals with `nodenv`, which I do not believe is used

## Other

* Pin dependencies
  * Allows for removal of `--no-cache` on docker image build to save from `npm i` on each build unless `package.json` or `package-lock.json` changes.
* Removed `@types` dir. IMO allows for simpler `tsconfig` since all types are covered by default and no risk of bad overwrite.
  * Some packages introduced types since we first looked, others are now ignored with `// @ts-expect-error`


shanefontaine:
Closed in favor of #545

PR #545, opened 2024-01-10
Feature/eslint config package

miguelmota:
- adds shared `.eslintrc.js` and `tsconfig.json` to the monorepo root for the packages to extend and use. The only additional configuration is in `frontend` that also extends eslint react plugins and rules.
- removes unused packages `ipfs-worker/`, and `sdk-demo-api/` (in favor of `api/` which is basically the same thing)
- most of the modified files are due to eslint `--fix` changes
- merges [PR 541](https://github.com/hop-protocol/hop/pull/541) into this one


packages/eslint-config-hop-protocol/index.js
```diff
@@ -0,0 +1,116 @@
...
+  "env": {
+    "node": true,
+    "browser": true,
+    "es2021": true,
+    "jest": true
+  },
+  "extends": [
+    "eslint:recommended",
+    "plugin:json/recommended",
+    "standard"
```
shanefontaine: I hesitantly removed `standard` as I think it is "old" at this point (though still useful). Feross has moved on and there is some [fighting / schism](https://github.com/standard/standard/issues/1948) going on. My goal was to update most of this stuff to be "modern" and thought that removing this was a good choice.

With that said, this is a blind spot for me. If using `standard` adds little or no overhead and allows broader support, them in 100% for it. I don't have enough knowledge here to make the call, so happy with whatever you think.

packages/eslint-config-hop-protocol/index.js
```diff
@@ -0,0 +1,116 @@
+module.exports = {
+  "env": {
+    "node": true,
+    "browser": true,
+    "es2021": true,
```
shanefontaine: Just a flag to discuss: no opinion or knowledge at all if this is ideal (same with `parserOptions.ecmaVersion`. My understanding is that this just allows for newer JS features and maybe some optimizations, but closes the door for deprecated features. IIUC, we neither use the new features nor the deprecated ones so it really doesn't matter.

Given that, `es2021` seemed like the right `LTS` one, but very low knowledge here.

packages/eslint-config-hop-protocol/package.json
```diff
@@ -0,0 +1,32 @@
...
+  },
+  "author": "",
+  "license": "MIT",
+  "devDependencies": {
+    "@typescript-eslint/eslint-plugin": "^6.18.1",
+    "@typescript-eslint/parser": "^6.18.0",
+    "eslint": "^8.56.0",
+    "eslint-config-standard": "^16.0.2",
+    "eslint-plugin-import": "^2.29.1",
+    "eslint-plugin-jest": "^27.6.1",
```
shanefontaine: I believe this is unused. If it is meant to be used in a child package, I would expect the child to install it.

packages/eslint-config-hop-protocol/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "eslint-plugin-json": "^3.1.0",
+    "eslint-plugin-jsx-a11y": "^6.4.1",
```
shanefontaine: I believe this is unused. If it is meant to be used in a child package, I would expect the child to install it.

packages/eslint-config-hop-protocol/index.js
```diff
@@ -0,0 +1,116 @@
+module.exports = {
+  "env": {
+    "node": true,
+    "browser": true,
```
shanefontaine: I believe this should go only in `frontend` since the other packages shouldn't use the browser api.

packages/eslint-config-hop-protocol/index.js
```diff
@@ -0,0 +1,116 @@
+module.exports = {
+  "env": {
+    "node": true,
```
shanefontaine: Technically I believe this should go in all packages except `frontend`, since they'll all be run in the node context, I believe? If that is true, then I _loosley_ think this should be removed from the top-level, but also recognize that it covers like 90% of cases if it exists here, so would be fine here as well.

packages/eslint-config-hop-protocol/index.js
```diff
@@ -0,0 +1,116 @@
...
+    "jest": true
+  },
+  "extends": [
+    "eslint:recommended",
+    "plugin:json/recommended",
+    "standard"
+  ],
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": {
+    "ecmaFeatures": {
```
shanefontaine: I believe this should live only in the `frontend`

.circleci/config.yml
```diff
@@ -68,7 +68,6 @@ jobs:
             bash ./ci/package_check.sh
             docker build \
               -f Dockerfile \
```
shanefontaine: (See high level comment about pinning deps)

I looked into this before and was worried that by removing it, we might get stuck on some bug because an NPM package doesn't use the version we expected. For example, if our deps are not pinned and we implicitly expect a new build to use the latest version of a package, I _think_ this wouldn't use the latest and we might have a hard time debugging.

However, my understanding is that this speeds up the build a ton, so not sure what the best answer is. I think pinning deps, but up for anything.

packages/api/package.json
```diff
@@ -20,7 +20,15 @@
     "memory-cache": "^0.2.0"
   },
   "devDependencies": {
+    "@babel/eslint-parser": "^7.23.3",
     "@types/express": "^4.17.13",
-    "standard": "^17.0.0"
+    "@typescript-eslint/eslint-plugin": "^6.18.1",
+    "@typescript-eslint/parser": "^6.18.1",
+    "eslint": "^8.56.0",
```
shanefontaine: I believe these `eslint` packages that are defined in the top level are not required here, but could be wrong.

packages/api/.eslintrc.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": [
+    "eslint-config-hop-protocol"
+  ],
+  "parser": "",
+  "parserOptions": {
+    "requireConfigFile": false,
```
shanefontaine: According to chat, we only need to specify the different option here and nothing else.

```
"extends": [
    "eslint-config-hop-protocol"
  ],
  "parserOptions": {
    "requireConfigFile": false
  }
]
```

packages/frontend/asdf
shanefontaine, line 1: :)

packages/eslint-config-hop-protocol/index.js
```diff
@@ -0,0 +1,114 @@
...
+      "jsx": true
+    },
+    "ecmaVersion": 2021,
+    "sourceType": "module",
+    "project": "./tsconfig.json"
```
shanefontaine: See `package.json` comment about the `node` plugin. IIUC, replace this with `n`.

packages/eslint-config-hop-protocol/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "license": "MIT",
+  "devDependencies": {
+    "@typescript-eslint/eslint-plugin": "^6.18.1",
+    "@typescript-eslint/parser": "^6.18.0",
+    "eslint": "^8.56.0",
+    "eslint-config-standard": "^16.0.2",
+    "eslint-plugin-jest": "^27.6.1",
+    "eslint-plugin-json": "^3.1.0",
+    "eslint-plugin-jsx-a11y": "^6.4.1",
+    "eslint-plugin-node": "^11.1.0",
```
shanefontaine: I believe `eslint-plugin-node` is [unmaintained](https://github.com/mysticatea/eslint-plugin-node/issues/300) and has been replaced with [`eslint-plugin-n`](https://github.com/eslint-community/eslint-plugin-n).

> forked from [eslint-plugin-node](https://www.npmjs.com/package/eslint-plugin-node) v11.1.0. as the original repository seems https://github.com/mysticatea/eslint-plugin-node/issues/300.

In that case, we can use `eslint-plugin-n`.

packages/eslint-config-hop-protocol/index.js
```diff
@@ -0,0 +1,114 @@
...
+    "node",
+    "unused-imports",
+    "sort-imports-es6-autofix"
+  ],
+  "rules": {
+    // Explicit offs
+    "@typescript-eslint/consistent-type-definitions": 0,
+    "@typescript-eslint/explicit-function-return-type": 0,
+    "@typescript-eslint/method-signature-style": 0,
+    "@typescript-eslint/no-dynamic-delete": 0,
```
shanefontaine: These are all off by default. `no-dynamic-delete` is on with `strict`. but we don't enable that at this time.

```suggestion
```

packages/eslint-config-hop-protocol/index.js
```diff
@@ -0,0 +1,114 @@
...
+    "@typescript-eslint/no-misused-promises": 0,
+    "@typescript-eslint/no-non-null-assertion": 0,
```
shanefontaine: Off by default.

```suggestion
```

packages/eslint-config-hop-protocol/index.js
```diff
@@ -0,0 +1,114 @@
...
+    "@typescript-eslint/no-unsafe-enum-comparison": 0,
+    "@typescript-eslint/require-await": 0,
+    "@typescript-eslint/restrict-template-expressions": 0,
+    "@typescript-eslint/strict-boolean-expressions": 0,
```
shanefontaine: Off by default.

```suggestion
```

packages/eslint-config-hop-protocol/index.js
```diff
@@ -0,0 +1,114 @@
...
+    "array-callback-return": 2,
+    "block-scoped-var": 2,
+    "dot-notation": 2,
+    "new-cap": [2, { "properties": false }], // Used by ethers event filters
+    "no-empty": [2, { "allowEmptyCatch": true }],
+    "no-new": 2,
+    "prefer-const": [2, { "destructuring": "all" }],
+    "node/no-exports-assign": 2,
+    "node/no-new-require": 2,
+    "node/no-path-concat": 2,
```
shanefontaine: `no-exports-assign` is `2` by default, update others with new `n` package.

```suggestion
    "n/no-new-require": 2,
    "n/no-path-concat": 2,
```

packages/eslint-config-hop-protocol/index.js
```diff
@@ -0,0 +1,114 @@
...
+    "no-multi-spaces": 0,
+    "no-multiple-empty-lines": 0,
+    "no-trailing-spaces": 0,
+    "no-underscore-dangle": 0,
+    "no-unreachable": 0,
+    "no-unreachable-loop": 0,
+    "no-unused-vars": 0,
+    "no-use-before-define": 0,
+    "node/handle-callback-err": 0,
+    "node/no-callback-literal": 0,
```
shanefontaine: Update with `n` package

```suggestion
    "n/handle-callback-err": 0,
    "n/no-callback-literal": 0,
```

packages/eslint-config-hop-protocol/index.js
```diff
@@ -0,0 +1,114 @@
...
+    "@typescript-eslint/explicit-function-return-type": 0,
+    "@typescript-eslint/method-signature-style": 0,
+    "@typescript-eslint/no-dynamic-delete": 0,
+    "@typescript-eslint/no-misused-promises": 0,
+    "@typescript-eslint/no-non-null-assertion": 0,
+    "@typescript-eslint/no-unsafe-enum-comparison": 0,
+    "@typescript-eslint/require-await": 0,
+    "@typescript-eslint/restrict-template-expressions": 0,
+    "@typescript-eslint/strict-boolean-expressions": 0,
+    "arrow-body-style": 0,
```
shanefontaine: default
```suggestion
```

packages/eslint-config-hop-protocol/index.js
```diff
@@ -0,0 +1,114 @@
...
+    "avoidEscape": 0,
+    "camelcase": 0,
+    "comma-dangle": 0,
+    "indent": 0,
+    "multiline-ternary": 0,
```
shanefontaine: deprecated
```suggestion
```

packages/eslint-config-hop-protocol/index.js
```diff
@@ -0,0 +1,114 @@
...
+    "no-async-promise-executor": 0,
+    "no-constant-condition": 0,
+    "no-multi-spaces": 0,
+    "no-multiple-empty-lines": 0,
+    "no-trailing-spaces": 0,
+    "no-underscore-dangle": 0,
```
shanefontaine: Deprecated or default in recommended

```suggestion
```

packages/eslint-config-hop-protocol/index.js
```diff
@@ -0,0 +1,114 @@
...
+    "@typescript-eslint/require-await": 0,
+    "@typescript-eslint/restrict-template-expressions": 0,
+    "@typescript-eslint/strict-boolean-expressions": 0,
+    "arrow-body-style": 0,
+    "avoidEscape": 0,
+    "camelcase": 0,
+    "comma-dangle": 0,
+    "indent": 0,
+    "multiline-ternary": 0,
+    "no-async-promise-executor": 0,
```
shanefontaine: IMO rule should be followed (ignore this suggestion if too much breaks).

```suggestion
```

packages/eslint-config-hop-protocol/index.js
```diff
@@ -0,0 +1,114 @@
...
+    "no-constant-condition": 0,
+    "no-multi-spaces": 0,
+    "no-multiple-empty-lines": 0,
+    "no-trailing-spaces": 0,
+    "no-underscore-dangle": 0,
+    "no-unreachable": 0,
```
shanefontaine: default off
```suggestion
```

packages/eslint-config-hop-protocol/index.js
```diff
@@ -0,0 +1,114 @@
...
+    "no-unreachable-loop": 0,
+    "no-unused-vars": 0,
+    "no-use-before-define": 0,
```
shanefontaine: default off
```suggestion
```

packages/eslint-config-hop-protocol/index.js
```diff
@@ -0,0 +1,114 @@
...
+    "node/handle-callback-err": 0,
+    "node/no-callback-literal": 0,
+    "padded-blocks": 0,
+    "quotes": 0,
+    "semi": 0,
+    "sort-imports": 0,
+    "space-before-function-paren": 0,
+    "spaced-comment": 0,
```
shanefontaine: Deprecated or off by default

```suggestion
```

shanefontaine (changes requested):
Note that some of the code suggestions below might be incorrect (i.e. default rules). I didn't _try_ the changes but plan on doing a big pass before merge, so I apologize if anything is unclear.

The topics below are my remaining high level thoughts.

# Package Management

From a high level, my current mental model to just use modern practices and LTS versions plus reference similar projects. The reference projects are [Optimism](https://github.com/ethereum-optimism/optimism), [Standardnotes](https://github.com/standardnotes/server), [Typescript](https://github.com/microsoft/TypeScript), [VSCode](https://github.com/microsoft/vscode), [Arbitrum](https://github.com/OffchainLabs/arbitrum-sdk/) (no monorepo), [Uniswap](https://github.com/Uniswap) (No monorepo).

### NPM Workspaces vs Lerna

I think it makes sense to use NPM `workspaces` instead of Lerna. This is a shift in thinking but I think is the modern way of doing things and how we do it now is antiquated and may cause us support issues in the future.

In 2023, Lerna [changed](https://lerna.js.org/docs/legacy-package-management) the way their model works to take advantage of the built-in workspaces of each package manger. They now use the native package workspaces and just build on top of it. **My understanding is that both workspaces and (modern) lerna now have `node_modules` at the root**. 

> For the final several years of his tenure as lead maintainer of lerna, Daniel, had been encouraging folks to strongly reconsider their use of the legacy package management commands in lerna, and instead leverage their package manager of choice to do what it does best.
> 

Given that, I think it makes sense to keep it simple and stick with the native package manager and scrap lerna (I do not believe what they offer on top is worth the additional layer of complexity/api/etc.). One other point is that none of the projects I’ve been referencing for all this use lerna. Though (1) I may be wrong and (2) this may just be confirmation bias.

### NPM Workspaces Flow

After research (but no practice), I think `npm workspaces` works very similarly to how we operate now. Some changes in thinking:

- Everything is run from the root (with native commands to do so)
- Monorepo-wide config can live at the root and people can go here to get an overall picture. Stuff like compiler configs, linters, ignores, etc. Each package still gets its own unique items if they need it. Reduces redundant files throughout each package.
- Some different config files that help the mapping (not new, just different format than lerna)
- Root level dependencies are shared among all packages (reference [Optimism](https://github.com/ethereum-optimism/optimism/blob/develop/package.json)). These should be stuff like:
    - frameworks (Jest)
    - build tools (Webpack)
    - Lint tools (eslint)
    - Formatters (Prettier)
- Packages should install their own deps that they depend on to be published

# ESLint updates

*This section assumes NPM workspaces is used and does not apply if it is not.*

I suggest we remove the `eslint-config-hop-protocol` and move the rules to the root dir. We can also introduce an `.eslintignore` and other root-level config files that the individual packages shouldn’t need to care about. This matches what I believe are best practices as well as all the reference projects (I believe).

This also seems logical if we hoist `tsconfig` up to the root as well, since they work together. Then the entire TS config lives at the root and is available to all who need it. I believe any user of a published package should not need to know about or be concerned with our packages compiler or lint options.

**With that said, I do like the thought of the separate hop lint package**, however, I don’t like the complexity/mental overhead of maintaining it, as I think we should optimize for simplicity. IMO this alone shouldn't be a problem, but where we get into trouble is when we have like 10+ little things to manage that require some context switching. If we plan on using it outside the monorepo, I would want the separate package, but my new thinking is that everything we do should be in monorepo. A quick search of our current GH org I don’t see any of the one-off packages using linting since they’re usually so ephemeral (though this might be different if we did expose a package to import 🤷).

# Other

- I *very* loosely think we should start (actually) pinning dependencies. I recognize there is no right answer here and therefore I’m open for anything. My current reasoning for this is:
    - Stuff like the recent [Ledger attack](https://support.ledger.com/hc/en-us/articles/15480255246365-Malicious-version-of-the-Ledger-Connect-Kit-?support=true) would be avoided
        - Devil’s Advocate, Log4j type stuff would be avoided if we don’t pin 🤷
    - I *think* this is useful for preventing bugs if. weremove `--no-cache` in the CCI config. My (possibly wrong) understanding is if our deps are not pinned and we implicitly expect a new build to use the latest version of a package, I *think* this wouldn't use the latest and we might have a hard time debugging.
    - Explicitness in upgrading and no unintended breakages on a random build (though I have not seen this in like 3+ years of looking for it in non-pinned deps)
- If the above is implemented, we can simplify all packages by hoisting pretty much all common config to the root. This includes:
    - tsconfig
    - eslint
    - nvmrc
    - npmrc
    - ignore files (I see projects do this but does scare me)
    - etc.

packages/eslint-config-hop-protocol/index.js
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/545/commits/19782a3ed6c46210c7cd1fea8dc39b26806c93d9

packages/eslint-config-hop-protocol/index.js
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/545/commits/19782a3ed6c46210c7cd1fea8dc39b26806c93d9

packages/api/package.json
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/545/commits/19782a3ed6c46210c7cd1fea8dc39b26806c93d9

packages/eslint-config-hop-protocol/package.json
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/545/commits/19782a3ed6c46210c7cd1fea8dc39b26806c93d9

packages/eslint-config-hop-protocol/index.js
miguelmota: Resolved wtih https://github.com/hop-protocol/hop/pull/545/commits/19782a3ed6c46210c7cd1fea8dc39b26806c93d9

packages/eslint-config-hop-protocol/index.js
miguelmota: Resolved wtih https://github.com/hop-protocol/hop/pull/545/commits/19782a3ed6c46210c7cd1fea8dc39b26806c93d9

packages/eslint-config-hop-protocol/index.js
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/545/commits/19782a3ed6c46210c7cd1fea8dc39b26806c93d9

packages/eslint-config-hop-protocol/index.js
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/545/commits/19782a3ed6c46210c7cd1fea8dc39b26806c93d9

packages/eslint-config-hop-protocol/index.js
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/545/commits/19782a3ed6c46210c7cd1fea8dc39b26806c93d9

packages/eslint-config-hop-protocol/index.js
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/545/commits/19782a3ed6c46210c7cd1fea8dc39b26806c93d9

packages/eslint-config-hop-protocol/index.js
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/545/commits/19782a3ed6c46210c7cd1fea8dc39b26806c93d9

packages/api/.eslintrc.json
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/545/commits/19782a3ed6c46210c7cd1fea8dc39b26806c93d9

packages/eslint-config-hop-protocol/index.js
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/545/commits/19782a3ed6c46210c7cd1fea8dc39b26806c93d9

packages/eslint-config-hop-protocol/index.js
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/545/commits/19782a3ed6c46210c7cd1fea8dc39b26806c93d9

packages/eslint-config-hop-protocol/index.js
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/545/commits/19782a3ed6c46210c7cd1fea8dc39b26806c93d9

packages/eslint-config-hop-protocol/package.json
miguelmota: Resolved with https://github.com/hop-protocol/hop/commit/19782a3ed6c46210c7cd1fea8dc39b26806c93d9

packages/eslint-config-hop-protocol/package.json
miguelmota: Resolved with https://github.com/hop-protocol/hop/commit/19782a3ed6c46210c7cd1fea8dc39b26806c93d9

packages/eslint-config-hop-protocol/index.js
miguelmota: Resolved with https://github.com/hop-protocol/hop/commit/19782a3ed6c46210c7cd1fea8dc39b26806c93d9

packages/eslint-config-hop-protocol/index.js
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/545/commits/6faaaa6e10cdfd87747c28d0be71feda0fdbe3ef

miguelmota:
@shanefontaine quick update; the feedback has been implemented (ie hoisting root files), and now figuring out the broken builds

packages/sdk/package.json
```diff
@@ -2,86 +2,68 @@
...
-    "lint": "npx eslint --fix \"src/**/*\"",
+    "lint": "eslint --fix",
     "version:bump": "npm version prerelease --preid=beta",
     "publish:latest": "npm publish --tag latest --access public",
     "publish:beta": "npm publish --tag beta --access public",
     "publish:next": "npm publish --tag next --access public",
     "dist-tag:latest": "npm dist-tag add \"@hop-protocol/sdk@$(jq -r .version <package.json)\" latest",
     "dist-tag:beta": "npm dist-tag add \"@hop-protocol/sdk@$(jq -r .version <package.json)\" beta",
     "dist-tag:next": "npm dist-tag add \"@hop-protocol/sdk@$(jq -r .version <package.json)\" next",
-    "version:sync": "package-sync-version",
```
miguelmota: by removing this, how do the other packages get updated with the latest version of this package?

.nvmrc
```diff
@@ -1 +1 @@
-20.10.0
```
miguelmota: what's the reason for downgrading node version?

packages/hop-node/src/vault/Vault.ts
```diff
@@ -1,11 +1,9 @@
-import { AaveVault } from './AaveVault'
 import { BigNumber, Signer } from 'ethers'
 import { Chain } from 'src/constants'
 import { YearnVault } from './YearnVault'
 
 export enum Strategy {
   Yearn = 'yearn',
-  Aaave = 'aave'
```
miguelmota, line 8: considering we've never used this feature, does it make sense to also remove yearn and the concept of vaults in hop node?

miguelmota:
Is it expected that when building `core`, with `npm run clean && npm run build`, it doesn't generate the top level dirs `networks/`, `addresses/`, `metadata/`, etc? Would not having those dirs break path imports, such as `const { networks } = require( '@hop-protocol/core/networks')`

shanefontaine:
* Remove custom type definitions (our own `types` dirs). Simplifies `tsconfig.json` by removing the need to define them but also needing to redefine the defaults, like `node_modules/types`. Most updated packages have correct types. Added `// @ts-expect-error` to a few imports that did need it. In theory, IMO if something doesn't have types we should be hesitant to use it as that means it is unmaintained and can cause headaches elsewhere.

* In public packages (`core`, `sdk`), use the `files` [option](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#files) in `package.json` as an NPM allowlist (published files must be included in `files` and not excluded in `.npmignore`).


* Use `rm -rf` for NPM scripts instead of `rimraf`. Removes possible debugging steps at the cost of compatibility.
  * I recognize that this breaks compatibility with non-Unix machines. However, (1) my understanding is that WSL is used by a decent amount of people on Window's these days (even more-so in OS world) and (2) I believe many other parts of the packages would not work with non-Unix, such as paths and env vars. I think `npx rimraf` makes sense for a lifecycle script, but can be reintroduced if/when we non-Unix-ify the repo.
  * I hit a case where `rimraf` wasn't working because I had stopped an install halfway through. `node_modules` knew about `rimraf` but didn't have all of its deps. In this case, `npx` used my `node_modules` `rimraf` instead of `global`, but it was corrupted. `npx` wouldn't ever try `global` because it still "saw" it locally. [Error message](https://gist.github.com/shanefontaine/208fa0ef2e35f8af284f8553212530f1) that was difficult to understand.

packages/hop-node/src/vault/Vault.ts
shanefontaine: Removed this along with some other unused code. Can reference this if we ever reintroduce it.

.nvmrc
shanefontaine: Re-updated to `20.10.0` everywhere ✔️

shanefontaine:
@miguelmota final commits have been pushed:

* Updated to `pnpm`
* Simplify/standardize scripts and deps + general cleanup

packages/sdk/package.json
shanefontaine: @miguelmota sorry I must have resovled this one without replying.

Updated all local deps to use the pnpm `workspace:*` [syntax](https://pnpm.io/workspaces#workspace-protocol-workspace). This resolves to the real numbers when published.

That means we just need to bump the `package.json` version instead of updating every file. We already do this with the `pnpm bump`, which uses native `pnpm` pogic.

shanefontaine:
@miguelmota final round of changes. Ready to merge at any time now.


* Remove binary from `hop-node`. Replaced with `ts-node` for local build/test and `node` for prod and docker.
  * IIUC binaries let someone run `hop-node` without a node env and w/o docker. I am under the impression that nobody does this (at this time). Current options to run `hop-node` is either:
    * Pull repo, build, run
    * Pull docker, run (build was already completed by CI prior to pushing to dockerhub)
  * IIUC the current `bin` calls into the `build` dir but instead should be using a fully compiled `bin` instead. If reintroduced, consider using `nexe` or some package that builds an executable bin.
  * Fun fact: I believe the latest version of [node](https://nodejs.org/api/single-executable-applications.html) has experimental binary support built-in (though likely won’t be stable for many years)

* Use `swc` for `hop-node` compile instead of Native Typescript for a 3.7x increase (5.39s → 1.46s) in `hop-node` build time (useful during dev and testing).
  * Generally goes against the goal of using natives/defaults, but feels practical in this case:
    * Meaningful speed increase (3.7x vs native `transpileOnly` and 6.7x vs how we were doing it)
    * [Suggested](https://typestrong.org/ts-node/docs/swc) as a main option by `ts-node` when using `ts-node`
    * No config (only an NPM package to download)

* RM `hop-node` and `stats-worker` `package_check.sh` since I believe they are now unused

* (Can be reverted) (SDK only) Use `esbuild` instead of `babel` for bundling + CDN
  * Gets rid of 65% of devDeps in SDK (40% of total deps) for faster / safer / easier installs & builds.

* Various package updates to non-outdated versions

packages/hop-node/src/cli/cli.ts
```diff
@@ -54,7 +52,9 @@ import './withdrawalProof'
 import './metrics/bonderBalance'
 import './metrics/bonderTxCost'
 
-program.version(packageJson.version)
+// TODO: Since package.json version isn't updated, it shouldn't be used here
+const version = '0.0.1'
```
miguelmota: maybe the git revision can be used as version here

miguelmota:
> Remove binary from hop-node. Replaced with ts-node for local build/test and node for prod and docker.

That seems fine. I think the Dockerfiles need to be updated to reflect that change.

> IIUC the current bin calls into the build dir but instead should be using a fully compiled bin instead. If reintroduced, consider using nexe or some package that builds an executable bin.

I think that the `bin/` folder in node apps are a convention to denote the entrypoint for the cli.
For example, the lodestar client has [`bin/lodestar.js`](https://github.com/ChainSafe/lodestar/blob/unstable/packages/cli/bin/lodestar.js) and in [package.json](https://github.com/ChainSafe/lodestar/blob/cbf349c73d7730c11d7295c27a61fbf36225df52/packages/cli/package.json#L20) they have `"bin": {"loadestar": "bin/loadstar.js"}}` for global command. In their Dockerfile they have the bin path as the [entrypoint](https://github.com/ChainSafe/lodestar/blob/cbf349c73d7730c11d7295c27a61fbf36225df52/Dockerfile#L47); `ENTRYPOINT ["node", "./packages/cli/bin/lodestar"]`.
In linux, the `/bin` folder means it contains executables and doesn't necessarily have to be compiled binary code. For example, `/bin/pip` is a readable python interpreted program (starts with `#!/usr/bin/python`) and it's not complied binary.
Just wanted to point that out to explain the reason it was like that.

> Use swc for hop-node compile instead of Native Typescript for a 3.7x increase (5.39s → 1.46s) in hop-node build time (useful during dev and testing).

Is `swc` only used during `start` script and not `pnpm build`? Seeing `tsc` being used instead of swc on build.

DeepinScreenshot_select-area_20240129013421


> (Can be reverted) (SDK only) Use esbuild instead of babel for bundling + CDN

Seems good to me 👍

I see that sdk CDN docs from README were removed. Any reason? For example, [ethers](https://docs.ethers.org/v5/getting-started/#getting-started--importing--web-browser) has cdn links/examples in docs

packages/hop-node/src/cli/cli.ts
shanefontaine: Agreed. Updated with 2ac8579f462ed501b0301ef582238b98e315cf0f.

shanefontaine:
>  I think the Dockerfiles need to be updated to reflect that change.

Updated with 9768d45da4759be33f811fdc4a38a7870b158b4f. I _believe_ that works, but plan on doing a quick pass on the Docker stuff today, most just for my understanding.

> bin

That is interesting and I did not know that. I thought `bin` 100% implied only binary files but see now I was incorrect. Thanks for the info 🙏 

> Is swc only used during start script and not pnpm build? Seeing tsc being used instead of swc on build.

Good catch, I didn't realize this. I just went to add it but then realized that it (understandably) completely [ignores `tsconfig`](https://github.com/swc-project/swc/issues/1348) and uses its [own config](https://swc.rs/docs/configuration/swcrc#compilation) that can be modified. I'm not too stoked on the mental overhead of "every package in Hop follows `tsconfig` except `hop-node`". But now also don't love that our local build/test is different than prod build 🤔 . Let me think about this but if you have thoughts I'm down.

One note is that there is a native way to do this too. Our `develop` branch builds are ~9.5s and our `swc` builds are 1.46s. However, our `develop` branch with `ts-node --transpileOnly` (i.e. don't look for type errs during build) takes 5.39s. Don't necessarily think that is a better solution, but think it is interesting.

> I see that sdk CDN docs from README were removed

I had removed bundling completely for a bit and removed from the README at that time. I dove in a bit and now have a better understanding so I added the bundle back but forgot to reintroduce to the README. Reintroduced with 2ac8579f462ed501b0301ef582238b98e315cf0f.

miguelmota:
> swc

Maybe there's a way to autogenerate the swc config from tsconfig, such as creating a small script to map tsconfig options to swc config options and output to .swcrc. I like the speed improvements and wouldn't be opposed for hop-node to use a different build system.

PR #547, opened 2024-01-26
frontend: mui v4 to v5 migration

miguelmota:
Updates frontend to use mui v5. 

Consists of dependency upgrades, updated imports, and theme config updates for mui v5. Should be pretty straightforward.


packages/frontend/package.json
```diff
@@ -24,14 +24,19 @@
     "not op_mini all"
   ],
   "dependencies": {
+    "@emotion/react": "^11.5.0",
```
shanefontaine: Pin deps here (unless this is intentional)

FYI I [updated](https://github.com/hop-protocol/hop/blob/develop/.npmrc#L5-L6) `.nvmrc` yesterday to pin by default. I _think_ I like this decision, but not 100% sold on it.

shanefontaine (approved):
@miguelmota Looks good to me. Does bumping those deps allow us to remove any of the overrides in root `package.json` or `.nvmrc`? I think no because they're related to `react-app-rewind`.

packages/frontend/package.json
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/547/commits/c82d01c381b94dd4eeac8a1338e5bf575a45ec0d

miguelmota:
@shanefontaine no since they are unrelated to mui, and related to react-app-rewired react-scripts

PR #550, opened 2024-02-10
hn: use node module specifiers

shanefontaine:
@miguelmota 

A few code (not infra) modernization updates. Prep for `v1`/`v2` split. All minor, mostly want to be on same page about first point:

* Use [`node:` module specifiers](https://2ality.com/2021/12/node-protocol-imports.html) for more explicit imports
* Various, non-exhaustive `#` private methods instead of `_` syntax
* Various, non-exhaustive types definitions.


miguelmota (approved):
Looks perfect :+1:

PR #584, opened 2024-02-16
core: Replace typechain with abitype

miguelmota:


miguelmota:
Closing draft PR.

Reasons / issues I ran into:

- using `abitype` to  implement a typed ethers `Contract` for dynamic abis became overly complex and requires in-depth knowledge of ethers Contracts and everything it extends, so would require a lot more effort and research.
- an [older version](https://github.com/wevm/abitype/discussions/48) of `@wagmi/core` has support for typed contracts for `ethers@5` using `abitype`, however some things were missing in it's type implementation which broke builds.
- `viem` has support for `abitype` for ethers `Contract`, however it only works with `ethers@6`. 


Possible solution:

Ricmoo from ethers mentioned two weeks in an [issue thread](https://github.com/ethers-io/ethers.js/issues/4564#issuecomment-1926341845) he's working on a version of `Contracts` based on `abitype`. Given that, I think it makes sense to try to upgrade to `ethers@6` at some point and use the typed Contract class when it's ready if the plan is to stick with ethers long term. Migrating to `viem` could also be a good alternative. Of course neither solution is trivial so would require more thought. In the meantime, sticking with `typechain` and `ethers@5` shouldn't be an issue.

PR #596, opened 2024-04-14
Feat/minor optimizations

shanefontaine:
* hnc: remove unused TSConfig rule and add explainer to custom rules
* sdkc: remove `luxon` to reduce ~6% of install size
* Remove docs scripts and deps from sdk-core (I believe they are all covered at the `vX-sdk` level)
* Update `sdk` readme to remove dead link and slightly update formatting and `sdk-core` to reflect itself and to provide less information as to promote usage of one of the SDK packages (similar to [Uniswap core](https://github.com/Uniswap/sdk-core/blob/main/README.md)).
* Update `sdk-core` from beta to non-beta (0.0.3). I think it is fair to not call it `beta` bu am not 100% convinced the way I updated is the best way to do it. Will see in practice and happy to discuss/adjust.
* Remove `ethers/lib/utils` in favor of `utils`
* One more `node` [security update](https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2/)
* Code updates
  * Use [new](https://nodejs.org/docs/v20.12.1/api/process.html#processloadenvfilepath) `node` built-in function `loadEnvFile` in place of `dotenv` (c93532055675271f503fda3203207761727d4f4e). Reduces number of import side effects for better tree shaking (in theory).
  * Use `never` return type for polling methods (73888f23dba3f7ba0d1e28ad42a568f0078518b1)


miguelmota (approved):

miguelmota:
Changes look good to me 👍

PR #597, opened 2024-04-20
chore: remove repetitive words

overallteach:


overallteach:
If anything need I to do, Please feel free to tell me.

PR #603, opened 2024-05-14
upgrade checkout action to version using node20

cario-dev:
Node16 is deprecated and Actions using it will stop working 3rd of June
https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

See warnings here
https://github.com/hop-protocol/hop/actions/runs/9075821996

### DIFF
--- a/.github/workflows/debug.yml
+++ b/.github/workflows/debug.yml
@@ -14,7 +14,7 @@ jobs:
         working-directory: .
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


shanefontaine:
Thanks for the contribution @cario-dev . These changes are already being handled in #602 so I will now close this.

cario-dev:
> Thanks for the contribution @cario-dev . These changes are already being handled in #602 so I will now close this.

I can't see that these changes are handled in the linked PR

PR #604, opened 2024-05-17
frontend: Migrate from onboardjs to web3-react lib

miguelmota:
Migrates from [onboard](https://github.com/blocknative/web3-onboard) to [web3-react](https://github.com/Uniswap/web3-react/tree/main).

Switching over cleaned up a lot of the Web3Context Provider/hook since there used to be a lot of overrides with onboardjs to customize it for our needs. The nice thing about web3-react is that it allows you to specify your own modal and also provides easy to use hooks. 

The changes should be pretty easy to follow.


packages/frontend/src/contexts/Web3Context.tsx
```diff
@@ -1,311 +1,361 @@
...
-  const onboardChains :any[] = []
+  const items: ChainInfo[] = []
 
   for (const chain of chains) {
-    const id = chainIdToHex(Number(chain.chainId))
+    const chainId = Number(chain.chainId)
     const token = chain.nativeTokenSymbol
-    const label = `${chain.name} ${capitalize(reactAppNetwork)}`
+    const explorerUrls = chain.explorerUrls
+    const isEthereum = chain.slug === 'ethereum'
```
shanefontaine: Just FYI, `isL1` now lives on the `chains` object and can be used here.

packages/frontend/src/contexts/Web3Context.tsx
```diff
@@ -1,311 +1,361 @@
...
-      if (_address) {
-        await onboard.setChain({ chainId: networkId })
+      try {
+        // Example: using rpc directly:
+        // const hexChainId = `0x${Number(wantChainId).toString(16)}`
+        // await provider.send('wallet_switchEthereumChain', [{ chainId: hexChainId }])
+
+        await connector.activate(wantChainId)
+      } catch (err) {
+        const chainNotAddedErrorCode = 4902
```
shanefontaine, line 322: Is this error code specific to MM or global? Either way, is it supported by all wallets (within reason)?

packages/frontend/src/contexts/Web3Context.tsx
```diff
@@ -1,311 +1,361 @@
...
 
-  function requestWallet() {
+  useEffect(() => {
     const update = async () => {
       try {
-        localStorage.clear()
-        const [primaryWallet] = onboard.state.get().wallets
-        if (primaryWallet) {
-          await onboard.disconnectWallet({ label: primaryWallet.label })
+        disconnectWallet()
```
shanefontaine: Should this be awaited? Little context about what these are doing, but looks like a possible edge-case race condition can set `connectorToUse.activate()` before completion of `connector.resetState()` in some weird case.

Not sure it even matters.

shanefontaine (approved):
👍

packages/frontend/src/contexts/Web3Context.tsx
miguelmota: I'm not sure where it originated (maybe metamask?) but that error code is standard across wallets that support `wallet_switchEthereumChain`

packages/frontend/src/contexts/Web3Context.tsx
miguelmota: Resolved with https://github.com/hop-protocol/hop/pull/604/commits/24f772651f9387405fb2026535e003ea40f55c7b

packages/frontend/src/contexts/Web3Context.tsx
miguelmota: good catch, awaits have been added. 

Resolved by https://github.com/hop-protocol/hop/pull/604/commits/2a8d3c237e73084fbb9d1f0dbd0e1c107b83d304

PR #605, opened 2024-05-18
Frontend lint fixes

miguelmota:
Depends on https://github.com/hop-protocol/hop/pull/604


packages/frontend/package.json
```diff
@@ -12,7 +12,7 @@
     "start": "REACT_APP_GIT_SHA=`git rev-parse --short HEAD` vite",
     "dev": "pnpm start",
     "lint": "eslint . --fix",
-    "build": "REACT_APP_GIT_SHA=`git rev-parse --short HEAD` tsc && vite build",
+    "build": "REACT_APP_GIT_SHA=`git rev-parse --short HEAD` tsc && vite build && echo '{\"type\": \"module\", \"imports\": { \"#*\": \"./*\" }}' > dist/package.json",
```
shanefontaine, line 15: This should only be required for dual exports. Since this is single, I believe this will then be redundant since package.json resolution will observe this file directly.

packages/frontend/package.json
```diff
@@ -77,14 +81,20 @@
...
-    "@types/react-transition-group": "4.4.10",
     "buffer": "6.0.3",
     "dotenv": "16.4.1",
     "eslint-plugin-react": "7.33.2",
     "eslint-plugin-react-hooks": "4.6.0",
     "vite": "5.0.13"
+  },
+  "imports": {
+    "#*": {
+      "import": "./src/*",
```
shanefontaine, line 96: Should just be a single `"#*": "./dist/*"`. Will help tools evaluate to `dist` and will be more representative of prod.

I believe `src` can be used but would require us to manually configure a bunch of other tools, I think.

shanefontaine (approved):
LGTM 👍 

**Suggestion**

* I think ESM requires using the node protocol (`import path from 'node:path'`) and might be needed here.

**Note**

* I _believe_ `vite` handles TS config and bundles this into ESM. But I do not 100% understand all the implications of the TS config nor what a FE should be, but do believe this is all good & handled by `vite`.

* For most packages I think we've been using relative imports for imports within a module and subpath imports when reaching outside of a module. In this case, for example, all the `components` reference's to each other should use relative imports. However, I believe it is ideal as-is and IIUC module classification/terminology is diff for each framework. So no change here, just a comment that FE appears to be a slightly different way of thinking than Node apps, as expected.

PR #607, opened 2024-05-22
Feature/fe relay page

miguelmota:


packages/frontend/src/pages/Relay/Relay.tsx
```diff
@@ -0,0 +1,213 @@
...
+          // example commit txs
+          // let commitTxHash = '0x86f1dfc3ced80aa27c1116ebda552b5cd5009eac5f6f3c2d01a521faae398677' // base ETH
+          // let commitTxHash = '0xb3b1e32b65aab3dd5374130fe92ff3108fef1d5d61be62bb99e041e596b64265' // optimism ETH
+          // let commitTxHash = '0x5a75ff2131895ec42da3bc851df8854683f20bb7dadb6f149e527da021d91456' // polygon ETH
+          // let commitTxHash = '0x43c5f6baf4100fbd7924a9f0664f20ed8b36bf3feb4218aea41d1e01f5076bf4' // arbitrum ETH
+          // let commitTxHash = '0x886158d6ce41723af5364b9f89d51416c09345ff920c8e7af23dfd67ccd34da2' // nova ETH
+          // let commitTxHash = '0xc02c7337289afbc5d73199e26585da58b7cf77334f8d49255751a6da421a13a4' // gnosis ETH
+          let commitTxHash = ''
+          if (!commitTxHash) {
+            const event = await getTransferCommittedEventForTransferId(selectedNetwork.slug, 'ETH', txHash)
```
shanefontaine: I think 'ETH' should be a passed-in token? Unless it is ignored within the called method. If so, maybe just note that somewhere.

shanefontaine (approved):
* The L2 SDKs are listed as both deps and peerDeps and I think they should only be deps.
* Linea is commented out in `relay/index.ts`. Not sure if that is intentional.
* Is the relay page displayed to the user or must they type in the url? It looks like the latter. I'd prefer that given how infrequent this should be and to avoid polluting the navbar, but haven't given this much thought.

Rest LGTM. High level I do think the `chains` dir in `hop-node` is substantial enough to rip out into its own package one day if it makes sense to help avoid duplication like this. However, that is a decent lift to package it all up and I don't think there would be a ton of benefit.

miguelmota:
@shanefontaine sorry didn't realize i marked it as ready for review as it was still a draft. 

- I just pushed more clean up updates so should be good for final review.
- I did move the relay logic to the frontend instead of having it in sdk to not bloat the sdk. Linea sdk was originally causing build errors so it was commented out but now it's working.
- All L2 chains should be supported for L2->L1 relays.
- The relay page is like the withdrawal page where they must type in the url.
- Agree that having the chains package from hop-node be a generalized package would be nice to have.

DeepinScreenshot_select-area_20240522201339

shanefontaine:
@miguelmota  LGTM 👍

